PR #3, opened 2026-03-03
doc: development prereq and librechat model default

gjanco:
Add to the development/README so that the user  knows they need aura web-server running
Changed the order of models in the librechat.yaml
since we are guiding the user toward OpenAI first.

Ref: LOG-22815

### DIFF
--- a/development/configs/librechat.yaml
+++ b/development/configs/librechat.yaml
@@ -8,13 +8,13 @@ endpoints:
   custom:
     # Aura Server - OpenAI-compatible endpoint
     - name: "Aura"
-      apiKey: "${MEZMO_API_KEY}"  # Use environment variable
+      apiKey: "not_used"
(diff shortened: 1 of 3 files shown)


ericlake (approved):

development/configs/librechat.yaml
```diff
@@ -8,13 +8,13 @@ endpoints:
   custom:
     # Aura Server - OpenAI-compatible endpoint
     - name: "Aura"
-      apiKey: "${MEZMO_API_KEY}"  # Use environment variable
+      apiKey: "not_used"  # Use environment variable
```
adrianfurlong: Should we remove the env var comment or update to explain why the fixed string?

gjanco: dang it...yes.

PR #7, opened 2026-03-04
chore(release): bump the version to the correct version

justintime4tea:
This will bump the version to be the correct version

Ref: LOG-XXXX

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "1.14.4"
+version = "1.14.5"
 edition = "2021"
 authors = ["Mezmo <support@mezmo.com>"]
 license = "Apache-2.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -8,7 +8,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "1.14.4"
+version = "1.14.5"
```
gjanco, line 11: lets skip to `1.14.6` so we can be ahead of answerbook.

gjanco (changes requested):

gjanco (approved):

teriyakichild (approved):

Cargo.toml
justintime4tea: We can't skip, the automation will "skip" for us because we used "chore:" the result should be 1.14.6

PR #12, opened 2026-03-05
fix(ci): use public mezmo/aura-mock-mcp for tests

justintime4tea:
This updates the base.yml for docker compose files so that the public mezmo/aura-mock-mcp image is used for tests.

Ref: LOG-23365


compose/base.yml
```diff
@@ -17,7 +17,7 @@ services:
     # Use GCR image in CI, local build for development (Mac ARM needs local build)
     # CI: Set MOCK_MCP_IMAGE in Jenkinsfile
     # Local: Run `make mock-mcp-build` to build from ../aura-mock-mcp
-    image: ${MOCK_MCP_IMAGE:-aura-mock-mcp:local}
+    image: ${MOCK_MCP_IMAGE:-mezmo/aura-mock-mcp}
```
dominic-mcallister-logdna, line 20: Should we remove the `MOCK_MCP_IMAGE` environment variable so both internal and external devs use the same image?

gjanco: Yes, please!

justintime4tea: The MOCK_MCP_IMAGE is there so you can choose a different image to build with. It defaults to "mezmo/aura-mock-mcp" so if not explicitly set, it will use the docker hub.

gjanco (approved):

PR #16, opened 2026-03-06
chore(ci): releases

kristof-mattei:
* Docker container is built on every PR, not published.
* Upon accepting PR, changelog is generated + new tag. 
* New release is published with ^ changelog + tag
* New tag kicks off Docker container build + publish for `v{x}`, `v{x.y}`, `v{x.y.z}` and `latest`

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -1,0 +1,49 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+name: Auto Release
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [main]
+
+concurrency:
+  group: "${{ github.workflow }}"
+  cancel-in-progress: false
+
+permissions:
+  contents: write
+
+jobs:
+  release:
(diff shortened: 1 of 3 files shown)


esatterwhite:
The binary and container should be built and released from what is on main after it has merged. We don't release from prs or branches

We don't need merge commits

esatterwhite (changes requested):
This is very much outside the way the company has managed releases.
We need to figure why there is such a disconnect with this repository, figure out how we can bring this closer to our current standards, and determine what from this and actually neccesary.

gjanco:
> This is very much outside the way the company has managed releases. We need to figure why there is such a disconnect with this repository, figure out how we can bring this closer to our current standards, and determine what from this and actually neccesary.

We discussed this in the Innovation Meeting Thursday.  There was concern that malicious code could be run in Jenkins when we accept a PR.  GH Actions was brought up as a good alternative (Phil will probably be moving vector/edge there as well).

Shearerbeard:
Closed - stale

PR #22, opened 2026-03-11
fix: strip system role and empty messages

Shearerbeard:
Interaction with openwebui and librechat will post their own system messages - we don't really have a way to modify system content after build time and it can be considered an attack vector - strip both the system message (incompabible with rig without upgrading to a user message) and empty messages that can blow up bedrock providers.

Ref: LOG-23402


Shearerbeard:

justintime4tea (approved):

mdeltito (approved):

PR #23, opened 2026-03-11
fix: stop breaking early on pre tool text

Shearerbeard:
Right now we mistakenly break out of our
streaming collector (makes streaming agents
compatible with non streaming calls) breaks
its loop on FinalMarker when in actually
FinalMarker is emitted at the end  of every agent turn so if the agent emits text before the tool call
we exit before we have the LLM get or or respond
to the result. Fix this by removing the early exit.

Ref: LOG-23401


Shearerbeard:
recheck

justintime4tea (approved):

mdeltito (approved):

PR #27, opened 2026-03-12
fix(jenkins): fix intermittent ci failures due to instance re-use

justintime4tea:
Updates RUSTUP_HOME and CARGO_HOME to use a workspace-local directory instead of /opt/ so that when a jenkins instance is re-used, cargo will continue to work for CI stages.

Ref: LOG-23413


dominic-mcallister-logdna (approved):
nice!

PR #31, opened 2026-03-21
feat(orchestration): add OTel tracing spans to orchestration mode

teriyakichild:
Add structured tracing spans covering all orchestration phases: orchestration, planning, iteration, worker, synthesis, evaluation. Record rich attributes (goal, routing, task metadata, quality score, token usage) and map span kinds in the OpenInference exporter.

Fix span propagation: tokio::spawn in Orchestrator::stream() and ToolWrapper::call now inherit the parent span so child spans nest correctly under agent.stream in Phoenix/Jaeger.

Known limitation: token counts on planning and evaluation spans will be zero until an upstream Rig change lands to always yield per-turn Final regardless of is_text_response.

ref: LOG-23471

### DIFF
--- a/crates/aura/src/logging.rs
+++ b/crates/aura/src/logging.rs
@@ -28,10 +28,10 @@
 //! agent.stream (AGENT, ROOT)        <- Phoenix root span, lives for full stream duration
 //!   ├── user.id, session.id, metadata, input.value, output.value, tokens
 //!   └── agent.turn (LLM)           <- from Rig fork (reuses agent.stream as parent)
-//!       ├── chat_streaming (LLM)    <- from Rig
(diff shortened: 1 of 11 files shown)


crates/aura/src/logging.rs
```diff
@@ -30,7 +32,25 @@
 //!       ├── execute_tool (TOOL)     <- from Rig (no error status — see below)
 //!       │   └── mcp.tool_call (TOOL) <- from Aura, canonical tool span with error status
 //!       └── chat_streaming (LLM)    <- from Rig
+//! ```
```
Shearerbeard, line 35: Some of this might be good to put into markdown docs

Shearerbeard (changes requested):

Shearerbeard (approved):

PR #32, opened 2026-03-25
feat(scratchpad): context window management for large mcp tool outputs

justintime4tea:
Intercept large MCP tool results before they fill the context window,
save them to disk, and give the LLM eight read-only exploration tools
to selectively extract only what it needs. Works for both single-agent
and orchestration modes.

Key components:
- ScratchpadWrapper: intercepts tool outputs exceeding per-tool token
  thresholds, writes to disk, returns a compact pointer to the LLM.
  Tool errors (`CallOutcome::is_error()`) bypass interception so the
  LLM always sees errors inline.
- Exploration tools: schema, item_schema, head, grep, get_in,
  iterate_over, slice, read — each budget-checked before returning
- ContextBudget: tracks estimated token consumption (LLM-reported
  ground truth + scratchpad reads) and prevents context overflow
- Per-agent fresh ContextBudget scoped to its effective LLM's
  context_window — single-agent and workers never share a budget
- Shared scratchpad/setup.rs wire-up used by both single-agent
  builder and orchestration's create_worker
- aura.scratchpad_usage SSE event emitted per-agent at end of
  stream (single-agent and worker contexts; base aura.* namespace)

Configuration:
```toml
memory_dir = "/path/to/aura-memory-dir"

[agent.scratchpad]
enabled = true
context_safety_margin = 0.20
max_extraction_tokens = 10_000
turn_depth_bonus = 6

[mcp.servers.<name>.scratchpad]
"*"         = { min_tokens = 512 }
"tool_name" = { min_tokens = 1024 }   # per-tool override

[orchestration.worker.<name>.scratchpad]   # optional per-worker
max_extraction_tokens = 5_000
```

Token-based budgeting, per-call limits and turns:
- Per-tool min_tokens thresholds (was min_bytes)
- Per-call extraction limit (max_extraction_tokens, default 10k)
- Auto-increased turn_depth when scratchpad is wired up
  (turn_depth_bonus, default 6) — applied in both modes
- Real tokenization (tiktoken-rs) for MCP schemas, preambles, and
  scratchpad/builtin tool definitions
- Per-worker initial_used uses mcp_filter to count only filtered
  tools
- LLM-reported per-turn input/output tokens fed back into budget
  as ground truth (orchestration via StreamItem::TurnUsage;
  single-agent via StreamingRequestHook's
  on_stream_completion_response_finish)
- Task prompt tokens recorded in budget before worker execution

Glob pattern matching:
- [mcp.servers.*.scratchpad] keys are glob patterns matched
  against tool names at interception time (not exact HashMap
  lookup)
- Most-specific pattern wins (longest match); on length ties the
  smallest threshold wins. Lets you write
  "*" = { min_tokens = 4096 } as a catch-all and override per tool

Memory dir resolution:
- New top-level memory_dir shared by single-agent scratchpad and
  orchestration persistence
- AgentConfig::effective_memory_dir() resolves with legacy fallback
  to [orchestration.artifacts].memory_dir
- Storage: {memory_dir}/scratchpad/ (single-agent) or
  {memory_dir}/{run_id}/iteration-{n}/scratchpad/ (orchestration)

Pretty-print & minified JSON:
- Pretty-print JSON at write time so line-based tools work on
  minified provider responses
- WriteResult struct replaces tuple returns, carries companion info

Companion file extraction:
- Auto-extract large structured string values from JSON as
  companion files (escaped JSON → pretty-printed .json; markdown
  → .md)
- ContentFormat::Markdown variant with detection heuristic
- Shared prepare_write / make_companion helpers
- Untrusted JSON keys slugified + path-validated to block traversal

Markdown detection:
- ATX header rule: 1–6 `#` chars followed by a space or tab.
  Header presence alone is sufficient - header-only documents are
  still markdown, and the prior list co-requirement missed real
  content.
- The space-after-hash requirement excludes leading `#` look
  alikes (shebangs `#!/bin/bash`, `#include`, `#TODO` comments)
  without needing a co-occurring marker.
- Schema tool dispatches by extension (.json → JSON, .md → md)
- analyze_markdown_structure() parses headers into a section tree
  with line ranges and keys
- Format-aware schema_too_large error suggestions

get_in string pagination:
- offset/limit params paginate large string values by line
- Raw string content returned (not JSON-encoded) so embedded
  newlines render as real lines
- Early return for non-string values, extracted get_in_paginated
  helper

iterate_over large-string truncation:
- String fields >5 lines truncated with preview + get_in path hint
  to prevent memory spikes when array items carry huge strings

grep & read guardrails:
- Reject regex patterns over GREP_MAX_PATTERN_LEN (1024 bytes)
- Cap grep output at GREP_MAX_OUTPUT_TOKENS (32_768) via
  per-section token accounting (avoids O(n²) re-tokenization)
- read tool pre-flight: stat-based size check rejects files whose
  byte size proves they can't fit (file_bytes/3 > limit), avoiding
  loading multi-MB files just to reject them on real tokenization

Wrapper composition & safety:
- Persistence wrapper captures RAW tool output via call-id-keyed
  cache before later wrappers rewrite it, so retries see real data
  rather than a [scratchpad: ...] pointer
- Orchestration ordering: [observer, duplicate_guard,
  ...scratchpad, persistence] - persistence runs first in reverse
  iteration so it sees raw before scratchpad rewrites
- Single-agent ordering: [scratchpad, existing_user_wrapper] so a
  caller-supplied wrapper observes raw output, matching the
  orchestration convention
- New composed-ordering test locks down the persistence-after
  scratchpad invariant through a real ComposedWrapper, not just
  the per-wrapper cache rendezvous in isolation
- Sync writes for scratchpad/companion files: actix-web uses
  current_thread tokio per worker so block_in_place would panic,
  and async-ifying ToolWrapper would force a trait-level refactor
  ( #[async_trait]'s Send-only futures vs WrappedTool's Send+Sync
  return type). KB-sized writes are sub-millisecond, far cheaper
  than the LLM round-trip that follows

Schema/doc cleanup on PersistenceWrapper:
- Drop `"minLength": 1` from the `_aura_reasoning` schema. The
  field is intentionally omitted from `required` (per the existing
  comment); advertising minLength on a non-required field could
  cause strict-validation providers to reject empty values.
  validate_args remains warn-only.
- Module + struct doc-comments updated to say the field is added
  as optional, not required.

Event surface symmetry:
- Scratchpad exploration tool calls (head/slice/grep/...) are
  suppressed from the single-agent SSE event surface
  (aura.tool_requested / aura.tool_complete / aura.tool_usage)
  via is_scratchpad_tool() in StreamingRequestHook. Matches
  orchestration, where these tools aren't wrapped by
  ObserverWrapper and therefore don't surface in
  aura.orchestrator.tool_call_* events either

Companion pointer hints in wrapper:
- Pointer message lists companion files with format-specific tool
  suggestions

Ref: LOG-23439


justintime4tea:
I must have rebased and brought a couple commits in from main (the commits that aren't mine). I can dig those out but those changes are minimal and not related to scratchpad so you should be able to easily see the changes that _are_ related to scratchpad.

Shearerbeard (approved):
LGTM

compose/scratchpad.yml
```diff
@@ -0,0 +1,106 @@
...
+        -d '{"jsonrpc":"2.0","method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"healthcheck","version":"1.0"}},"id":1}'
+        > /dev/null || exit 1
+      interval: 5s
+      timeout: 5s
+      retries: 10
+      start_period: 15s
+    networks:
+      - aura-test
+
+  aura-web-server:
```
dominic-mcallister-logdna, line 61: Seems like a healthcheck should be added to aura-web-server too

dominic-mcallister-logdna, line 61: Thinking out loud, should we be explicit in the naming of elements (including server name) such as the toml and memory_dir that orchestration is in the pathway here? Or is it just assumed that Orchestration is "the way"

crates/aura-config/src/config.rs
```diff
@@ -41,6 +46,10 @@ pub struct WorkerConfig {
     /// scoped to whichever model the worker actually runs.
     #[serde(default)]
     pub llm: Option<LlmConfig>,
+    /// Per-worker override of `[agent.scratchpad]`. Parsed from
+    /// `[orchestration.worker.<name>.scratchpad]`.
+    #[serde(default)]
+    pub scratchpad: Option<aura::scratchpad::ScratchpadConfig>,
```
dominic-mcallister-logdna: For consistency with LlmConfig, should a `use` statement be used in the file so we don't need to fully qualify ScratchpadConfig?

crates/aura-config/src/config.rs
```diff
@@ -302,6 +311,9 @@ pub enum McpServerConfig {
         env: HashMap<String, String>,
         #[serde(default)]
         description: Option<String>,
+        /// Per-tool scratchpad interception thresholds (glob-matched on tool name).
+        #[serde(default)]
+        scratchpad: HashMap<String, aura::scratchpad::ScratchpadToolEntry>,
```
dominic-mcallister-logdna: Same thought for ScratchpadToolEntry.

crates/aura/src/orchestration/orchestrator.rs
```diff
@@ -512,6 +507,89 @@ impl Orchestrator {
...
+                &accessible_tools,
+                worker_filter,
+                &scratchpad_tool_map,
+            );
+
+            if !has_matching_tool {
+                tracing::info!(
+                    "Worker {}: scratchpad enabled but no MCP tool matches a scratchpad threshold; skipping",
+                    task_id
+                );
```
dominic-mcallister-logdna: Is this situation a misconfiguration that should hard-error?

crates/aura/src/orchestration/orchestrator.rs
```diff
@@ -512,6 +507,89 @@ impl Orchestrator {
...
+                        token_counter,
+                    })
+                    .await?;
+
+                scratchpad_tools.push(build.wrapper);
+                worker_config.scratchpad_tools_config = Some(build.tools_config);
+            } else {
+                tracing::warn!(
+                    "Worker {}: scratchpad enabled but context_window unset on effective LLM; skipping",
+                    task_id
```
dominic-mcallister-logdna: Same question here, should this be a hard-error?

crates/aura/src/scratchpad/storage.rs
```diff
@@ -0,0 +1,701 @@
...
+        let mut has_header = false;
+        let mut has_list = false;
+        for line in content.lines().take(50) {
+            let trimmed = line.trim_start();
+            if trimmed.starts_with('#') {
+                has_header = true;
+            }
+            if trimmed.starts_with("- ") {
+                has_list = true;
+            }
```
dominic-mcallister-logdna: Why is the presence of a list required to be considered markdown?

crates/aura/src/builder.rs
```diff
@@ -104,9 +120,98 @@ pub struct Agent {
...
+            .unwrap_or_default();
+        let filter = config.agent.mcp_filter.as_deref().unwrap_or(&[]);
+        if !scratchpad::has_accessible_scratchpad_tool(
+            &accessible_tools,
+            filter,
+            &scratchpad_tool_map,
+        ) {
+            tracing::info!(
+                "Single-agent scratchpad enabled but no MCP tool matches a scratchpad threshold; skipping"
+            );
```
dominic-mcallister-logdna, line 176: I asked about the orchestration version too, but should this be a hard-stop since scratchpad is enabled but not fully configured?

crates/aura/src/scratchpad/context_budget.rs
```diff
@@ -0,0 +1,524 @@
...
+
+/// Create a token counter for the given provider and model.
+///
+/// Currently uses tiktoken-rs for all providers. OpenAI models resolve to
+/// their exact tokenizer; others default to `o200k_base`. To add a
+/// provider-specific tokenizer, implement `TokenCounter` and add a match arm.
+pub fn token_counter_for_provider(provider: &str, model: &str) -> Arc<dyn TokenCounter> {
+    match provider {
+        "openai" => Arc::new(TiktokenCounter::for_model(model)),
+        // Future: add dedicated tokenizers for other providers here
```
dominic-mcallister-logdna, line 99: The tiktoken-rs crate suggests using HuggingFace tokenizers for non-openai stuff. Is that of value at this time or is that meant as a future follow up per this comment?

compose/scratchpad.yml
justintime4tea: The aura-web-server inherits its healthcheck from compose/base.yml (the curl -f http://localhost:8080/health one defined there). This file is a stack overlay layered on top of base.yml, so the service definition gets merged. The single-agent service has its healthcheck inline because it's a new service introduced in this overlay with no base definition to inherit from. The test-integration-scratchpad-local-up target in the Makefile actually relies on that inherited healthcheck — it polls aura-web-server's Health status.

crates/aura/src/orchestration/orchestrator.rs
justintime4tea: Intentionally soft here. With inheritance from [agent.scratchpad], a worker can opt-in by default while its mcp_filter legitimately excludes every tool that has a threshold configured — e.g., a quick summarizer worker that only uses small-output tools. Hard-erroring would force every such worker to explicitly disable scratchpad. What about bumping this from info! to warn! for higher visibility since it seems worth the surfacing/attention? — I don't think it should fail worker construction though.

crates/aura/src/builder.rs
justintime4tea: Intentionally soft here as well for similar reasons as previous comment. If we hard error it forces configuration to explicitly disable scratchpad in certain scenarios where inheritance occurs which adds extra toil. Switching this to warning instead of info could work here as well? so at least attention is drawn to it?

crates/aura/src/scratchpad/context_budget.rs
justintime4tea: We landed on future follow-up for specific tokenizers but a good call out none-the-less.

Two things make the o200k_base fallback acceptable for v1: 

(1) context_safety_margin = 0.20 reserves headroom that easily covers the estimation error vs. true provider tokenizers. 

(2) the budget gets corrected to ground truth every turn by the LLM-reported input_tokens + output_tokens feedback loop (orchestration: StreamItem::TurnUsage; single-agent: on_stream_completion_response_finish in the streaming hook).

The tokenizer matters mostly for pre-flight extraction estimates, where small errors are absorbed. Totally worth adding HF tokenizers for Anthropic/Llama if we hit accuracy issues with non-OpenAI workers.

crates/aura/src/scratchpad/storage.rs
justintime4tea: Good catch — this is too restrictive. A header-only markdown doc (e.g., # Title followed by paragraphs, no bullets) wouldn't be detected, and there's no good reason for that.

The reason both checks ended up coupled was that I was trying to avoid false positives from non-markdown files that happen to start lines with # like Python comments, shell scripts (#!/bin/bash, #include). The cleaner fix is to tighten the header check itself: require the space ("# ", "## ", "### "), which is what proper markdown headers actually use and what naturally excludes code-comment markers. Then headers alone are sufficient and the list requirement goes away. Might be a couple other things I can do to clean this up 🙏

I'll update this. Appreciate the callout Dom!

crates/aura/src/orchestration/orchestrator.rs
justintime4tea: Nice catch — going to update this to hard-error so it matches builder.rs, which already returns an error on this exact condition (with the comment "Validation enforces these upstream; re-check here so runtime misconfiguration fails loudly instead of silently degrading"). Lining the two paths up nicely removes the asymmetry 👍

Fun side note: this is technically redundant! aura-config::Config::validate_scratchpad() already catches it at startup — it walks every scratchpad-enabled worker, resolves the effective LLM the same way (worker.llm.unwrap_or(&agent.llm)), and hard-errors if context_window is unset. So in the normal load-and-validate flow, this branch is unreachable. 

The defense-in-depth only really matters if someone constructs an AgentConfig programmatically and skips validation (tests, downstream library consumers, future code paths) — but it's cheap to keep, and after I update it, happily consistent.

crates/aura/src/scratchpad/context_budget.rs
Shearerbeard: I think its fine to move additional tokenizers to a followup ticket - you can kind of tune the inaccuracy with the threshold which I assume is configurable @justintime4tea ?

justintime4tea: The threshold is configurable. If you were running into inaccuracy that leads to pushing past the budget, you could increase the threshold as one way to prevent busting through the budget. The current threshold should already save us from going past budget because it's larger than potential inaccuracy.

compose/scratchpad.yml
justintime4tea: I'm not 100% sure I understand - forgive me if I'm being dense 😅 Could you try reframing or spelling it out more for me?

dominic-mcallister-logdna: Had a conversation offline. No action is necessary.

justintime4tea: Since aura-web-server can load multiple configs, I updated things so that there's only one aura-web-server - the tests can test both single and orch by supplying different "model" at request time :)

crates/aura/src/orchestration/config.rs
```diff
@@ -205,6 +205,11 @@ pub struct WorkerConfig {
     /// math (e.g. scratchpad sizing, LOG-23439).
     #[serde(default)]
     pub llm: Option<LlmConfig>,
+
+    /// Per-worker override of `[agent.scratchpad]`. Parsed from
+    /// `[orchestration.worker.<name>.scratchpad]`.
+    #[serde(default)]
+    pub scratchpad: Option<crate::scratchpad::ScratchpadConfig>,
```
dominic-mcallister-logdna: nit: should we have a use statement at the top of the file instead of fully qualifying the config reference?

CLAUDE.md
```diff
@@ -72,16 +72,35 @@ aura/
...
+- `memory_dir` is a top-level TOML field shared by single-agent scratchpad and orchestration persistence
+
 ### Orchestration (Multi-Agent)
 - Coordinator/worker architecture with DAG-based parallel task execution
+- Per-worker LLM overrides: workers inherit `[agent.llm]` by default; `[orchestration.worker.<name>.llm]` overrides it (different model, same provider config). Resolved inline at worker construction (`worker.llm.as_ref().unwrap_or(&agent.llm)`)
 - Dependency-aware multi-wave execution with quality evaluation
 - Iterative re-planning loops (`quality_threshold`, `max_planning_cycles`)
 - Three-way routing: direct answer, orchestrated plan, clarification
-- 11 `aura.orchestrator.*` SSE events for real-time visibility (see `docs/streaming-api-guide.md`)
+- 13 `aura.orchestrator.*` SSE events for real-time visibility (see `docs/streaming-api-guide.md`)
```
dominic-mcallister-logdna: I think this should be 11

dominic-mcallister-logdna (approved):
lgtm!

Shearerbeard (approved):

teriyakichild (approved):
🎉 🎉 🎉

PR #33, opened 2026-03-25
feat(tools): add support for client-side tools

justintime4tea:
This adds the ability for clients to submit a list of client-side tools, via the OpenAI chat completions request spec, and have them "passthrough" to the client instead of being executed "server-side".

Ref: LOG-23583


Shearerbeard (dismissed):
This looks good - as it rebased orchestration we should probably consider organizing this into some of the library types splitting and reorganization - streaming/types.rs already needed a split. I didn't review the docs heavily but this should be noted as a use at your own risk for server side enterprise deployments - exposing client side tools to your agent thats public facing (like our web deployment) could be an attack vector

justintime4tea:
These changes have been added to the PR which adds the aura-cli - the aura-cli is currently the only user of client-side tools.

https://github.com/mezmo/aura/pull/60

PR #41, opened 2026-04-03
fix(orchestration): route MCP progress events to SSE in orchestration…

teriyakichild:
… mode

Introduce OrchestratorFactory, a lightweight StreamingAgent that defers Orchestrator construction to stream(). Previously, Orchestrator::new() ran twice per request — once at setup (factory) and once inside the tokio::spawn that runs orchestration — because &self can't move into the task. The factory's MCP clients received the HTTP request ID but the inner orchestrator's did not, silently dropping aura.progress events. The double construction also created duplicate persistence directories and wasted MCP connections.

OrchestratorFactory stores only AgentConfig and two Arc<RwLock> bridges: one to pass the request ID into the spawned task, and one to expose the inner orchestrator's MCP manager for client-disconnect cancellation.

ref: LOG-23565


Shearerbeard:
Closed - fixed in mshearer/LOG-23565-factory-refactor #53

PR #43, opened 2026-04-03
ci(deploy): add multi-config directory support

teriyakichild:
CONFIG_PATH now points to the /etc/aura/ directory in both the Helm chart and Razee MustacheTemplate. The Helm chart's config.content accepts either a string (backward compatible, rendered as config.toml) or a map of filename to TOML content for the config directory pattern. Added DEFAULT_AGENT env var for selecting the default agent when multiple configs are loaded.

ref: LOG-23574


mdeltito (approved):
LGTM

PR #45, opened 2026-04-06
chore: update persistence config

dominic-mcallister-logdna:
summary: swapped out the artifacts config from memory_dir to execution_persistence_base_path to be a bit more concise. The previous memory_dir and memory_path are still supported. Updated all docs and examples. Also cleaned up some duplicate directory creation code in persistence.

ref: LOG-23545

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,7 +256,7 @@ max_planning_cycles = 3
 tools_in_planning = "summary"
 allow_direct_answers = true
 allow_clarification = true
-memory_dir = "/tmp/orchestration-memory"
+execution_memory_base_path = "/tmp/orchestration-memory"
 
 [orchestration.worker.operations]
 description = "Operational analysis and diagnostics"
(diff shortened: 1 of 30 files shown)


dominic-mcallister-logdna:
~The main orchestration build was failing due to commit linting, so you'll see some commit message amendments by me. Hence the 3x commits via the timeline.~ Mike force-pushed commits with updated lint to the feature/orchestration-mode branch, so I rebased on it

justintime4tea:
I think "execution_persistence_base_path" is a bit verbose. I like the intent behind the change, to make it more readable, but I feel "memory" is more concise and readable. I know part of LOG-23545 was to figure out whether memory_dir is the "clearest name" but, fwiw, between the two I'm leaning towards memory_dir.

dominic-mcallister-logdna:
> I think "execution_persistence_base_path" is a bit verbose. I like the intent behind the change, to make it more readable, but I feel "memory" is more concise and readable. I know part of LOG-23545 was to figure out whether memory_dir is the "clearest name" but, fwiw, between the two I'm leaning towards memory_dir.

Its definitely verbose, but I'm a fan of verbosity for clarity. When I read "memory_dir", I take that to mean temporary storage that is completely ephemeral and/or readonly to the process only. This directory is used for execution and continuation while also hanging around if we need to debug any issues that arise.

Shearerbeard:
> > I think "execution_persistence_base_path" is a bit verbose. I like the intent behind the change, to make it more readable, but I feel "memory" is more concise and readable. I know part of LOG-23545 was to figure out whether memory_dir is the "clearest name" but, fwiw, between the two I'm leaning towards memory_dir.
> 
> Its definitely verbose, but I'm a fan of verbosity for clarity. When I read "memory_dir", I take that to mean temporary storage that is completely ephemeral and/or readonly to the process only. This directory is used for execution and continuation while also hanging around if we need to debug any issues that arise.

Specificity is nice but removing memory from the name is less than ideal. The idea of memory is pretty common in agent harnesses and file based memory is definitely the most common and well understood variant.

dominic-mcallister-logdna:
After considering, we're going to leave the `memory_dir` name in place as-is.

PR #46, opened 2026-04-06
chore(build): Create custom check

gjanco:
When jenkins checks Github and the PR is from an External source, Jenkins is setting up different checks than internal PRs.  This makes it difficult to enforce checks.  This change will create a custom check to enforce for both internal and external PRs.

Ref: LOG-23558

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -87,7 +87,7 @@ pipeline {
       }
     }
 
-    stage('Test') {
+    stage('Test Suite') {
       when {
         beforeAgent true
         not {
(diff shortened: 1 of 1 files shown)


justintime4tea (approved):

Jenkinsfile
```diff
@@ -87,69 +87,98 @@ pipeline {
...
+                          sh 'npx semantic-release --dry-run --no-ci'
+                        }
+                      }
+                    }
+                  }
+                )
+              } catch (Exception e) {
+                failed = true
+                failureText = e.getMessage() ?: 'See build logs for details.'
+                throw e
```
jmoses, line 157: Why this way (try/catch) instead of with a `post` action on failure?  Or you could do a `post { always {` and set the status using the `currentBuild.currentResult` value.

PR #48, opened 2026-04-07
chore(config): moved llm specific configs

dominic-mcallister-logdna:
summary: some llm specific configurations were either duplicated or inappropriately setup on the agent and passed down. Moved max_tokens, context_window, reasoning_effort, and additional_params to the llm config itself. Deleted the duplicated llm config classes in favor of a single class. Moved the lenient_int serde helpers to their own file. Updated model_context_limit to u64 to match RIG.

ref: LOG-23546

### DIFF
--- a/configs/math-orchestration-claude-thinking.toml
+++ b/configs/math-orchestration-claude-thinking.toml
@@ -19,6 +19,8 @@
 provider = "anthropic"
 api_key = "{{ env.ANTHROPIC_API_KEY }}"
 model = "claude-sonnet-4-5-20250929"
+temperature = 1.0
+additional_params = { thinking = { type = "enabled", budget_tokens = 8000 } }
 
 [agent]
 name = "Math Orchestrator (Thinking)"
@@ -31,8 +33,6 @@ ROUTING GUIDANCE:
 - Multi-step calculations or mixed domains: create a plan with workers
 - Vague or ambiguous requests: ask for clarification
 """
-temperature = 1.0
-additional_params = { thinking = { type = "enabled", budget_tokens = 8000 } }
 
 [mcp]
 sanitize_schemas = false
--- a/crates/aura/src/config.rs
+++ b/crates/aura/src/config.rs
@@ -3,59 +3,9 @@
 use crate::orchestration::OrchestrationConfig;
 use crate::tool_wrapper::{ToolCallContext, ToolWrapper};
 use serde::{Deserialize, Serialize};
+use std::fmt;
 
-/// Serde helpers for accepting whole-number floats (e.g. `8000.0`) as integers.
-///
-/// Helm's YAML parser represents all numbers as Go float64, so `toToml`
-/// renders `max_tokens = 8000.0` instead of `8000`. Rather than fixing this
-/// in Helm templates, we accept both forms on the Rust side.
-pub mod lenient_int {
-    use serde::{Deserialize, Deserializer};
-
-    /// Deserialize a value that may be either an integer or a whole-number float.
-    pub fn deserialize_option_u32<'de, D: Deserializer<'de>>(
-        deserializer: D,
-    ) -> Result<Option<u32>, D::Error> {
-        Option::<f64>::deserialize(deserializer)?
-            .map(|f| float_to_int(f, "u32"))
-            .transpose()
-    }
-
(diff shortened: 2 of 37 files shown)


configs/math-orchestration-claude-thinking.toml
```diff
@@ -32,7 +33,6 @@ ROUTING GUIDANCE:
 - Vague or ambiguous requests: ask for clarification
 """
 temperature = 1.0
```
Shearerbeard: This one is left over - temperature is a model parameter for the LLM

Shearerbeard (changes requested):
Overall looks good but I found a few regressions and have found a few things out of sync.

Ollama defaulting to None for additional params breaks our ability to customize Ollama models - the config for ollama is very free form and unpredictable model to model - I propose we pull out the hard coded enforced fields like num_ctx, num_predict and stop merging - Ollama can just take one flat additional_params configurable by the user. Users of ollama should be familiar with the unique config patterns. This can be just a loose serde Value.

To avoid issues with porting configs over with fields that did not move we should consider moving AgentConfig to #[serde(deny_unknown_fields)] to AgentConfig to fail on unknown config values to give the user clear feedback that they may be silently dumping a field they think is guiding the orchestration events.

Some config updates that claude caught:

▎ Three configs still have additional_params under [agent] — silently ignored, never read:
  ▎ - configs/math-orchestration-qwen35.toml:25 — thinking disable not applied
  ▎ - configs/math-orchestration-qwen3.toml:25 — thinking disable not applied
  ▎ - configs/math-orchestration-opus-bedrock.toml:27 — thinking mode not applied
  ▎
  ▎ Move each to [llm.additional_params...].
▎ Commented max_tokens: 2000 example is under the agent: section. Should be under llm:.

Some comments that claude caught:

  ▎ Four stale comments still say "agent-level" after the move:
  ▎ - crates/aura/src/builder.rs:232 — "reasoning_effort + agent-level params"
  ▎ - crates/aura/src/builder.rs:427 — "Ollama-specific + agent-level"
  ▎ - crates/aura/src/orchestration/orchestrator.rs:2150 — "reasoning_effort + agent-level"
  ▎ - crates/aura/src/orchestration/orchestrator.rs:2287 — "Ollama-specific + agent-level"
  ▎
  ▎ Should say LLM-level.

It might also be worth whipping up a quick markdown migration document.

docs/breaking-changes/20260410-agent-llm-toml-configuration.md
justintime4tea, line 1: I'm not exactly sure where best to put this but I feel reserved putting it at the root of the project. 

- Does it make sense under the "docs" folder? 

- Does it make sense to make a new folder either at root or within "docs"?

Whether we have only one or two of these, or we have a lot, I feel the root is an awkward place for this to go.

Thoughts?

dominic-mcallister-logdna: Maybe a "changelog" folder under docs to hold major changes?

Shearerbeard: I like the idea of having a structured way to document breaking changes across versions

dominic-mcallister-logdna: I went ahead and just called the folder "breaking changes" so its explicit.

crates/aura/src/config.rs
```diff
@@ -279,6 +309,51 @@ impl LlmConfig {
...
+            LlmConfig::OpenAI { context_window, .. }
+            | LlmConfig::Anthropic { context_window, .. }
+            | LlmConfig::Bedrock { context_window, .. }
+            | LlmConfig::Gemini { context_window, .. }
+            | LlmConfig::Ollama { context_window, .. } => *context_window,
+        }
+    }
+
+    /// Get additional_params regardless of provider.
+    /// Note: Ollama's additional_params is a HashMap, while others use serde_json::Value
```
Shearerbeard: Special handling comment no longer true

Shearerbeard (commented):
One inline comment and a realization that the AgentSettings.temperatures field is currently orphaned in usage and can just be dropped.

Shearerbeard (approved):

PR #52, opened 2026-04-13
Mshearer/log 23514 coordinator driven loop

Shearerbeard:
This is a step towards softening planning for our orchestrated flows while simplifying code. Planning is overplanning and way to prescriptive - evaluation and re-plan are a bit loose and context poor. This leads to odd behavior if our coordinator doesn't know enough about the problem to effectively one shot the plan. This lets the agent be more "intelligent" and so far has no regressions for well scoped out one shot plans but huge advantages in discovery or efficient planning failure recovery. The changes really help with the use case of explore first and solve one step of the problem at a time vs having to steer the agent into a multi phase plan ahead of time which ends up quite wasteful.

Detailed Summary (lots of work so generated by Claude - edited by me)

    Replaces the rigid evaluate→replan loop with a coordinator-driven iterative delegation model. The coordinator now drives the entire orchestration lifecycle — planning, reflection, and
    re-delegation — as a single continuous agent turn, eliminating the previous multi-phase architecture where separate evaluation and synthesis agents made disconnected decisions.

    Key Changes

    1. Coordinator-driven loop - The coordinator handles planning, reflection on results, and re-planning in a single continuous conversation rather than handing off to separate
    evaluation/synthesis agents between iterations. This gives the coordinator full context across iterations.
    2. Worker self-assessment - Workers now report structured self-assessments (Objective/Result/Process) as part of their output.
    Workers have the best context on whether their task succeeded then coordinator judges how usable and what to do with the output (a step toward further clarification of coordinator as judge - more research and testing needed here).
    3. Reflection redesign - The reflection prompt was redesigned to present completed/blocked/failed task results back to the coordinator with categorization, replacing the old evaluation
    scoring. The coordinator decides what to re-delegate based on actual results, not numeric quality scores (old eval step had limited context to make this decision and only score synthesized answer without "smart" coordinator assessment).
    4. Routing neutralization - Routing tool descriptions were made context-aware to prevent the coordinator from re-routing (direct answer, clarification) during replan iterations when it
    should be re-delegating work (this one actually dropped a ton of llm calls an iterations from the loop).
    5. Evaluation made passive - The evaluation agent is now passive/optional (observability only), no longer gating replan decisions. Quality scores are still emitted in SSE events but don't
     drive the loop (this may change but is a step in the right direction - if we do score we need to make sure we have the right context and success metric to do so).
    6. Single-task synthesis skip - When only one task exists, its result is returned directly without a separate synthesis LLM call.
    7. Worker preamble softened for transform tasks - Workers that transform/summarize/analyze provided context (rather than calling tools) no longer get overly aggressive "you must use
    tools" guidance.
    8. Dead synthesis iteration template removed - Cleaned up vestigial template code from the old multi-phase synthesis approach.

    Tickets Subsumed

    - LOG-23539 (replan reuses completed-but-empty task results) - solved by worker self-assessment + reflection categorization
    - LOG-23540 (Direct/Clarification during replan fake single-task) - solved by single-task synthesis skip + routing neutralization
    - LOG-23611 (eval truncation mismatch) - no longer relevant since eval is passive
    - LOG-23612 (transform-task worker conflict) - solved by softened worker preamble


Shearerbeard:
Some further clarification of "free the coordinator" tasks that this PR does not solve yet to give some realistic expectations for testing:

- LOG-23617: Synthesis/phase continuation coordinators receive wrong preamble claiming routing tools they don't have
    - LOG-23618: get_conversation_context tool is wired to workers but never called (0 uses across all E2E), architecturally unnecessary with coordinator-driven delegation
    - LOG-23619: Framework prompts hardcode math-specific constraints ("literal number", "do NOT call any tools") that block non-math use cases like SRE
    - LOG-23620: Session context injects only 200-char result previews with no guidance to use read_artifact for full prior-turn results
    - LOG-23621: Tool-call JSON artifacts are write-only, no readback mechanism for follow-up queries about prior tool behavior
    - LOG-23616: PlanningResponse::Orchestrated variant and phase infrastructure are dead code (~650 lines) now that CreatePlanTool produces StepsPlan exclusively

Shearerbeard:
I mentioned code simplification/reduction in the description even though the diff shows a small expansion - the savings come later with a few things that can be removed in the related work mentioned above and in the base ticket.

Shearerbeard:
Closing this PR in lieu of some revisions and movement to a more wholistic solution. Superseded by LOG-23633 in an effort to further simplify and add even more freedom to the coordinator

PR #53, opened 2026-04-13
Mshearer/log 23565 factory refactor

Shearerbeard:
Fixes LOG-23565 reported by @teriyakichild and refactors the structural issue at the heart of the problem.

Summary:

MCP progress notifications were silently dropped in orchestration mode because the inner Orchestrator (created per-request inside tokio::spawn) never received the HTTP request ID needed to
  route notifications to the correct SSE subscriber. This also caused duplicate persistence directories and MCP connections per request.

  Replaced the Orchestrator's StreamingAgent impl with a zero-state OrchestratorFactory that defers construction to stream(), where request_id flows as a direct parameter. Removed
  set_mcp_request_id/clear_mcp_request_id from the StreamingAgent trait to eliminate the temporal coupling they introduced. Added an orchestration integration test using mock-mcp's
  task_with_progress tool to verify progress event routing end-to-end.

### DIFF
--- a/crates/aura-web-server/tests/orchestration_events_test.rs
+++ b/crates/aura-web-server/tests/orchestration_events_test.rs
@@ -479,6 +479,84 @@ async fn test_division_error_emits_task_failure() {
 }
 
 // ---------------------------------------------------------------------------
+// MCP progress notifications in orchestration mode (LOG-23565)
+// ---------------------------------------------------------------------------
+
+/// Verifies that MCP progress notifications route through orchestration SSE.
+///
+/// Query triggers the progress_task worker which calls task_with_progress on
+/// mock-mcp. This tool emits MCP notifications/progress as it runs. The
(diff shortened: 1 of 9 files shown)


crates/aura-web-server/tests/orchestration_events_test.rs
```diff
@@ -478,6 +478,84 @@ async fn test_division_error_emits_task_failure() {
     }
 }
 
+// ---------------------------------------------------------------------------
+// MCP progress notifications in orchestration mode (LOG-23565)
```
dominic-mcallister-logdna, line 482: I think I'd remove the LOG-XXXXX reference as that's not useful outside Mezmo internal

Shearerbeard: Your right - when we cut over we need to take a quick pass at some of this.

PR #59, opened 2026-04-21
feat(config): nest llm under agent with per-worker overrides

justintime4tea:
Move the top-level `[llm]` TOML table under `[agent.llm]` and allow `[orchestration.worker.<name>.llm]` to override the agent LLM. Workers without an explicit override inherit `[agent.llm]`; each worker's resolved `context_window` now populates Agent.context_window so per-worker context budgets (LOG-23439) can be computed from it.

The loader pre-parses raw TOML and emits a self-contained migration error whenever a legacy top-level `[llm]` table is present, so misconfigured configs fail fast with a clear message instead of a cryptic serde error.

All shipped TOML configs, examples, README, streaming/ollama docs, and unit tests are migrated to the new shape.

WARNING: (breaking change) `[llm]` no longer parses at the top level. Move it to `[agent.llm]` (and `[llm.additional_params]` to `[agent.llm.additional_params]`). See
docs/breaking-changes/20260421-llm-under-agent.md.

Ref: LOG-23606

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,11 @@ Key capabilities:
   - [Usage](#usage)
     - [Web API Server](#web-api-server)
   - [Configuration](#configuration)
+    - [Multiple Agents](#multiple-agents)
+    - [Configuration Sections](#configuration-sections)
     - [Orchestration](#orchestration)
+      - [Orchestration fields](#orchestration-fields)
+      - [Worker fields (`[orchestration.worker.<name>]`)](#worker-fields-orchestrationworkername)
     - [Ollama](#ollama)
     - [Observability](#observability)
   - [Docker Deployment](#docker-deployment)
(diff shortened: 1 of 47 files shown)


dominic-mcallister-logdna:
I think we should add one or two example configs under the configs folder that demonstrate the usage of per-worker LLM configs.

I also think in the documentation it should be explicit that a per-worker LLM config is to be a complete config, not partial property overrides.

justintime4tea:
> I think we should add one or two example configs under the configs folder that demonstrate the usage of per-worker LLM configs.

Added `examples/complete/kubernetes-sre-orchestrated-multi-model.toml` which basically adds orchestration and multiple models to the `examples/complete/kubernetes-sre.toml` example.

> I also think in the documentation it should be explicit that a per-worker LLM config is to be a complete config, not partial property overrides.

This has been addressed :pray: thanks for the call-out/catch.

Shearerbeard (approved):

PR #60, opened 2026-04-21
feat(cli): add aura-cli crate with HTTP + standalone backends

justintime4tea:
Interactive terminal REPL with conversation persistence and streaming responses. Ships alongside a new lightweight aura-events crate for shared SSE event types between the CLI and web server.

- Two backends: HTTP (connects to aura-web-server) and standalone (--standalone --config, builds agents in-process)
- --model works in both modes (HTTP: starting model; standalone: matches agent.name/alias from config)
- --system-prompt is standalone-only (append/replace the agent's TOML prompt); HTTP mode rejects it unless --force
- --force bypasses non-critical warnings in query mode
- Conflict resolution prompts when resuming conversations with mismatched --model or --system-prompt
- Local tools (Shell, Read, ListFiles, Update, SearchFiles, FindFiles, FileInfo) are standalone-only, with a permission system via .aura/settings.json
- Conversation persistence under ~/.aura/conversations/

Additionally, this adds client-side tool support. This is off by default, has a BIG warning in server and CLI README, and requires opting-in both in the agent config TOML and the cli arg --enable-client-tools. This only works with single agent configurations - client tools are never supported with orchestration.

Ref: LOG-23587


teriyakichild:
@justintime4tea This is awesome.  Great work.  My only feedback:
* A new session is created on every user turn when using the aura-cli to connect to a remote Aura instance.  I fixed this in my local copy by passing the UUID from the conversation store in the `X-Chat-Session-Id` header.
* An option to disable LLM-based title generation would be useful for one-shot inference and slower local models, where it adds a bunch of overhead.

PR #63, opened 2026-04-22
chore(sse): consolidate properties and cleanup docs

dominic-mcallister-logdna:
summary: updated information relating to orchestration events
- consolidated like properties of orchestration events to flat objects
- updated tracing statements from info to debug during orchestration event construction
- swapped task_count for tasks, maintaining the list of task descriptions so the consumer has an idea of what is going to take place
- updated max tool result length from 100 to 1000 to provide more context into tool call actions
- update documentation for streaming events to include missing events, improved example formatting, and swapped ascii charts for mermaid variants.


Shearerbeard (approved):
This looks good for me - I played around with the flattened outputs for SSE and I think the only significant change is task_count being migrated to Vec<String> in PlanCreated. @justintime4tea should review for CLI implications.

Issue #70, opened 2026-04-24
Orchestrator marks soft-failed worker tasks as success=true


Shearerbeard:
## Problem

In `execute_task`, the `Ok(result_str)` arm after rig's multi_turn loop marks `TaskStatus::Complete, success=true` whenever the loop terminates without a rig-level error. The result text is never inspected. A worker that enters a duplicate call loop and writes "I was unable to complete the task" is indistinguishable from genuine success.

This means:
- `build_task_context` embeds soft-failure text as `PRIOR_WORK` for dependents
- `ready_tasks()` unblocks dependents that should be blocked
- The coordinator has no structural signal forcing a replan

## Evidence

Live SSE captures from SRE integration testing show `"success": true` on tasks where the worker explicitly reported inability to complete, counted as `task_failed_count: 0` in analysis.

## Solution

After `stream_and_forward` returns, check the escalation flag set by the duplicate-call guard (#69). If set, run the worker response through `maybe_create_artifact` (inline if short, artifact pointer if long) and return `Err` so the task is marked `Failed` and dependents are blocked in the DAG. The coordinator sees the worker's findings via the existing reflection prompt and can `read_artifact` for long outputs.

## Related code

- `crates/aura/src/orchestration/orchestrator.rs` — `execute_task` flag check after `stream_and_forward`
- `crates/aura/src/orchestration/types.rs` — `ready_tasks()` blocks dependents on `TaskStatus::Failed`

Issue #69, opened 2026-04-24
[BUG] Duplicate-call guard uses surrender phrasing and hides tool output


Shearerbeard:
## Problem

The duplicate-call guard has two coupled issues:

1. **Surrender-inducing phrasing + output hiding.** When a tool call repeats identically, the guard returns a `ToolError` ending with "Write your final answer now. Do not call this tool again." The phrasing tips the worker into surrender — models respond with "I was unable to complete the task" despite having turn budget remaining. Returning `Err` at `validate_args` means the tool is never called and the real output is hidden.

2. **Over-firing on legitimate retries.** The guard counts ANY repeated `(tool, args, result)` triple, including transient MCP errors (503, timeouts, rate limits). An LLM retrying a temporarily-unavailable tool is penalized the same as one stuck in a schema-error loop.

The surrender output is then mislabeled `success=true` by the downstream grading path — see "Orchestrator marks soft-failed worker tasks as success=true" (filed separately).

## Solution

Two-stage escalation via annotation on the real tool output (never hidden):
- `nudge_threshold` — appends `[DUPLICATE_CALL_GUIDANCE]` with observation framing and alternatives
- `block_threshold` — appends `[DUPLICATE_CALL_ABORT]` and sets an `Arc<AtomicBool>` escalation flag

Error-kind-aware counting using `CallOutcome` from #68: `GeneralToolError` does not increment the counter (legitimate retries). `Success` and `SchemaError` with identical output do.

Per-fingerprint tracking via `HashMap<CallFingerprint, CallState>` prevents ping-pong evasion where alternating tools reset each other's counters.

## Related code

- `crates/aura/src/orchestration/duplicate_call_guard.rs` — guard implementation
- `crates/aura/src/prompts/duplicate_call_guidance.md` — nudge template
- `crates/aura/src/prompts/duplicate_call_abort.md` — abort template
- `crates/aura/src/orchestration/orchestrator.rs` — `AgentWithPreamble` escalation flag plumbing

PR #71, opened 2026-04-24
Duplicate guard soft failure grading

### LINKED ISSUE #68: Preserve structured MCP error-kind through the tool wrapper chain
## Problem

`mcp_response.rs::extract_tool_result` stringifies every MCP response before it reaches the tool wrapper chain. For error responses, the function collapses schema-validation failures, rate limits, timeouts, and server-internal errors into a single opaque string class prefixed with `"Tool returned an error: "`. Downstream consumers that need to make semantics-aware decisions have no structured signal to work with.

## Solution

Introduced `CallOutcome` enum (`Success` / `SchemaError` / `GeneralToolError`) in `crates/aura/src/mcp_response.rs`. Threaded through `ToolWrapper::transform_output` so downstream consumers can branch structurally instead of parsing string prefixes. Existing callers use `into_prefixed_string()` for backward compatibility.

## Related code

- `crates/aura/src/mcp_response.rs` — `CallOutcome` enum and `extract_tool_result`
- `crates/aura/src/tool_wrapper.rs` — `transform_output` signature carries `&CallOutcome`
- `crates/aura/src/mcp_streamable_http.rs` — callers use `into_prefixed_string()` at boundary

Shearerbeard:
Enhanced the looping guard to explicitly fail dependent worker tasks and give a clearer signal to the coordinator as to why a task was force killed. Also opens up the plumbing for other worker escalates to coordinator workflows.

closes #68 #69 #70 

LOG-23734
LOG-23735
LOG-23743

### DIFF
--- a/crates/aura/src/orchestration/orchestrator.rs
+++ b/crates/aura/src/orchestration/orchestrator.rs
@@ -473,8 +473,8 @@ impl Orchestrator {
             self.tool_call_observer.clone(),
             task_id,
         ));
-        let nudge = self.config.duplicate_call_nudge_threshold.unwrap_or(3);
-        let block = self.config.duplicate_call_block_threshold.unwrap_or(5);
+        let nudge = self.config.duplicate_call_nudge_threshold;
+        let block = self.config.duplicate_call_block_threshold;
         let escalation_flag = Arc::new(std::sync::atomic::AtomicBool::new(false));
         let duplicate_guard = Arc::new(DuplicateCallGuard::new(
             nudge,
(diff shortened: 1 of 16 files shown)


crates/aura/src/orchestration/orchestrator.rs
```diff
@@ -469,11 +473,14 @@ impl Orchestrator {
             self.tool_call_observer.clone(),
             task_id,
         ));
-        let max_dup = self
-            .config
-            .max_consecutive_duplicate_tool_calls
-            .unwrap_or(1);
-        let duplicate_guard = Arc::new(DuplicateCallGuard::new(max_dup));
+        let nudge = self.config.duplicate_call_nudge_threshold.unwrap_or(3);
+        let block = self.config.duplicate_call_block_threshold.unwrap_or(5);
```
dominic-mcallister-logdna: Instead of defining the default of 3 or 5 here, we can do so on the struct itself?

Shearerbeard: Good call and yes.

dominic-mcallister-logdna (approved):

PR #77, opened 2026-04-27
feat(aura): add VFS-agnostic MemoryFS for orchestration memory

henryjandrews:
## Summary

Adds a VFS-agnostic MemoryFS layered over `orchestration.artifacts.memory_dir`, plus coordinator-only tools and an internal writer so Aura's coordinator can consult durable orchestration memory before routing. Backend-agnostic: works with local disk, Archil, NFS, Docker volumes, or K8s PVCs as a mounted path.

## Changes

**Config**
- New `[orchestration.memory]` section (disabled by default) with `enabled`, `root_dir`, `max_read_bytes`, `max_search_results`
- Root resolves to `memory.root_dir`, falls back to `artifacts.memory_dir`; fails clearly if enabled with neither
- Mirrored in `crates/aura-config`

**MemoryFS layout (Markdown-first, OpenChronicle-inspired)**
- `memory/index.md`, `event-YYYY-MM-DD.md`, `worker-*.md`, `failure-*.md`, etc.
- YAML frontmatter, append-only timestamped entries, atomic writes
- Run artifacts remain in existing session/run layout

**Coordinator-only read tools**
- `list_memories`, `read_memory`, `search_memory`, `recent_memory`, `memory_shell`
- Native Rust search (fixed-string, regex, case-insensitive) with result caps and truncation reporting
- `memory_shell` is a read-only DSL (`pwd`, `ls`, `cat`, `head`, `tail`, `stat`, `find`, `grep`, `query`) — no subprocess execution, no pipes/redirects/writes
- Virtual paths only; traversal and host-path escapes rejected

**Internal `MemoryWriter`**
- Invoked after `write_run_manifest` when memory is enabled
- Appends timeline + per-worker success/failure entries, regenerates `index.md`
- Failures are warning-only; no LLM classification in v1

**Coordinator behavior**
- Preamble + planning prompt updated to consult memory before routing on prior-reference, follow-up, or underspecified queries
- Memory tools registered before routing tools; memory calls do not trigger early routing exit
- Recalled facts embedded directly into worker task descriptions (workers remain memory-blind in v1)


henryjandrews:

henryjandrews:
recheck

henryjandrews:
recheck

henryjandrews:

henryjandrews:
recheck

sdelmas:
> **CTO-agent review on behalf of [MEZA-6](https://mezmo.atlassian.net/browse/LOG-23702) PRISM Memory consolidation.** Posting as Sven because the CTO agent doesn't have its own GitHub identity; technical content authored by the agent and approved by CEO before cross-posting.

@henryjandrews — flagging two correctness items from a CTO-side review of this PR at SHA `af961c31`. Want these on your radar before merge; happy to pair on triage if useful.

## Blocker: depth-1 hardcode in `stream_and_collect`

`MemoryFs::stream_and_collect` (in `crates/aura-orchestrator/src/memory_fs.rs`) hardcodes a depth limit of `1`, which means every new memory tool that walks past the top of `memory/` returns empty under the coordinator's documented flow. `list_memories`, `search_memory`, `find` in `memory_shell`, and `recent_memory` are all affected for any layout deeper than the root — i.e. the moment events, workers, and failures land in their own dated/keyed files (which is the layout the README itself shows). Symptoms: tools succeed, return zero results, coordinator falls through to "no prior memory found" and routes blind.

This is a single-line fix (parameterize the depth, default to the configured `max_search_results` or an explicit traversal cap), but it's the difference between memory being usable and silently absent. Treating this as a merge-blocker for the PR.

## Same severity, less obvious: `read_memory.tail_n`

`read_memory` with `tail_n` reads the **first** N bytes of the file, not the last — so any "recent memory" / "tail of the timeline" call returns the oldest content, which is the opposite of intent. No error, no warning, just wrong data quietly flowing into routing decisions. Same severity tier as the depth-1 issue, just harder to spot in review because the call signature looks right.

## Smaller follow-up (non-blocking, post-merge OK)

YAML frontmatter would benefit from a `schema: v1` field so the writer/reader can version the format independently of the binary. Cheap to add now, expensive to retrofit once customer environments have memory on disk.

## Other items I noted

A few additional issues (symlink escape from memory root, per-path search caps composing past `max_search_results`, a couple of others) are tracked CTO-side under MEZA-6. Happy to scaffold those as child issues *after* you've triaged the two above — don't want to pre-create noise on your queue.

Context: full review notes live on the PRISM Memory architecture doc (MEZA-6), which also captures the strategic framing from your "Unified Agent Memory" Notion spike — the two-directory `sessions/` vs `knowledge/` model, HITL promotion gate, and MCP-first contract. PR #77 reads cleanly as the working-memory substrate layer in that picture; A1/A4 just need to land before it's coordinator-usable.

PR #81, opened 2026-04-30
chore(setup): rework ci and release setup

esatterwhite:
This replaces release handling that was being done inline in the
Jenkinsfile to something that exists as part of the project.
Additionally did some work to simplify the Jenkins pipeline.
Removing the scripting workflow infavor of the simple decarative syntax.
There is a helper function that makes it easier to send and manage
custom checks to github. The are additional follow up tasks to improve
visibility into what is happening, but this mostly removes the
dependency on internal tooling.


esatterwhite:

.config.mk
```diff
@@ -0,0 +1,3 @@
+# Make Variable overrids go here
```
zeldanut: overrides (sp)

zeldanut (dismissed):
lgtm just a spelling nit

zeldanut (dismissed):

.config.mk
esatterwhite: fixed

zeldanut (approved):

PR #83, opened 2026-05-04
chore: fix flakey tests that write to env vars

justintime4tea:
There are several places where tests write/read from env vars. When this is done in parallel it can cause flakey tests. This update adds a crate wide lock for tests to get rid of the flakey-ness.

Ref: LOG-23806

### DIFF
--- a/crates/aura-config/src/config_test.rs
+++ b/crates/aura-config/src/config_test.rs
@@ -321,6 +321,7 @@ model = "gpt-4"
 
     #[test]
     fn test_environment_variable_placeholders() {
+        let _env_lock = crate::test_env_lock::lock();
         println!("\n=== TEST_ENVIRONMENT_VARIABLE_PLACEHOLDERS ===");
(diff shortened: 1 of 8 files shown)


Shearerbeard (changes requested):
I had to add the lock to the test configs test to get it to run a full test suite locally. We can either fix that or move the test config tests back to integration tests since the rest of the tests that require a key go there.

justintime4tea:
> I had to add the lock to the test configs test to get it to run a full test suite locally. We can either fix that or move the test config tests back to integration tests since the rest of the tests that require a key go there.

I added the lock - I found a few other set_env cases but they use test-specific, and unique, env vars and aren't an issue. Not sure how I missed this one the first pass 😅 

Appreciate you catching this 🙏

Shearerbeard (approved):

Shearerbeard:
@justintime4tea runs fine locally after your most recent fix

PR #84, opened 2026-05-05
build: docker based test setup via nextest

esatterwhite:
updates the make setup to leverage nextest and grcov to output
structured report formats (cobertura,html, junit). additionally
adds docker helper for running most all make targets inside a docker
container. This makes it easier to get up and running with out needing
to setup your local machine to mirror our expected setup. In this sense
-- the make file serves as both CI/Local for most all things.
Setting `CI=true` sets commands to run in a way that is suitable for CI
with reports, `ENABLE_DOCKER=true` makes the commands run in a docker
container where applicable


.makefiles/aura.mk
```diff
@@ -0,0 +1,14 @@
+
+BBUILD_TAG ?= 1
```
jmoses: BBUILD?

.makefiles/aura.mk
```diff
@@ -0,0 +1,14 @@
...
+
+.PHONY:start
+start:: ## Start the docker compose setup
+	docker compose -f compose/base.yml -f compose/dev.yml up
```
jmoses: should this be `up --wait -d`?

esatterwhite: This is just a stub out at the moment, I'm going to fix up this compose stuff in a bit. I'm just trying to get the reporting functional. Needs a PR for all of it to work right

Jenkinsfile
```diff
@@ -97,19 +99,56 @@ pipeline {
...
+                qualityGates: [[threshold: 1, type: 'TOTAL', criticality: 'FAILURE']]
+              )
+            }
+          }
+        } // end rustfmt
+
+        stage("Rust Lint") {
+          steps {
+            withChecks('clippy') {
+              sh "make lint-rust || true"
```
jmoses: if the exit code doesn't matter you can:

```
sh script: "make lint-rust", returnStatus: true
```

 ¯\\\_(ツ)_/¯

jmoses: or leave it to exit 1, and remove the quality gate on the record issues call

.makefiles/rust.mk
```diff
@@ -1,10 +1,115 @@
...
-.PHONY:lint-rust | .check-env-render
-lint-rust: ## lint rust code via clippy
-	cargo clippy $(if $(CI),-q,) --all-targets --all-features $(if $(CI),--message-format=json,) -- -D warnings $(if $(CI),> $(OUTPUT_DIR)/clippy.json,)
+CARGO_BIN_DIR ?= .bin
+NEXTEST_BIN = $(CARGO_BIN_DIR)/cargo-nextest
+LLVM_PROFILE_FILE ?= $(COVERAGE_DIR)/profile-%p-%m.profraw
+GRCOV_VERSION ?= v0.10.7
+GRCOV_BIN = $(CARGO_BIN_DIR)/grcov
+PATH := $(CARGO_BIN_DIR):$(PATH)
+RUSTFLAGS=--allow=warnings -Cinstrument-coverage
```
esatterwhite: We need to set this to `?='` empty and only set the rust flags during the coverage target

zeldanut (approved):
Lgtm

Shearerbeard (approved):
Runs great - thanks

PR #86, opened 2026-05-05
refactor(webserver): swap actix-web for axum

dominic-mcallister-logdna:
summary: in preparation of bringing in the A2A rust library, swapping the webserver to axum as its the main dependency web server framework of the A2A library. This'll save us from needing to jump through many hoops or run
multiple servers.

ref: LOG-23816

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,239 +3,6 @@
 version = 4
 
 [[package]]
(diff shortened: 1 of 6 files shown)


justintime4tea (approved):
Looks good! 👍

PR #94, opened 2026-05-08
feat(a2a): initial implementation

dominic-mcallister-logdna:
summary: bring in and implement a2a-rs crate.
Handles task generation and maintenance calling
internal aura SSE processing. Currently uses
an in-memory store to keep dependencies down.

ref: LOG-23822


crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+    ) -> Self {
+        Self {
+            app_state,
+            event_tx,
+            task_store,
+        }
+    }
+
+    fn resolve_config(&self) -> Option<aura_config::Config> {
+        if self.app_state.configs.len() == 1 {
```
kristof-mattei: This can be simplified to 

```
        if let [only] = &**self.app_state.configs {
            return Some(only.clone());
        }
```

That way we use the invariant 'there is only 1 item' and we're cloning that item.

kristof-mattei: Thinking about this some more: this block is an optimization.

If we remove it, it doesn't change the outcome of the function. The only moment where additional work would be done is in the case there is 1 item in the list, and the agent's alias doesn't match. In that case, we would take it anyway, see line 57.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+        &self,
+        message: Message,
+        auth: Option<AuthContext>,
+    ) -> HandlerResult<SendMessageResponse> {
+        // take all the message parts and concatenate the text parts together to form the full message text
+        let text: String = message
+            .parts
+            .iter()
+            .filter_map(|p| {
+                if let Part::Text { text, .. } = p {
```
kristof-mattei: What do we want to do if the caller sends a file or structured (JSON?) data? In this case we discard it.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+            })
+            .collect::<Vec<_>>()
+            .join("\n");
+
+        let task_id = Uuid::new_v4().to_string();
+        debug!(
+            "Invoking new task {} with requested chat: {}",
+            task_id,
+            text.clone()
+        );
```
kristof-mattei: Separate the text from the data, makes it easier for JSON logs:

```
event!(Level::DEBUG, task_id, text, Invoking new chat"); 
```

Equally, the clone here isn't needed.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+        // task is generated and ran in the background for the caller to subscribe
+        // to the events and get the response as it is generated
+        let task_store = self
+            .task_store
+            .get()
+            .ok_or_else(|| HandlerError::processing_failed("task store not initialized"))?
+            .clone();
+
+        let task = Task {
+            kind: "task".to_string(),
```
kristof-mattei: Use `to_owned()`, `to_string()` returns the a human readable output of the value. For `&str` to `String` it happens to match, but here you want an owned version of `&str`.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+                        &context_id,
+                        &event_tx,
+                        &e.to_string(),
+                    )
+                    .await;
+                    return;
+                }
+            };
+
+            let request_id = format!("a2a_{}", task_id.clone());
```
kristof-mattei: `clone()` not needed.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+            let cancel_token = CancellationToken::new();
+
+            let mut stream = match agent.stream(&text, vec![], cancel_token, &request_id).await {
+                Ok(s) => {
+                    debug!("stream was generated successfully: {}", request_id);
```
kristof-mattei: Same as above, separate log text from actual data.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+                        &event_tx,
+                        &e.to_string(),
+                    )
+                    .await;
+                    return;
+                }
+            };
+
+            let request_id = format!("a2a_{}", task_id.clone());
+            let cancel_token = CancellationToken::new();
```
kristof-mattei: THINK: Do we want to do anything with this token? Like store it somewhere to allow graceful shutdown?

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+                    return;
+                }
+            };
+
+            let mut chunk_index = 0u32;
+
+            while let Some(item) = stream.next().await {
+                match item {
+                    Ok(StreamItem::StreamAssistantItem(StreamedAssistantContent::Text(t))) => {
+                        debug!("stream content received: {} / {}", request_id, t.clone());
```
kristof-mattei: As above, separate, also, `.clone()` not needed.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+
+                        let artifact = Artifact {
+                            artifact_id: "response".to_string(),
```
kristof-mattei: `to_owned()`.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+                            name: Some("Response".to_string()),
+                            description: None,
+                            parts: vec![Part::text(t)],
+                            metadata: None,
+                            extensions: vec![],
+                        };
+                        let append = chunk_index > 0;
+                        let artifact_clone = artifact.clone();
```
kristof-mattei: This variable is not needed, we can clone the artifact itself on 215 and 219.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+                                        } else {
+                                            artifacts.push(artifact_clone.clone());
+                                        }
+                                    }
+                                } else {
+                                    task.artifacts = Some(vec![artifact_clone.clone()]);
+                                }
+                            })
+                            .await;
+                        let _ = event_tx.send(StreamResponse::ArtifactUpdate(
```
kristof-mattei: `let _` is considered harmful. 

If we want to `drop` the result immediately, it's better to be explicit: `let r: Type = ...; drop(r);`.

If we don't care about the result, it's better to `let _r: Type = ...`, because that way when the signature changes the code here breaks.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+                            TaskArtifactUpdateEvent {
+                                kind: "artifact-update".to_string(),
```
kristof-mattei: `to_owned()`.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+                                artifact,
+                                append: Some(append),
+                                last_chunk: Some(false),
+                                metadata: None,
+                            },
+                        ));
+                        chunk_index += 1;
+                    }
+                    Ok(StreamItem::StreamAssistantItem(StreamedAssistantContent::ToolCall(tc))) => {
+                        debug!("tool call received: {} / {}", request_id, tc.name);
```
kristof-mattei: Separate.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+                            parts: vec![],
+                            metadata: Some(serde_json::json!({
+                                "type": "tool_call",
+                                "id": tc.id,
+                                "name": tc.name,
+                                "arguments": tc.arguments,
+                            })),
+                            extensions: vec![],
+                        };
+                        let artifact_clone = artifact.clone();
```
kristof-mattei: I think we don't need the cloned variable here either.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+                                if let Some(artifacts) = &mut task.artifacts {
+                                    artifacts.push(artifact_clone.clone());
+                                } else {
+                                    task.artifacts = Some(vec![artifact_clone.clone()]);
+                                }
+                            })
+                            .await;
+                        let _ = event_tx.send(StreamResponse::ArtifactUpdate(
+                            TaskArtifactUpdateEvent {
+                                kind: "artifact-update".to_string(),
```
kristof-mattei: `to_owned().`

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+        let first = self.app_state.configs.first();
+        let name = first
+            .map(|c| c.agent.name.as_str())
+            .unwrap_or("Aura Agent")
+            .to_string();
+        let description = first
+            .map(|c| c.agent.system_prompt.as_str())
+            .unwrap_or("Aura AI agent")
+            .chars()
+            .take(200)
```
kristof-mattei, line 362: +1 for being UTF-8 safe!!

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,446 @@
...
+            documentation_url: None,
+            capabilities: AgentCapabilities {
+                streaming: Some(true),
+                push_notifications: Some(false),
+                extended_agent_card: Some(false),
+                ..Default::default()
+            },
+            supported_interfaces: vec![
+                AgentInterface {
+                    url: format!("{base_url}/v1/message:send"),
```
kristof-mattei: NIT: format with inline is the same color in the UI, and makes it harder to parse. Personally I prefer pulling it out as variables.

crates/aura-web-server/src/main.rs
```diff
@@ -218,7 +221,56 @@ async fn run() -> std::io::Result<()> {
...
+                .and_then(|s| s.strip_prefix("Bearer "))
+                .unwrap_or("")
+                .to_string(),
+            metadata: Some(serde_json::Value::Object(header_map)),
+        })
+    })
+    .rpc_path("/v1/rpc")
+    .rest_prefix(Some("/v1"));
+
+    let _ = event_tx_cell.set(a2a_server.get_event_sender());
```
kristof-mattei, line 282: We pass in this cell at #232, why don't we initialize it as part of the creation of the Arc? That way we don't need the set here, and we can get rid of the `let _`.

crates/aura-web-server/src/main.rs
```diff
@@ -218,7 +221,56 @@ async fn run() -> std::io::Result<()> {
...
+            .filter_map(|(k, v)| {
+                v.to_str()
+                    .ok()
+                    .map(|val| (k.to_string(), serde_json::Value::String(val.to_string())))
+            })
+            .collect();
+        Some(AuthContext {
+            user_id: header_map.get("X-Auth-Subject-Id")
+                .and_then(|v| v.as_str())
+                .unwrap_or("anonymous")
```
kristof-mattei: Using `anonymous` as a moniker for a user-id that isn't authenticated is weird. Can we omit the user_id to indicate absence thereof?

crates/aura-web-server/src/main.rs
```diff
@@ -218,7 +221,56 @@ async fn run() -> std::io::Result<()> {
...
+                .to_string(),
+            access_token: header_map.get("Authorization")
+                .and_then(|v| v.as_str())
+                .and_then(|s| s.strip_prefix("Bearer "))
```
kristof-mattei: What about Token / anything else?

crates/aura-web-server/src/main.rs
```diff
@@ -229,6 +281,7 @@ async fn run() -> std::io::Result<()> {
         ))
         .with_state(app_state);
 
+    let app = aura_router.fallback_service(a2a_router);
```
kristof-mattei, line 304: Q: do we want to next the a2a under `/v1`? Since we have an agent-card we can completely separate that from the existing aura endpoints.

Shearerbeard: I would strongly agree we shouldn't nest this under the OpenAI style v1 root route.

crates/aura-web-server/src/a2a/handler.rs
dominic-mcallister-logdna: Great point, I'll add some validation.

crates/aura-web-server/src/main.rs
dominic-mcallister-logdna: You're right, this is weird. I went and added an env var to get the user_id if things are available, otherwise its going to be blank just because its required by the AuthContext. Since we've yet to use it, I think that'll be fine.

crates/aura-web-server/src/main.rs
dominic-mcallister-logdna: Good callout. I moved both the key and the strip value to env vars if we decide to use them.

crates/aura-web-server/src/main.rs
dominic-mcallister-logdna: I moved these to v2

crates/aura-web-server/src/a2a/handler.rs
dominic-mcallister-logdna: I found theres a cancellation token on the app state and used that. Things should be more graceful now.

crates/aura-web-server/src/a2a/handler.rs
dominic-mcallister-logdna: Validation added along with a unit test that'll spell the issue out directly to the user.

crates/aura-web-server/src/main.rs
dominic-mcallister-logdna: I didn't do a thing to this since its a chicken/egg thing. I need to define the Arc for the handler, but the handler need to access the event sender of the server. Or perhaps I'm misunderstanding things.

justintime4tea:
A couple thoughts, in general:

* Documentation — docs/a2a-implementation.md only lists /message:send, /tasks, /tasks/{id}, /tasks/{id}:cancel, and /rpc. It doesn't mention /message:stream or /tasks/{id}:subscribe, so consumers don't know the streaming surface exists.
* Tests — tests/a2a_test.rs only exercises :send and the polling endpoints. No test consumes the SSE stream.

crates/aura-web-server/src/a2a/handler.rs
```diff
@@ -0,0 +1,462 @@
...
+                obj.iter()
+                    .filter_map(|(k, v)| v.as_str().map(|s| (k.clone(), s.to_string())))
+                    .collect()
+            })
+            .unwrap_or_default();
+
+        let stream_shutdown_token = self.app_state.stream_shutdown_token.clone();
+
+        // run the task in the background and stream updates to the task store as they are generated
+        tokio::spawn(async move {
```
justintime4tea, line 148: There's a subtle ordering bug that affects both `/tasks/{id}:subscribe` and (to a lesser extent) `:stream`. Worth flagging here because the worst-case symptom is a client that hangs forever on a task that has actually completed. Essentially, this runtime task "block" here is async and we return Ok(Task) while this block is off executing - the underlying a2a-server-rs crate has some code that could lead to events being dropped and never recieved by consumers.

The "actual" fix needs to land upstream in the a2a-server-rs crate but there's a way we can address the issue locally (partially) while waiting for an upstream fix.

I'm going to draft up a PR with this branch as a base so I can describe the condition and proposed workaround.

justintime4tea:
**Multi-turn conversations via context_id don't reach the agent**

**A2A's conversation model is built around context_id:** clients send a sequence of message:send calls sharing one context_id, and the server is expected to treat them as a single ongoing conversation. Right now I don't think we're honoring that:
- [handler.rs:97-100](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L97-L100) — `context_id` is captured (defaulted to a fresh UUID if absent) and stored on the `Task`, but never used as a session key downstream.
- [handler.rs:149](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L149) — `session_id = Some(task_id.clone())`, so every `message:send` gets a brand-new session in Aura's eyes. `task_id` is unique per call by design [(handler.rs:89)](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L89), so two messages sharing a `context_id` still land in two unrelated sessions.
- [handler.rs:173](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L173) — `agent.stream(&text, vec![], …)` chat_history is hard-coded empty, so even the messages stored on prior tasks in the same context aren't being replayed.
- [handler.rs:125](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L125) — the task's own history only contains the current inbound message; we never look up or carry forward prior tasks in the same context.

**Impact:** any A2A client that sends a follow-up like "explain that more" or "now do it for the staging cluster" will get a fresh agent with no memory of the previous turn. For single-agent chat this just feels broken; for orchestration it's worse — the coordinator loses any plan/result context from the prior task.

- **OpenAI path in handlers.rs:**
  - (a) resolves a chat_session_id from metadata/headers 
  - (b) reconstructs chat_history from req.messages and 
  - (c) passes that session_id into build_streaming_agent_with_headers. 
- **A2A has the equivalent inputs already**; they're just not being wired through.

**Suggested shape of the fix:**
1. Use `context_id` (not `task_id`) as the `session_id` passed to `build_streaming_agent_with_headers`. That alone restores Aura's existing per-session persistence (orchestration artifacts, scratchpad dirs under {memory_dir}/{run_id}/…).
2. Before spawning the agent, query `task_store` for prior tasks with the same `context_id`, walk their history + completed artifacts (the "response" text artifact in particular), and build a Vec<aura::Message> of alternating user/assistant turns to pass
 as `chat_history`.
3. Decide an ordering/cutoff policy (chronological by status.timestamp, cap on N prior turns) so a long-running context doesn't blow up the prompt.

We could always treat this as a follow-up if you'd rather land the initial impl first — worth at least capturing in the docs or somewhere that this is a known gap, since "streaming: true" on the agent-card implies a richer conversational model than what's currently delivered.

justintime4tea:
**Cancellation: CancelTask only updates state, doesn't stop the agent**

[handler.rs:171](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L171) builds a child of `stream_shutdown_token` and passes it into `agent.stream(...)`, but that token isn't reachable from any cancel path. The `MessageHandler::cancel_task` hook (default no-op on the trait) is not overridden on
`AuraMessageHandler`, so when a client hits `POST /a2a/v1/tasks/{id}:cancel` or sends a `CancelTask` JSON-RPC, the a2a-rs-server REST/JSON-RPC layer:

1. Flips the task's state to `TaskState::Canceled` in the TaskStore,
2. Calls our `cancel_task(...)` — which is a no-op,
3. Broadcasts a `StatusUpdate`.

The spawned `tokio::spawn` keeps running, MCP tool calls in flight aren't aborted, tokens keep being consumed, and when the stream eventually finishes naturally `complete_task` at [(handler.rs:346)](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L3466) overwrites the Canceled state with Completed — so the API lied to the caller.

The OpenAI handler already has the right pattern, _we just need to plug A2A into the same registry._: `RequestCancellation::register(request_id)` registers a `CancellationToken` in a global registry that `mcp_streamable_http` observes to abort tool calls and emit MCP notifications/cancelled (see
[handlers.rs:371](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/handlers.rs#L371) and [crates/aura/src/request_cancellation.rs](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura/src/request_cancellation.rs)).

**Suggested fix:**

1. In `handle_message`, replace the bare `stream_shutdown_token.child_token()` with a registration into the existing registry, keyed off the same `request_id = format!("a2a_{}", task_id)` that's already constructed at [handler.rs:170](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L170):

```rust
let cancellation = aura::RequestCancellation::register(request_id.clone());
let cancel_token = cancellation.token.clone();
// pass cancel_token into agent.stream(...) as today
```

1. This gets MCP notifications/cancelled propagation for free.
2. Override the trait hook on `AuraMessageHandler`:

```rust
async fn cancel_task(&self, task_id: &str) -> HandlerResult<()> {
    let request_id = format!("a2a_{task_id}");
    aura::RequestCancellation::cancel(&request_id, "A2A cancelTask RPC");
    Ok(())
}
```

3. In the spawned task, check `cancel_token.is_cancelled()` at loop exit and skip `complete_task` in that case — the REST/JSON-RPC layer already set Canceled. Otherwise we'll clobber it.
4. Make sure `RequestCancellation::unregister(&request_id)` runs on every exit path (success, error, cancel). The OpenAI handler does this via `RequestResourceGuard` [(handlers.rs:47-73)](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/handlers.rs#L47-L73) — easy to reuse the same RAII pattern here so the registry doesn't
leak entries on long-running servers.

Worth adding an integration test that fires :cancel, polls GetTask, and asserts (a) final state stays Canceled, (b) no further artifacts arrive** after the cancel, and (c) the underlying MCP call observed a cancellation (e.g. via the existing slow_task
tool fixture).

justintime4tea:
**Orchestration runs through A2A but its event stream isn't bridged**

The handler's match arms [(handler.rs:194-322)](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L194-L322) only consume` StreamItem::StreamAssistantItem(Text | ToolCall | Reasoning)` plus Final / FinalMarker. It doesn't subscribe to the request-scoped brokers the OpenAI handler uses (`request_progress_subscribe` /
`tool_event_subscribe` / `tool_usage_subscribe` at [handlers.rs:434-435](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/handlers.rs#L434-L435)), so when a config has `orchestration.enabled = true`:

- Orchestrator executes correctly (it implements StreamingAgent, so .stream(...) works), but
- The 11 `aura.orchestrator.* events`, `worker_phase`, `scratchpad_usage`, `progress`, and `tool_complete` payloads are dropped.

**Net effect for A2A consumers:** a flattened, polling-only view — final assistant text, tool-call announcements (without results), and the terminal state flip. The differentiating orchestration signal isn't visible.

**Question:** is wiring these brokers into A2A Artifacts / metadata planned as a follow-up? If yes, could we capture that in docs/a2a-implementation.md (and maybe tone down the agent card's streaming: true / description) so consumers don't expect orchestrator visibility today?

crates/aura-web-server/src/a2a/handler.rs
justintime4tea: I wrote up an issue over on the a2a-rs github:
- https://github.com/tolgaki/a2a-rs/issues/5

And a PR is now submitted:
- https://github.com/tolgaki/a2a-rs/pull/6

justintime4tea:
**Multi-agent selection & agent-card skill surface**

`resolve_config()` [(handler.rs:44-58)](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L44-L58) only ever resolves to `DEFAULT_AGENT` (when set and matched) or `configs.first()` — there's no per-request selection. 
- This is the OpenAI handler's req.model resolution path that's missing [(handlers.rs:187-208)](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/handlers.rs#L187-L208), and it surfaces in two places that I think we should talk through:

**Selection at request time**

Right now an A2A client invoking an Aura server with N loaded configs can only ever hit one of them. The OpenAI side treats this as a first-class concern (single-config passthrough → explicit model → `DEFAULT_AGENT`); A2A
doesn't have any equivalent dispatch. 

**A few directions worth weighing:**

- message.metadata field (metadata.agent or metadata.model) — cheapest, mirrors the OpenAI model field semantically, non-standard but A2A explicitly permits free-form metadata.
- Skill ID dispatch — advertise each Aura config as a distinct AgentSkill on the card, then route on the skill the client invokes. 
  - This is arguably the most A2A-native pattern — skills exist precisely to disambiguate capabilities on a single agent — and it's a natural fit for our alias model.
- Per-agent agent cards at distinct paths — /.well-known/agent-card.json per agent (URL-scoped or tenant-scoped). 
  - Probably overkill for our deployment shape, but it's what the tenant field on `AgentInterface` [handler.rs:382](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L382), and [handlers.rs:388](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L388) hints at.

**I'd lean toward skill ID dispatch + metadata fallback**, but the right answer depends on *how we expect Aura to be invoked over A2A* (single-purpose deployments per model? a multi-model gateway? a tenant-routed shared instance?).

**The agent card itself** 

Today the card [(handler.rs:352-408)](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L352-L408) takes `configs.first()` regardless of how many are loaded, and advertises exactly one hardcoded skill (id: "chat", text/plain only). For a system that composes MCP tools + RAG + orchestration, that's not providing much of the AURA "sauce" through A2A.

- Should each loaded config become a skill? If we go with skill-ID dispatch above, this falls out for free — and consumers can discover the agent roster from the card.


Smaller asks regardless of which direction we pick:

- The card's description truncates system_prompt at 200 chars [(handler.rs:361-363)](https://github.com/mezmo/aura/blob/63178747f7f5a3aa6361a39a8133a3032d9e9139/crates/aura-web-server/src/a2a/handler.rs#L361-L363) 
  - fine as a placeholder, but the system prompt is internal config, not a user-facing description. 
  - Is this worth an explicit `agent.description` field on the TOML so we're not leaking prompt-engineering content as a public capability description?
- If `DEFAULT_AGENT` is set but doesn't match any loaded config, `resolve_config()` silently falls back to `configs.first()`. The startup validation in `main.rs:176-194` should make that unreachable today, but it's a footgun if someone refactors that validation away.

crates/aura-web-server/src/a2a/handler.rs
justintime4tea: [Here is a PR into dominic/LOG-23822](https://github.com/mezmo/aura/pull/145) that is one way to go about resolving some of the race-condition stuff 
- the JSON-RPC race is a larger refactor to override locally, hopefully upstream accepts the fix presented or offers an alternative fix.

crates/aura-web-server/src/a2a/handler.rs
justintime4tea: We may want to support data parts, it's cheap and high-value - Concatenate data JSON into the prompt (e.g. as a fenced block or Context: <json> prefix) alongside the text. Then, we can update the agent card to add "application/json" to the `inputModes` section. This unlocks ~80% of real-world A2A coordinator/worker patterns without touching Aura's internal pipeline at all.

dominic-mcallister-logdna:
Closing this PR and generating a new version off the official a2a rust implementation.

PR #95, opened 2026-05-08
chore: Rename Aura to AURA

gjanco:
Changing capitalization of AURA.  Also removing
references to open alpha.

Ref: LOG-23836


andrewelizondo (approved):
lgtm 🚢

PR #102, opened 2026-05-11
fix(otel): wrap force_flush in spawn_blocking to avoid deadlock

justintime4tea:
Calling TracerProvider::force_flush() synchronously from inside a tokio::spawn'd task blocks a runtime worker thread, which prevents the BatchSpanProcessor's background export task from being polled. After the first orchestrated request (~55 spans across nested spawns), the exporter stalls permanently and no further traces are exported until the process restarts.

Make flush_tracer async and dispatch force_flush via tokio::task::spawn_blocking, mirroring the pattern shutdown_tracer already uses. The single call site in handlers.rs is already in an async context.

Introduced by the actix-web -> axum migration (cb87687); actix's current_thread runtime masked the deadlock.

### DIFF
--- a/crates/aura-config/src/config_test.rs
+++ b/crates/aura-config/src/config_test.rs
@@ -997,7 +997,7 @@ name = "Test"
 system_prompt = "Test"
 "#;
 
-        let config = load_config_from_str(&config_str);
+        let config = load_config_from_str(config_str);
         assert!(config.is_err(), "no additional fields allowed");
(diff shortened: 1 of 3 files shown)


Shearerbeard (approved):
This looks great - I'm putting my approval on it based on my testings but I'll differ to @teriyakichild 's tests here because he authored the OTEL hierarchy for orchestration

PR #106, opened 2026-05-11
fix(config): validate worker names case-insensitively

justintime4tea:
TOML 0.8 already rejects exact-duplicate [worker.X] headers, but two latent gaps remained that downstream artifact namespacing quietly depends on:

- Case-insensitive collisions: [worker.Alpha] and [worker.alpha] parse as distinct HashMap entries but collide as filenames on macOS APFS and Windows NTFS.
- Empty names: [worker.""] is silently accepted.

Adds validate_worker_names() on OrchestrationConfig in both aura-config and aura crates. Called from Config::validate() so the user-facing load path rejects bad names with a clear error naming the offending workers and the [orchestration.worker.*] section.

Regression test pins toml's exact-duplicate rejection so a future parser bump can't silently regress it.

Ref: LOG-23844

### DIFF
--- a/crates/aura-config/src/config.rs
+++ b/crates/aura-config/src/config.rs
@@ -626,6 +626,10 @@ impl Config {
         // Scratchpad validation
         self.validate_scratchpad()?;
 
+        if let Some(orch) = &self.orchestration {
+            orch.validate_worker_names()?;
+        }
+
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


crates/aura-config/src/config.rs
```diff
@@ -671,6 +675,45 @@ impl Config {
...
+                    "Duplicate worker name in [orchestration.worker.*]: \
+                     '{}' and '{}' collide (names must be unique \
+                     case-insensitively to avoid filesystem collisions in \
+                     artifact persistence)",
+                    existing, name
+                )));
+            }
+        }
+        Ok(())
+    }
```
Shearerbeard, line 727: I think this is a duplicated of below

Shearerbeard (changes requested):

crates/aura-config/src/config.rs
justintime4tea: This is now resolved :pray:

Shearerbeard (approved):

Shearerbeard (approved):

PR #80, opened 2026-04-28
feat: add structured error taxonomy and prometheus metrics (AURA-RM-008 + RM-001)

brandon-shelton-mezmo:
## Summary
- **AURA-RM-008 (Error Taxonomy)**: `ErrorCategory` enum (13 categories), `AuraError` sanitization layer, taxonomy `code` field in API error responses. Client-facing messages are generic — internal details logged server-side only.
- **AURA-RM-001 (Prometheus Metrics)**: `/metrics` endpoint with 6 metric families — request duration histogram, token counters, tool duration histogram, error counters, in-flight gauge, MCP connection gauge. Includes `AURA_METRICS_ENABLED` kill switch and cardinality guards.
- **Grafana dashboard**: Starter dashboard JSON with 12 panels at `examples/grafana/aura-dashboard.json`
- **Development process**: Updated to 8 phases (added Documentation phase before Release)

## Verified
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [x] `cargo test --workspace --lib` — 299 unit tests pass
- [x] 9 integration tests pass (`integration-metrics` feature flag)
- [x] Live tested with Bedrock LLM + Mock MCP server + Prometheus + Grafana
- [x] Token counters match API response usage exactly across 100+ requests
- [x] In-flight gauge showed 30 during concurrent burst, 0 after
- [x] Error counters fire for validation errors (400s) and client disconnects
- [x] MCP tool duration recorded for mock_tool and list_files
- [x] MCP connection gauge shows connected servers
- [x] `/health` and `/metrics` accessible during graceful shutdown
- [x] All 16 acceptance criteria verified
- [x] All quality gates checked
- [x] No secrets in diff
- [x] Backward compatible — existing configs load without changes

## Specs
- `.specify/specs/AURA-RM-008/` — product, architecture, quality specs + review records (4 rounds → clean)
- `.specify/specs/AURA-RM-001/` — product, architecture, quality specs + review records (3 rounds → clean)

Ref: LOG-00000

PR #117, opened 2026-05-12
feat: add deep health and readiness checks (AURA-RM-002)

brandon-shelton-mezmo:
## Summary

- Add `/health/live` (liveness) and `/health/ready` (readiness) endpoints with per-subsystem connectivity verification for LLM providers, MCP servers, and vector stores
- Lightweight probes: HTTP model-listing for LLM, GET for MCP, REST health for Qdrant, AWS credential resolution for Bedrock
- Cached results with configurable TTL (`HEALTH_CHECK_CACHE_TTL_SECS`, default 10s)
- Per-probe timeout (`HEALTH_CHECK_TIMEOUT_SECS`, default 5s)
- Prometheus metrics: `aura_health_ready` gauge, per-subsystem `aura_health_subsystem_status` gauges, `aura_health_check_duration_seconds` histogram
- Sanitized error messages in health responses (no internal IPs/hostnames)
- Shutdown guard exemption for `/health/*` paths
- Existing `/health` endpoint unchanged (backward compatible)

**Note:** Includes AURA-RM-008 (error taxonomy + prometheus metrics) commits as prerequisites. Those are also in PR #80.

## Test plan

- [ ] `cargo fmt --check` passes
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [ ] `cargo test --workspace --lib` — 48 tests pass (includes 17 new health check tests)
- [ ] Verify `/health/live` returns `200 {"status": "alive"}`
- [ ] Verify `/health/ready` returns `200` with per-subsystem JSON when healthy, `503` when unhealthy
- [ ] Verify existing `/health` endpoint unchanged
- [ ] Verify Docker Compose healthcheck still works
- [ ] Jenkins integration tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/aura-web-server/src/health.rs
+++ b/crates/aura-web-server/src/health.rs
@@ -1,0 +1,996 @@
+//! Deep health and readiness checks for Kubernetes probes.
+//!
+//! Provides `/health/live` (liveness) and `/health/ready` (readiness with per-subsystem
+//! connectivity verification). Results are cached with a configurable TTL to avoid
+//! flooding upstream dependencies on every K8s probe.
+
+use actix_web::{HttpResponse, web};
+use aura_config::config::{LlmConfig, McpServerConfig, VectorStoreConfig, VectorStoreType};
+use futures_util::future::join_all;
+use serde::Serialize;
+use std::collections::{BTreeMap, HashSet};
+use std::sync::{Arc, LazyLock};
+use std::time::{Duration, Instant};
+use tokio::sync::RwLock;
+use tracing::{debug, info, warn};
+
+use crate::types::AppState;
+
+/// Shared HTTP client for probe requests. Reuses connection pool across probes.
+static PROBE_CLIENT: LazyLock<reqwest::Client> = LazyLock::new(|| {
+    reqwest::Client::builder()
+        .no_proxy()
+        .redirect(reqwest::redirect::Policy::none())
+        .build()
+        .unwrap_or_default()
+});
+
+// ── Types ────────────────────────────────────────────────────────────────────
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum HealthStatus {
+    Healthy,
+    Unhealthy,
+}
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(tag = "status", rename_all = "snake_case")]
+pub enum SubsystemStatus {
+    Ok,
+    Error { message: String },
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct LlmHealthResult {
+    pub provider: String,
+    pub model: String,
+    pub agent: String,
+    #[serde(flatten)]
+    pub status: SubsystemStatus,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub latency_ms: Option<u64>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct McpHealthResult {
+    pub transport: String,
+    #[serde(flatten)]
+    pub status: SubsystemStatus,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub latency_ms: Option<u64>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct VectorStoreHealthResult {
+    #[serde(rename = "type")]
+    pub store_type: String,
+    #[serde(flatten)]
+    pub status: SubsystemStatus,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub latency_ms: Option<u64>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct HealthChecks {
+    pub llm: Vec<LlmHealthResult>,
+    pub mcp: BTreeMap<String, McpHealthResult>,
+    pub vector_stores: BTreeMap<String, VectorStoreHealthResult>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct HealthCheckResult {
+    pub status: HealthStatus,
+    pub checks: HealthChecks,
+    pub check_duration_ms: u64,
+    pub cached: bool,
+}
+
+// ── Probe Functions ──────────────────────────────────────────────────────────
+
+/// Sanitize a probe error into a safe category string.
+/// Never exposes internal IPs, hostnames, or DNS names.
+fn sanitize_probe_error(err: &reqwest::Error) -> String {
+    if err.is_timeout() {
+        "timeout".to_string()
+    } else if err.is_connect() {
+        "connection_refused".to_string()
+    } else if err.is_redirect() {
+        "redirect_error".to_string()
+    } else {
+        "probe_error".to_string()
+    }
+}
+
+/// Probe a single LLM provider with a lightweight API call.
+async fn probe_llm(config: &LlmConfig, agent_name: &str, timeout: Duration) -> LlmHealthResult {
+    let (provider, model) = config.model_info();
+    let start = Instant::now();
+
+    let status = match tokio::time::timeout(timeout, probe_llm_inner(config)).await {
+        Ok(result) => result,
+        Err(_) => SubsystemStatus::Error {
+            message: "timeout".to_string(),
+        },
+    };
+
+    let latency_ms = Some(start.elapsed().as_millis() as u64);
+
+    debug!(
+        provider = provider,
+        model = model,
+        agent = agent_name,
+        latency_ms = latency_ms,
+        "LLM probe completed"
+    );
+
+    if let SubsystemStatus::Error { ref message } = status {
+        warn!(
+            provider = provider,
+            model = model,
+            agent = agent_name,
+            error = message,
+            "LLM probe failed"
+        );
+    }
+
+    LlmHealthResult {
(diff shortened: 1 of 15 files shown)


crates/aura-web-server/src/streaming/handlers.rs
```diff
@@ -755,24 +766,43 @@ fn handle_tool_result(
     let mut output = Vec::with_capacity(2);
     state.needs_separator = true;
 
+    // Calculate tool duration (always, for metrics even when custom events are off)
+    let duration_ms = state
+        .tool_start_times
+        .remove(&tool_result.id)
+        .map(|start| start.elapsed().as_millis() as u64)
```
kristof-mattei: Unneeded cast to `u64`, we cast to `f64` below.

crates/aura-web-server/src/streaming/handlers.rs
```diff
@@ -755,24 +766,43 @@ fn handle_tool_result(
...
+        .unwrap_or(0);
+
+    let tool_name_for_metrics = state
+        .tool_call_map
+        .get(&tool_result.id)
+        .map(|(name, _)| name.clone())
+        .unwrap_or_else(|| "unknown".to_string());
```
kristof-mattei: We can avoid cloning here to save an allocation:

```
    let tool_name_for_metrics = state
        .tool_call_map
        .get(&tool_result.id)
        .map(|(name, _)| &**name)
        .unwrap_or_else(|| "unknown");
```

crates/aura-web-server/src/metrics.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+/// Track unique tool names globally to enforce cardinality cap.
+static KNOWN_TOOLS: LazyLock<Mutex<HashSet<String>>> = LazyLock::new(|| Mutex::new(HashSet::new()));
+
+const MAX_UNIQUE_TOOL_LABELS: usize = 100;
+const MAX_TOOL_NAME_LEN: usize = 64;
+
+/// Record MCP tool call duration with cardinality guards on the tool label.
+pub fn record_tool_duration(server: &str, tool: &str, status: &str, duration_secs: f64) {
```
kristof-mattei: Why not take in an `std::time::Duration` and convert internally to what you want to display? That way we can leverage the type system to avoid someone misreading the parameter name and send you seconds or milliseconds or nanoseconds.

crates/aura-web-server/src/metrics.rs
```diff
@@ -0,0 +1,208 @@
...
+    let tool_label = if tool.len() > MAX_TOOL_NAME_LEN {
```
kristof-mattei: `.len()` on a string returns the amount of bytes, which is different from the character length (i.e. the width on screen) you're testing here. And even that isn't 100% if we consider graphemes. 

It _probably_ doesn't matter in this code base, where we only have tools with ASCII names, but ideally we would do something like `tool.chars().count()` to check the UTF-8 character count.

But wouldn't it be better to truncate?

crates/aura-web-server/src/metrics.rs
```diff
@@ -0,0 +1,208 @@
...
+        "_other"
+    } else {
+        let mut known = KNOWN_TOOLS.lock().unwrap_or_else(|e| e.into_inner());
```
kristof-mattei: This requires some more thought.

When a lock is poisoned it means that somewhere in the codebase a lock was acquired and panicked before the lock was properly dropped. 

Which means that this moment we cannot make any assumptions of the contents of the lock. 

1. While the `std` library does it's very best to ensure that even in the case of a panic the internal state of the HashMap remains consistent, it's worth thinking about whether we want to rely on that mechanism, and we probably should document our thinking.

2. Equally if there are multiple code-paths accessing `KNOWN_TOOLS`, that do multiple operations while holding a single lock, we don't know whether these changes have been atomically committed. If we do a remove and an add within the same lock, it might be that the remove was called, but the add was not.

While the `HashMap` is still in a valid state from Rust's point of view, it now violates the invariants we have placed on top of it. 


For 1, this is something we don't really control. The only way I can see corruption there is when the thread would forcefully get shot down, which... shouldn't happen.

For 2, If we are able to validate that the usages of `KNOWN_TOOLS`, and the taking of the lock are atomically, or we don't care about the overall state of the HashMap itself, we can do `into_inner()`. However, in the case that it is poisoned (which we know based on the `lock()` result we probably want to clear the the poison with [`clear_poison`](https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.clear_poison).

On the other hand, if we're not able (or don't wish) to continue with a state we can't trust, we can always re-init the `HashMap`, losing our state.

In either case, I'd love to see an `ERROR` or `WARN` log when the lock is poisoned, because when that happens, we probably have an issue somewhere else.

crates/aura-web-server/src/streaming/handlers.rs
```diff
@@ -755,24 +766,43 @@ fn handle_tool_result(
...
-            .tool_call_map
-            .get(&tool_result.id)
-            .map(|(name, _)| name.clone())
-            .unwrap_or_else(|| "unknown".to_string());
-
-        // Aura's ToolResult has result as a plain String
-        // Try to unescape JSON-quoted strings for error detection
-        let result_text = serde_json::from_str::<String>(&tool_result.result)
-            .unwrap_or_else(|_| tool_result.result.clone());
+        let tool_name = tool_name_for_metrics.clone();
```
kristof-mattei: We probably don't need the clone here.

crates/aura-web-server/src/health.rs
```diff
@@ -0,0 +1,996 @@
...
+async fn probe_mcp_server(
+    server_name: &str,
+    config: &McpServerConfig,
+    timeout: Duration,
+) -> (String, McpHealthResult) {
+    let (transport, result) = match config {
+        McpServerConfig::Stdio { .. } => (
+            "stdio",
+            McpHealthResult {
+                transport: "stdio".to_string(),
```
kristof-mattei: `.to_owned()` to get an owned instance of a `&str`.

crates/aura-web-server/src/metrics.rs
```diff
@@ -0,0 +1,208 @@
...
+            "type" => token_type.to_string(),
+            "provider" => provider.to_string(),
+            "agent" => agent.to_string(),
+        )
+        .increment(count);
+    }
+}
+
+/// Track unique tool names globally to enforce cardinality cap.
+static KNOWN_TOOLS: LazyLock<Mutex<HashSet<String>>> = LazyLock::new(|| Mutex::new(HashSet::new()));
```
kristof-mattei: These should be moved to the top.

crates/aura-web-server/src/main.rs
```diff
@@ -89,20 +91,37 @@ struct Args {
...
+    /// this duration to avoid flooding subsystems with probes on every K8s check.
+    /// Recommend setting slightly less than K8s periodSeconds.
+    #[arg(long, env = "HEALTH_CHECK_CACHE_TTL_SECS", default_value = "10")]
+    health_check_cache_ttl_secs: u64,
+
+    /// Health check probe timeout in seconds. Individual subsystem probes that
+    /// exceed this timeout are marked as unhealthy. Must be shorter than K8s
+    /// probe timeoutSeconds (recommend K8s timeout >= probe timeout + 5s).
+    #[arg(long, env = "HEALTH_CHECK_TIMEOUT_SECS", default_value = "5")]
+    health_check_timeout_secs: u64,
```
kristof-mattei: We have an opportunity here to have health_check_timeout as Duration,  then internally we don't have to convert it. We can still name env variable as `..._SECS` and convert it internally. 

Same for the `long` option. 

This way we use the type system to pass on a Duration, and no-one internally needs to worry about whether it's seconds, ...

A `u64` has no provenance of whether it came from seconds or milliseconds or ...

Story about a couple of dollars lost because 2 teams used different units and stored them in a type that didn't carry that information: 

> An investigation attributed the failure to a measurement mismatch between two measurement systems: [SI units](https://en.wikipedia.org/wiki/International_System_of_Units) (metric) by NASA and [US customary](https://en.wikipedia.org/wiki/United_States_customary_units) units by spacecraft builder [Lockheed Martin](https://en.wikipedia.org/wiki/Lockheed_Martin_Space).

https://en.wikipedia.org/wiki/Mars_Climate_Orbiter

kristof-mattei:
General remark: many places where `to_string()` should be replaced with `to_owned()`.

crates/aura-web-server/src/main.rs
```diff
@@ -105,6 +105,16 @@ struct Args {
     health_check_timeout_secs: u64,
```
kristof-mattei:
```
#[arg(
    env = "HEALTH_CHECK_TIMEOUT_SECS",
    default_value = "5",
    long = "health_check-timeout-secs",
    help = "...",
    value_parser = parse_duration_from_secs
)]
pub health_check_timeout: Duration,
...


fn parse_duration_from_secs(value: &str) -> Result<Duration, String> {
    let seconds = value
        .parse()
        .map_err(|error| format!("Could not parse `{}`: {}", value, error))?;

    Ok(Duration::from_secs(seconds))
}
```

Using this you leverage clap to do the conversion for you, and no-where in accessible code do we have access to the broadly typed version.

Issue #120, opened 2026-05-12
bug(orchestration): large tool outputs overflow LLM context when scratchpad is disabled


teriyakichild:
## Summary

In orchestration mode without scratchpad enabled, large MCP tool outputs are passed directly into the LLM conversation, overflowing the per-message size limit. A `deduplicate_logs_relative_time` call returned 11MB of structured content, and OpenAI rejected the next turn with `string_above_max_length`.

## Reproduction

1. Run orchestration **without** scratchpad config
2. Send a query that triggers a broad log dedup — e.g. "what errors are happening in dev?"
3. Worker calls `deduplicate_logs_relative_time` with a broad query:
   ```json
   {"query": "tag:dev (app:jaeger OR _app:jaeger OR \"jaeger-collector\" ...)", "since": "last 15 minutes"}
   ```
4. MCP returns ~302K logs, 11MB structured content
5. Full output goes into conversation as-is
6. Next LLM call fails: `messages[7].content: string too long (12,839,884 > 10,485,760)`

## Workaround

Enable scratchpad — with `[agent.scratchpad]` and `[mcp.servers.*.scratchpad]` configured, the output is intercepted and offloaded to disk as expected.

## Notes

- Only observed in orchestration mode
- The artifact system's `tool_output_artifact_threshold` (500 chars) saves the output to an artifact file but does NOT truncate it from the LLM conversation — the full output still goes into context
- Scratchpad is the only mechanism that actually prevents large outputs from entering the worker's context window

## Environment

- Branch: `feature/orchestration-mode` at `c30b4f5`
- Model: GPT-5.2
- MCP: mezmo_mcp (production)

PR #121, opened 2026-05-13
fix(orchestration): classify size limit errors as context overflow

Shearerbeard:
## Summary

- Widen `categorize_failure_error` to recognize per-message size limit errors (`string_above_max_length`, `"string too long"`) as `ContextOverflow` instead of the catch-all `AgentError`
- Add patterns for Anthropic (`"prompt is too long"`), Bedrock (`"input is too long"`), and fix the `"exceeded"` → `"exceed"` substring gap that missed `exceeds`/`exceed` variants from Anthropic, Ollama, and OpenAI Responses API
- Cap error strings in the continuation prompt and execution summary at `result_summary_length` (default 2000 bytes) to prevent oversized errors from cascading into the coordinator's context window

## Context

When scratchpad is disabled and a worker's MCP tool returns oversized output (e.g. 11MB from `deduplicate_logs_relative_time`), OpenAI rejects the next turn with `string_above_max_length`. This error was misclassified as `AgentError`, causing the coordinator to replan with the same failing task up to `max_planning_cycles` times — each cycle a wasted LLM round-trip reproducing the same failure.

Scratchpad remains the correct prevention mechanism. This fix handles the failure gracefully when it occurs: classify it immediately so the worker breaks without retrying, and bound error strings so they can't poison the coordinator's context.

## Test plan

- [x] 5 new classification unit tests covering OpenAI, Anthropic, Bedrock error formats
- [x] 1 new truncation test verifying long errors are capped with `[truncated]` suffix
- [x] 784/784 unit tests pass (781 existing + 3 net new)
- [x] Commitlint passes
- [ ] Live reproduction via aura-sandbox with no-scratchpad config

Fixes: #120

### DIFF
--- a/crates/aura/src/orchestration/frame_validation_tests.rs
+++ b/crates/aura/src/orchestration/frame_validation_tests.rs
@@ -256,7 +256,7 @@ fn test_continuation_full_scenario() {
     };
 
     let ctx = IterationContext::new(2, plan, Some(fs), failure_history, traces);
-    let prompt = ctx.build_continuation_prompt(3, true);
+    let prompt = ctx.build_continuation_prompt(3, true, 2000);
 
     // Header
     assert!(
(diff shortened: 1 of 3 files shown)


Shearerbeard:
@teriyakichild — this fixes the classification gap from #120 so `string_above_max_length` is caught as `ContextOverflow` (worker breaks immediately, no futile replans). Also widens the classifier for Anthropic/Bedrock/Ollama equivalents and caps error strings entering the coordinator's context at `result_summary_length`.

PR #142, opened 2026-05-14
feat(config): auto-migrate deprecated [llm] to [agent.llm]

Shearerbeard:
## Summary
- Old configs with top-level `[llm]` now auto-migrate to `[agent.llm]` at load time instead of failing
- Deprecation warning logged so users know to update
- Ambiguous configs (both `[llm]` and `[agent.llm]`) and missing `[agent]` still fail with clear errors
- Quickstart example config updated to new shape

### DIFF
--- a/crates/aura-config/src/config_test.rs
+++ b/crates/aura-config/src/config_test.rs
@@ -1412,7 +1412,7 @@ context_window = 200000
     }
 
     #[test]
-    fn test_legacy_top_level_llm_produces_migration_error() {
+    fn test_legacy_top_level_llm_migrated_with_warning() {
         let legacy_config = r#"
 [llm]
 provider = "openai"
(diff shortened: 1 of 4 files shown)


andrewelizondo:
A few small suggestions from a review pass:

1. **Stale doc comment on `migrate_legacy_top_level_llm`**: the doc still frames the function as emitting a migration error ("Detect the legacy top-level `[llm]` table shape and emit a migration error") and the second paragraph explains the silent-ignore rationale that predates this PR. Worth rewriting to reflect that it now auto-migrates and only errors on ambiguous / missing-`[agent]` cases.

2. **Missing test for `[llm.additional_params]` migration**: the previous hard-error message specifically called out `[llm.additional_params]` as a thing users have, and it's a common shape (temperature, max_tokens, etc.). Since the whole `toml::Value` is moved it *should* migrate correctly, but there's no test asserting it does. A regression here would be easy to miss — the renamed `test_legacy_top_level_llm_migrated_with_warning` only checks `api_key` and `model`.

3. **Pin a removal version in the deprecation warning**: the current message says "this fallback will be removed in a future release" without a target. Naming a version (e.g. "removed in v1.22") makes the deprecation actionable for users reading the warning in logs.

PR #145, opened 2026-05-14
fix(a2a): resolve REST streaming subscribe race (proposal)

justintime4tea:
Proposing a local override for the two REST streaming endpoints in a2a-rs-server 1.0.26. Not prescriptive - happy to iterate on shape, naming, or wait on an upstream fix if you'd rather.

The bug
-------
rest_send_streaming_message (rest.rs:397) and
rest_subscribe_to_task (rest.rs:613) snapshot the task before subscribing to the broadcast channel:

    task = task_store.get_flexible(id).await;
    rx   = event_tx.subscribe();

Any event broadcast in that window goes to a channel with zero subscribers - tokio drops it. The event still lands in the task store, but the SSE consumer never sees it.

For synchronous handlers this is mostly invisible. Our AuraMessageHandler tokio::spawns a worker and returns immediately (handler.rs:148), so the worker can emit StatusUpdate(Completed) before the upstream subscribe call. The client then loops on rx.recv() forever waiting for a terminal event that already happened.

The fix
-------
`crates/aura-web-server/src/a2a/overrides.r`s adds two handlers that subscribe first, then snapshot / invoke the handler:

    rx = event_tx.subscribe();
    task = handle_message(...);
    yield initial Task(task);
    loop { rx.recv() ... }

Race window collapses to "may yield one duplicate Task event". A2A clients dedupe by task_id / artifact_id so it's harmless.

Supporting changes
------------------
- `AppState` exposes shared event_tx + task_store OnceLock cells so the override handlers see the same channel as the handler.
- `extract_auth_context()` is pulled out of the A2aServer builder closure into a2a/mod.rs and shared with the overrides.
- **Lagged** is surfaced as a synthetic statusUpdate carrying a lagged_events counter (upstream silently swallows it). If the post-lag store snapshot shows terminal state the loop breaks, otherwise the dropped events could have included the terminal and the loop would hang on the now-empty channel.
- **Message** event matching uses the cached `context_id` from the initial snapshot, not the upstream non-blocking store fetch (which silently drops events under lock contention).

Out of scope
------------
JSON-RPC streaming (message/stream, tasks/resubscribe at server.rs:1052 and :1376) has the same race upstream but isn't overridden here
- Overriding her would mean re-implementing the whole JSON-RPC dispatcher.
- Documented as a known limitation in docs/a2a-implementation.md.
- Tackling this locally would be a much larger refactor. This PR solves the REST streaming issue and adds tests to prevent regressions, but doesn't attempt to fix the JSON-RPC streaming race.
- Upstream fix in flight: [tolgaki/a2a-rs#6](https://github.com/tolgaki/a2a-rs/pull/6) covers all four endpoints (both REST and JSON-RPC) with the same `subscribe`-then-snapshot reorder.
  - When/if this merges and ships in a crates.io release, the local overrides in this directory can be deleted and this caveat removed.

Tests
-----
- 5 deterministic unit tests in `a2a::overrides::tests` cover the fix mechanics (event matching, lagged handling) without any network or timing dependence.
- 2 integration tests under integration-a2a exercise end-to-end SSE flow against a running server. They can't deterministically force the race window (microseconds over the network), but they verify the override path works and would hang loudly on a regression.

Verification
------------
- cargo build --release -p aura-web-server
- cargo clippy -p aura-web-server --tests --features integration-a2a -- -D warnings
- cargo fmt --check
- cargo test --bin aura-web-server  (46/46 pass)

### DIFF
--- a/crates/aura-web-server/src/a2a/handler.rs
+++ b/crates/aura-web-server/src/a2a/handler.rs
@@ -42,16 +42,15 @@ impl AuraMessageHandler {
     }
 
     fn resolve_config(&self) -> Option<aura_config::Config> {
-        if let Some(ref name) = self.app_state.default_agent {
-            if let Some(agent_config) = self
+        if let Some(ref name) = self.app_state.default_agent
+            && let Some(agent_config) = self
                 .app_state
                 .configs
                 .iter()
                 .find(|c| c.agent.alias.as_deref().unwrap_or(&c.agent.name) == name)
                 .cloned()
-            {
-                return Some(agent_config);
-            }
+        {
+            return Some(agent_config);
         }
 
         self.app_state.configs.first().cloned()
--- a/crates/aura-web-server/src/a2a/mod.rs
+++ b/crates/aura-web-server/src/a2a/mod.rs
@@ -1,3 +1,62 @@
 mod handler;
+pub mod overrides;
 
 pub use handler::AuraMessageHandler;
+
+use a2a_rs_server::AuthContext;
+use axum::http::HeaderMap;
+use std::env;
+
+/// Build an `AuthContext` from an HTTP request's headers.
+///
+/// Used by both the `A2aServer` `auth_extractor` closure (for the upstream JSON-RPC and
+/// unmodified REST endpoints) and the local override handlers in [`overrides`] (which serve
+/// `:stream` and `:subscribe` with the race fix applied).
+///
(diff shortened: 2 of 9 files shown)


justintime4tea:
Original branch and PR have been closed - initial implementation has been moved to a different branch/pr using a different underlying A2A crate. Closing this as no longer relevant.

PR #100, opened 2026-05-11
doc: move quickstart to repo root

teriyakichild:
## Summary

- Moves the quickstart stack (AURA + LibreChat + Phoenix) from `examples/quickstart/` to the repo root so `git clone && docker compose up` works without changing directories
- Replaces the stale root `docker-compose.yml` and `.env.example` (which referenced nonexistent files and caused confusion) with the working quickstart versions
- Restructures the README to lead with the 3-step quickstart inline, followed by customization guidance and a "What's Next" section linking to advanced quickstarts
- Adds `quickstart.toml` at root (renamed from `config.toml` to avoid the gitignore and not collide with the dev workflow)
- Updates `examples/README.md` to redirect newcomers to the root quickstart and focus on reference configs and advanced quickstarts

## Motivation

People were cloning the repo, reading the quickstart README, then running `docker compose up` from the repo root instead of `cd examples/quickstart`. The root-level compose file was a stale artifact from early development that referenced nonexistent config files, so it would fail silently or confusingly. Moving the quickstart to root eliminates this entire class of error.

## Test plan

- [ ] `cp .env.example .env`, set API key, `docker compose up -d` from repo root
- [ ] Verify LibreChat loads at :3080, Phoenix at :6006, AURA health at :3000/health
- [ ] Verify `quickstart.toml` changes take effect after `docker compose restart aura`
- [ ] Verify `.env` changes take effect after `docker compose up -d`
- [ ] Check all README links resolve correctly
- [ ] Verify `examples/quickstart-orchestration-math/` and `examples/quickstart-k8s-sre/` still work independently

PR #155, opened 2026-05-18
Add legacy SSE MCP transport support

### LINKED ISSUE #153: Add legacy SSE MCP transport support
## Summary

Add support for MCP servers using the legacy (pre-2025-11-05) SSE
protocol alongside the existing HTTP streamable and STDIO transports.

## Motivation

Some MCP servers still expose only the original SSE-based protocol
where the client GETs an SSE endpoint, receives a message URL via an
`endpoint` event, then POSTs JSON-RPC requests to that URL while
receiving responses on the SSE stream.

This is also a prerequisite for TerminalBench integration — the
TerminalBench harness exposes tools via an SSE MCP server that workers
need to connect to for keystroke injection and screen capture.

## Configuration

```toml
[mcp.servers.my_sse_server]
transport = "sse"
url = "http://localhost:8082/sse"
description = "Legacy SSE MCP server"

[mcp.servers.my_sse_server.headers]
Authorization = "Bearer my-token"
```

SSE servers support the same `headers`, `headers_from_request`, and
`scratchpad` options as `http_streamable`.

Shearerbeard:
Adds SSE as a transport option for MCP servers alongside http_streamable and stdio. Some MCP servers -- including TerminalBench's tool harness -- only expose the legacy SSE protocol, so we need this to connect.

Everything http_streamable supports (headers, headers_from_request, scratchpad, cancellation) works with SSE too. Refactored the old StreamableHttpMcpClient into a transport-agnostic McpClient so both transports share the same plumbing.

Tested against @modelcontextprotocol/server-everything in SSE mode -- ran scripted multi-turn conversations with back-to-back tool calls, rapid-fire requests, and streaming, all held up fine.

Should review after Tony's #100 docs changes land since there's some overlap.

Fixes: #153

### DIFF
--- a/crates/aura-config/src/builder.rs
+++ b/crates/aura-config/src/builder.rs
@@ -125,6 +125,19 @@ impl RigBuilder {
                             headers_from_request: headers_from_request.clone(),
                             scratchpad: scratchpad.clone(),
                         },
+                        crate::config::McpServerConfig::Sse {
+                            url,
+                            headers,
+                            description,
+                            headers_from_request,
+                            scratchpad,
+                        } => McpServerConfig::Sse {
+                            url: url.clone(),
+                            headers: headers.clone(),
+                            description: description.clone(),
+                            headers_from_request: headers_from_request.clone(),
+                            scratchpad: scratchpad.clone(),
+                        },
                     };
                     (name.clone(), converted_server)
                 })
@@ -282,17 +295,24 @@ fn resolve_mcp_headers(
     };
 
     for (server_name, server_config) in mcp_config.servers.iter_mut() {
-        let McpServerConfig::HttpStreamable {
-            headers: server_headers,
-            headers_from_request,
-            ..
-        } = server_config
-        else {
-            tracing::debug!(
-                "Server '{}': STDIO transport, skipping header injection",
-                server_name
-            );
-            continue;
+        let (server_headers, headers_from_request) = match server_config {
+            McpServerConfig::HttpStreamable {
+                headers,
+                headers_from_request,
+                ..
+            } => (headers, headers_from_request),
+            McpServerConfig::Sse {
+                headers,
+                headers_from_request,
+                ..
+            } => (headers, headers_from_request),
+            McpServerConfig::Stdio { .. } => {
+                tracing::debug!(
+                    "Server '{}': STDIO transport, skipping header injection",
+                    server_name
+                );
+                continue;
+            }
         };
 
         // Resolve headers_from_request mappings using the incoming request headers.
--- a/crates/aura-web-server/tests/mcp_progress_test.rs
+++ b/crates/aura-web-server/tests/mcp_progress_test.rs
@@ -3,7 +3,7 @@
 //! Integration tests for MCP progress notification forwarding.
 //!
 
-use aura::mcp_streamable_http::StreamableHttpMcpClient;
+use aura::mcp_streamable_http::McpClient;
 use aura::{request_progress_global, request_progress_subscribe, request_progress_unsubscribe};
 use std::collections::HashMap;
 use std::time::Duration;
@@ -12,11 +12,7 @@ use tokio::time::timeout;
 #[tokio::test]
 async fn test_mcp_progress_notifications_received() {
     // Skip if mock server not running
-    let client = match StreamableHttpMcpClient::new(
-        "http://127.0.0.1:9999/mcp".to_string(),
-        &HashMap::new(),
-    )
-    .await
+    let client = match McpClient::new("http://127.0.0.1:9999/mcp".to_owned(), &HashMap::new()).await
     {
         Ok(c) => c,
         Err(_) => {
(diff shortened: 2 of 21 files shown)


crates/aura-config/src/builder.rs
```diff
@@ -282,17 +295,24 @@ fn resolve_mcp_headers(
...
+                headers,
+                headers_from_request,
+                ..
+            } => (headers, headers_from_request),
+            McpServerConfig::Sse {
+                headers,
+                headers_from_request,
+                ..
+            } => (headers, headers_from_request),
+            _ => {
```
kristof-mattei: Nit: `_` here makes it so that future additions to `McpServerConfig` will be added to this branch, yet the comment implies it's for `STDIO`. 

Suggestion: switch `_` for `McpServerConfig::Stdio`.

crates/aura-web-server/tests/mcp_progress_test.rs
```diff
@@ -12,18 +12,14 @@ use tokio::time::timeout;
...
-    .await
-    {
-        Ok(c) => c,
-        Err(_) => {
-            eprintln!("Skipping test: mock MCP server not running on port 9999");
-            return;
-        }
-    };
+    let client =
+        match McpClient::new("http://127.0.0.1:9999/mcp".to_string(), &HashMap::new()).await {
```
kristof-mattei: Use `.to_owned()` instead of `.to_string()` to go from `&str` to `String`.

crates/aura/src/mcp_sse.rs
```diff
@@ -0,0 +1,235 @@
...
+        let response = response
+            .error_for_status()
+            .map_err(SseTransportError::Http)?;
+
+        match response
+            .headers()
+            .get(CONTENT_TYPE)
+            .and_then(|v| v.to_str().ok())
+        {
+            Some(ct) if ct.starts_with("text/event-stream") => {}
```
kristof-mattei: This will break when there are multiple content-types. 

[`get`](https://docs.rs/reqwest/latest/reqwest/header/struct.HeaderMap.html#method.get) only returns the first item:

> Returns a reference to the value associated with the key.
> 
> If there are multiple values associated with the key, then the first one is returned. **Use [`get_all`](https://docs.rs/reqwest/latest/reqwest/header/struct.HeaderMap.html#method.get_all)  to get all values associated with a given key**. Returns None if there are no values associated with the key.

Emphasis mine.

 
```rust
use http::header::{CONTENT_TYPE, HeaderMap, HeaderValue};

fn main() {
    let mut headers = HeaderMap::new();

    headers.insert(CONTENT_TYPE, HeaderValue::from_static("application/json"));

    headers.append(CONTENT_TYPE, HeaderValue::from_static("text/event-stream"));

    println!("{:?}", headers.get(CONTENT_TYPE));

    let content_types = headers.get_all(CONTENT_TYPE);

    println!("{:?}", content_types);

    for val in content_types {
        println!("Header: {:?}", val);
    }
}
```

crates/aura/src/mcp_streamable_http.rs
```diff
@@ -96,7 +96,31 @@ impl Clone for StreamableHttpMcpClient {
...
+    ///
+    /// This is the transport-agnostic constructor used by both HTTP streamable
+    /// and legacy SSE transports.
+    pub(crate) async fn from_transport<T>(transport: T, server_url: String) -> Result<Self>
+    where
+        T: rmcp::transport::Transport<RoleClient> + Send + 'static,
+        T::Error: std::error::Error + Send + Sync + 'static,
+    {
+        let current_http_request_id = Arc::new(RwLock::new(None));
+        let handler = ProgressEnabledHandler::new(current_http_request_id.clone());
```
kristof-mattei: For clarity, it's better to clone an `Arc` with `Arc::clone(&current_http_request_id)`

crates/aura/src/mcp.rs
```diff
@@ -466,6 +472,71 @@ impl McpManager {
...
+    ) -> Result<usize, BuilderError> {
+        info!("  Connecting to SSE server at: {}", url);
+
+        match self.try_connect_sse(server_name, url, headers).await {
+            Ok(tools_count) => {
+                info!("  SSE connection successful");
+                Ok(tools_count)
+            }
+            Err(e) => {
+                warn!("  SSE MCP connection failed: {}", e);
```
kristof-mattei: `connect_http_streamable` has separate error for 401 errors. This one doesn't.

kristof-mattei (changes requested):

crates/aura/src/mcp_sse.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// 1. GET the SSE endpoint with `Accept: text/event-stream`
+    /// 2. Verify the response Content-Type
+    /// 3. Read SSE events until the `endpoint` event is received
+    /// 4. Resolve the message endpoint URL
+    /// 5. Return a connected `SseTransport`
+    pub async fn connect(
+        url: &str,
+        headers: &HashMap<String, String>,
+    ) -> Result<Self, SseTransportError> {
+        let sse_endpoint = url.parse::<http::Uri>()?;
```
kristof-mattei: Why parse this to `http::Uri`? 

The only place it's used is in 

```
                break resolve_message_endpoint(sse_endpoint, ep)?;
```

Which converts it to `url::Url`. Can you immediately go to `url::Url`?

kristof-mattei (approved):

PR #160, opened 2026-05-18
feat(a2a): initial implementation

dominic-mcallister-logdna:
summary: codes A2A processing against the official a2a project rust implementation. An official crate has yet to be released, so directly imports the Github projects pinned to the latest version as of 13 May 2026. Currently handles task generation
calling internal aura SSE processing. Uses an in-memory store to keep dependencies down.

ref: LOG-23822

### DIFF
--- a/crates/aura-web-server/Cargo.toml
+++ b/crates/aura-web-server/Cargo.toml
@@ -28,8 +28,8 @@ uuid = { version = "1.0", features = ["v4"] }
 chrono = "0.4"
 async-trait = "0.1.89"
 
-a2a = { package = "a2a-lf", git = "https://github.com/a2aproject/a2a-rs.git", tag = "a2a-server-lf-v0.3.0" }
-a2a-server = { package = "a2a-server-lf", git = "https://github.com/a2aproject/a2a-rs.git", tag = "a2a-server-lf-v0.3.0" }
+a2a = { package = "a2a-lf", git = "https://github.com/a2aproject/a2a-rs.git", tag = "a2a-server-lf-v0.3.1" }
+a2a-server = { package = "a2a-server-lf", git = "https://github.com/a2aproject/a2a-rs.git", tag = "a2a-server-lf-v0.3.1" }
 
 [features]
 default = ["otel"]
--- a/crates/aura-web-server/Cargo.toml
+++ b/crates/aura-web-server/Cargo.toml
@@ -9,8 +9,8 @@ repository.workspace = true
 [dependencies]
 aura = { path = "../aura" }
 aura-config = { path = "../aura-config" }
-axum = { version = "0.7", features = ["json"] }
-tower-http = { version = "0.6", features = ["trace"] }
+axum = { version = "0.8", features = ["json"] }
+tower-http = { version = "0.6", features = ["trace", "timeout"] }
 bytes = "1"
 async-stream = "0.3"
 clap = { version = "4.5", features = ["derive", "env"] }
@@ -26,6 +26,10 @@ opentelemetry = { workspace = true, optional = true }
 env_logger = "0.11"
 uuid = { version = "1.0", features = ["v4"] }
 chrono = "0.4"
+async-trait = "0.1.89"
+
+a2a = { package = "a2a-lf", git = "https://github.com/a2aproject/a2a-rs.git", tag = "a2a-server-lf-v0.3.0" }
+a2a-server = { package = "a2a-server-lf", git = "https://github.com/a2aproject/a2a-rs.git", tag = "a2a-server-lf-v0.3.0" }
 
 [features]
 default = ["otel"]
@@ -34,6 +38,7 @@ otel = ["aura/otel", "dep:tracing-opentelemetry", "dep:opentelemetry"]
 # Integration test feature flags
 # Parent flag - enables all integration tests
 integration = [
+    "integration-a2a",
     "integration-streaming",
     "integration-header-forwarding",
     "integration-mcp",
@@ -43,6 +48,7 @@ integration = [
 ]
 
 # Per-suite flags (can run individually)
+integration-a2a = []                      # a2a tests
 integration-streaming = []                # streaming_tests, concurrent_streaming_isolation_test
 integration-header-forwarding = []        # header_forwarding_tests
 integration-mcp = []                      # fastmcp_structured_content_test
--- a/crates/aura-web-server/src/a2a/agent_executor.rs
+++ b/crates/aura-web-server/src/a2a/agent_executor.rs
@@ -9,7 +9,7 @@ use a2a::{
     TaskStatusUpdateEvent, VERSION,
 };
 use a2a_server::{AgentExecutor, ExecutorContext, TaskStore};
-use aura::{StreamItem, StreamedAssistantContent};
+use aura::{RequestCancellation, StreamItem, StreamedAssistantContent, StreamingAgent};
 use aura_config::RigBuilder;
 use futures_util::StreamExt;
 use futures_util::stream::BoxStream;
@@ -18,14 +18,23 @@ use tokio::sync::Mutex;
 use tokio_util::sync::CancellationToken;
 use tracing::{Level, event};
 
-use crate::{a2a::SharedTaskStore, types::AppState};
+use crate::{
+    a2a::SharedTaskStore,
+    types::{ActiveRequestGuard, AppState},
+};
 
 const PLAIN_TEXT: &str = "text/plain";
 
 pub struct AuraAgentExecutor {
     app_state: Arc<AppState>,
     task_store: SharedTaskStore,
-    task_cancel_tokens: Arc<Mutex<HashMap<String, CancellationToken>>>,
+    task_cancel_state: Arc<Mutex<HashMap<String, TaskCancelEntry>>>,
+}
+
+struct TaskCancelEntry {
+    token: CancellationToken,
+    agent: Arc<dyn StreamingAgent>,
+    request_id: String,
 }
 
 impl AuraAgentExecutor {
@@ -33,7 +42,7 @@ impl AuraAgentExecutor {
         Self {
             app_state,
             task_store,
-            task_cancel_tokens: Arc::new(Mutex::new(HashMap::new())),
+            task_cancel_state: Arc::new(Mutex::new(HashMap::new())),
         }
     }
 
@@ -114,7 +123,7 @@ impl AgentExecutor for AuraAgentExecutor {
     ) -> BoxStream<'static, Result<StreamResponse, A2AError>> {
         let config = self.resolve_config();
         let stream_shutdown_token = self.app_state.stream_shutdown_token.clone();
-        let task_cancel_tokens = self.task_cancel_tokens.clone();
+        let task_cancel_state = self.task_cancel_state.clone();
         let active_request_tracker = self.app_state.active_requests.clone();
         let task_store = self.task_store.clone();
         let mut append_tracker: HashMap<(String, String, String), bool> = HashMap::new();
@@ -166,32 +175,49 @@ impl AgentExecutor for AuraAgentExecutor {
             let history = get_history_for_context(task_store.clone(), &request_id, &context_id, &task_id).await?;
 
             let cancel_token = stream_shutdown_token.child_token();
+            // Register with the global cancellation registry for parity with the OpenAI handler
+            // and to let any future code address this request by id.
+            RequestCancellation::register(request_id.clone());
             {
-                // keep track of this cancel token in case the a2a framework
-                // receives a cancel call and we need to propagate
-                let mut cancel_map = task_cancel_tokens.lock().await;
-                cancel_map.insert(task_id.clone(), cancel_token.clone());
+                let mut cancel_map = task_cancel_state.lock().await;
+                cancel_map.insert(task_id.clone(), TaskCancelEntry {
+                    token: cancel_token.clone(),
+                    agent: agent.clone(),
+                    request_id: request_id.clone(),
+                });
             }
 
-            let mut stream = match agent.stream(&text, history, cancel_token, &request_id).await {
+            let mut stream = match agent.stream(&text, history, cancel_token.clone(), &request_id).await {
                 Ok(s) => s,
                 Err(e) => {
-                    // processing never started, so we can remove the cancel token from tracking
-                    let mut cancel_map = task_cancel_tokens.lock().await;
-                    cancel_map.remove(&task_id);
+                    {
+                        let mut cancel_map = task_cancel_state.lock().await;
+                        cancel_map.remove(&task_id);
+                    }
+                    RequestCancellation::unregister(&request_id);
 
                     yield Ok(fail_status(&task_id, &context_id, &e.to_string()));
                     return;
                 }
             };
 
-            // let the main state machinery know we've got a job in progress
-            active_request_tracker.increment();
+            // RAII guard: drop on any generator exit (loop break, early return, panic,
+            // consumer drop) produces exactly one decrement. Replaces the manual
+            // increment/decrement pair that previously raced with cancel() — a fast
+            // cancel before this line, or a cancel mid-loop followed by natural
+            // loop-exit cleanup, could double-decrement and wrap the counter.
+            let _request_guard = ActiveRequestGuard::new(active_request_tracker);
 
             let mut success = true; // assume everything is successful
 
             let mut reasoning_num = 0;
-            while let Some(item) = stream.next().await {
+            loop {
+                let next = tokio::select! {
+                    biased;
+                    _ = cancel_token.cancelled() => break,
+                    next = stream.next() => next,
+                };
+                let Some(item) = next else { break };
                 match item {
                     Ok(StreamItem::StreamAssistantItem(StreamedAssistantContent::Text(t))) => {
                         event!(Level::DEBUG, request_id, t, "stream content received");
@@ -360,17 +386,41 @@ impl AgentExecutor for AuraAgentExecutor {
                 }
             }
 
-            // processing done, remove the token from the map, no longer need to track
-            {
-                let mut cancel_map = task_cancel_tokens.lock().await;
-                cancel_map.remove(&task_id);
+            // If cancel_token fired but our entry is still in the map, the cancel came
+            // from the parent stream_shutdown_token (server shutdown), not from our
+            // cancel() hook — cancel() removes its entry before firing the token.
+            // In that case the executor has to drive MCP cleanup itself and emit a
+            // terminal Canceled status (the OpenAI handler does the equivalent in its
+            // Shutdown post-loop arm).
+            let entry_still_present = {
+                let mut cancel_map = task_cancel_state.lock().await;
+                cancel_map.remove(&task_id).is_some()
+            };
+            let shutdown_initiated_cancel = cancel_token.is_cancelled() && entry_still_present;
+            RequestCancellation::unregister(&request_id);
+
+            if shutdown_initiated_cancel {
+                agent
+                    .cancel_and_close_mcp(&request_id, "server shutdown")
+                    .await;
+
+                yield Ok(StreamResponse::StatusUpdate(TaskStatusUpdateEvent {
+                    task_id: task_id.clone(),
+                    context_id: context_id.clone(),
+                    status: TaskStatus {
+                        state: TaskState::Canceled,
+                        message: None,
+                        timestamp: Some(chrono::Utc::now()),
+                    },
+                    metadata: None,
+                }));
             }
 
-            // let the main state machinery know we've completed this particular job
-            active_request_tracker.decrement();
+            // _request_guard drops at end of generator scope → exactly one decrement.
 
-            // if successful (hit either finalizer), mark the task as such.
-            if success {
+            // Skip Completed if cancel() or the shutdown path already emitted Canceled —
+            // yielding here would clobber it.
+            if success && !cancel_token.is_cancelled() {
                 yield Ok(StreamResponse::StatusUpdate(TaskStatusUpdateEvent {
                     task_id,
                     context_id,
@@ -388,19 +438,28 @@ impl AgentExecutor for AuraAgentExecutor {
     fn cancel(&self, ctx: ExecutorContext) -> BoxStream<'static, Result<StreamResponse, A2AError>> {
         let task_id = ctx.task_id.clone();
         let context_id = ctx.context_id.clone();
-        let task_cancel_tokens = self.task_cancel_tokens.clone();
-        let active_request_tracker = self.app_state.active_requests.clone();
+        let task_cancel_state = self.task_cancel_state.clone();
 
         Box::pin(futures_util::stream::once(async move {
-            // a2a got a cancel call, so make sure rig cancels as well
-            // and remove the token from tracking
-            let mut cancel_map = task_cancel_tokens.lock().await;
-            if let Some(token) = cancel_map.remove(&task_id) {
-                token.cancel();
-            }
+            let entry = {
+                let mut cancel_map = task_cancel_state.lock().await;
+                cancel_map.remove(&task_id)
+            };
 
-            // let the main state machinery know this task is effectively complete
-            active_request_tracker.decrement();
+            // Token-cancel wakes execute()'s select! → loop breaks → generator drops
+            // → ActiveRequestGuard drops → exactly one decrement. cancel() never
+            // touches the tracker directly, which closes the prior double-decrement
+            // / underflow race.
+            if let Some(entry) = entry {
+                // Send notifications/cancelled to in-flight MCP tool calls. No-op in
+                // orchestration mode (workers manage their own MCP cancellation).
+                entry
+                    .agent
+                    .cancel_and_close_mcp(&entry.request_id, "A2A cancelTask")
+                    .await;
+                entry.token.cancel();
+                RequestCancellation::unregister(&entry.request_id);
+            }
 
             Ok(StreamResponse::StatusUpdate(TaskStatusUpdateEvent {
                 task_id,
--- a/crates/aura-web-server/src/a2a/agent_executor.rs
+++ b/crates/aura-web-server/src/a2a/agent_executor.rs
@@ -60,7 +60,14 @@ impl AuraAgentExecutor {
         self.app_state.configs.first().cloned()
     }
 
-    pub fn build_agent_card(&self) -> AgentCard {
+    /// Build the A2A agent card.
+    ///
+    /// `base_url` is the externally-reachable origin (e.g. `https://aura.example.com`)
+    /// used to make the interface endpoints absolute. The A2A spec requires absolute
+    /// interface URLs — clients pass them straight to their HTTP layer, which rejects
+    /// relative paths.
+    pub fn build_agent_card(&self, base_url: &str) -> AgentCard {
+        let base = base_url.trim_end_matches('/');
         let config = self.resolve_config();
         let name = config
             .as_ref()
@@ -94,8 +101,8 @@ impl AuraAgentExecutor {
                 extended_agent_card: None,
             },
             supported_interfaces: vec![
-                AgentInterface::new("/a2a/v1", TRANSPORT_PROTOCOL_HTTP_JSON),
-                AgentInterface::new("/a2a/v1/rpc", TRANSPORT_PROTOCOL_JSONRPC),
+                AgentInterface::new(format!("{base}/a2a/v1"), TRANSPORT_PROTOCOL_HTTP_JSON),
+                AgentInterface::new(format!("{base}/a2a/v1/rpc"), TRANSPORT_PROTOCOL_JSONRPC),
             ],
             skills: vec![AgentSkill {
                 id: "chat".to_owned(),
--- a/crates/aura-web-server/src/a2a/agent_executor.rs
+++ b/crates/aura-web-server/src/a2a/agent_executor.rs
@@ -1,0 +1,542 @@
+use std::cmp::Reverse;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use a2a::{
+    A2AError, AgentCapabilities, AgentCard, AgentInterface, AgentSkill, Artifact, ListTasksRequest,
+    Message, Part, PartContent, Role, StreamResponse, TRANSPORT_PROTOCOL_HTTP_JSON,
+    TRANSPORT_PROTOCOL_JSONRPC, TaskArtifactUpdateEvent, TaskState, TaskStatus,
+    TaskStatusUpdateEvent, VERSION,
+};
+use a2a_server::{AgentExecutor, ExecutorContext, TaskStore};
+use aura::{StreamItem, StreamedAssistantContent};
+use aura_config::RigBuilder;
+use futures_util::StreamExt;
+use futures_util::stream::BoxStream;
+use serde_json::Value;
+use tokio::sync::Mutex;
+use tokio_util::sync::CancellationToken;
+use tracing::{Level, event};
+
+use crate::{a2a::SharedTaskStore, types::AppState};
+
+const PLAIN_TEXT: &str = "text/plain";
+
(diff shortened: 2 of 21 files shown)


crates/aura-web-server/Cargo.toml
```diff
@@ -26,6 +26,11 @@ opentelemetry = { workspace = true, optional = true }
 env_logger = "0.11"
 uuid = { version = "1.0", features = ["v4"] }
 chrono = "0.4"
+async-trait = "0.1.89"
+
+a2a = { package = "a2a-lf", git = "https://github.com/a2aproject/a2a-rs.git", rev = "73c72eeed997fddf5a00be44068575437e7f3f82" }
```
justintime4tea: Instead of `rev = 73c72ee...` can we use the tags?

**a2a**
`tag = "a2a-lf-v0.3.0"`

**a2a-server**
`tag = "a2a-server-lf-v0.3.0"`

And are we using a2a-grpc at all? If not, let's remove it? If so, could we also use tags?
**a2a-grpc**
`tag = "a2a-grpc-v0.2.2"`

crates/aura-web-server/src/a2a/agent_executor.rs
```diff
@@ -0,0 +1,406 @@
...
+            let context_id = ctx.context_id.clone();
+
+            let text = ctx.message
+                .ok_or_else(|| A2AError::invalid_params("Message has no parts to use as a command."))
+                .and_then(|msg| extract_text(msg.parts))?;
+
+            let req_headers: HashMap<String, String> = ctx
+                .service_params
+                .iter()
+                .filter_map(|(k, v)| v.first().map(|val| (k.clone(), val.clone())))
```
justintime4tea: We're only picking `.first()` from each `Vec<String>`. Once `ServiceParams` is populated for real (see previous comment, upstream is sending empty ServiceParams), multi-valued headers (e.g. Cookie, Accept) would silently drop subsequent values. 

We may need to revisit our request_headers impl/design and either support multiple header values or add a comment noting that only the first value is forwarded?

crates/aura-web-server/src/a2a/agent_executor.rs
```diff
@@ -0,0 +1,406 @@
...
+
+        Box::pin(async_stream::stream! {
+            let task_id = ctx.task_id.clone();
+            let context_id = ctx.context_id.clone();
+
+            let text = ctx.message
+                .ok_or_else(|| A2AError::invalid_params("Message has no parts to use as a command."))
+                .and_then(|msg| extract_text(msg.parts))?;
+
+            let req_headers: HashMap<String, String> = ctx
```
justintime4tea, line 146: ### TLDR;

`headers_from_request` MCP forwarding silently no-ops when using A2A - root cause is upstream.

I've submitted [an issue](https://github.com/a2aproject/a2a-rs/issues/69) and [accompanying PR](https://github.com/a2aproject/a2a-rs/pull/70) upstream.

### Breakdown / What's happening

The code here reads incoming request headers from `ctx.service_params` and forwards them to `build_streaming_agent_with_headers(...)` but `ctx.service_params` is always empty when arriving from upstream (*a2aproject/a2a-rs*). 
- Every REST handler in `a2a-server/src/rest.rs` and the JSON-RPC handler in `jsonrpc.rs:43` construct `ServiceParams::new()` and pass it straight through (see linked issue above for details).

**Net effect:**
headers_from_request MCP injection silently no-ops over `/a2a/v1/*` — no error, no log line. 
- Anything that depends on per-request header forwarding to MCP-fronted APIs will quietly stop working when invoked over A2A.

### a2a-rs crate is WIP, it's more a gap than bug 

The sibling transport servers in the same workspace populate `ServiceParams` properly:

- `a2a-grpc/src/server.rs` defines `extract_service_params(metadata: &tonic::metadata::MetadataMap) -> ServiceParams` and calls it at every method (send_message, get_task, cancel_task, …).
- `a2a-slimrpc/src/server.rs` does the same via `context_to_service_params(&context)`.

The axum HTTP transport in a2a-server **just doesn't have the equivalent** — the `HeaderMap → ServiceParams` extractor was never written. There was no open issue or PR tracking this so I went ahead and created both (linked above).

### What to do downstream?

We could try to implement a workaround locally but one path that is a non-obvious dead-end: the upstream `InterceptedHandler / CallInterceptor` machinery in `middleware.rs` *looks like a natural seam*, but **InterceptedHandler has no RequestHandler impl anywhere in the crate** — the before/after hooks never run on the `rest_router` / `jsonrpc_router` paths. It's **scaffolding**, not a **hook**.

**A couple options that are worth noodling:**

1. Land this PR as-is and wait for upstream.
   - Cleanest, but ships with the silent no-op until upstream merges and we bump the rev. Worth pairing with a doc note in docs/a2a-implementation.md flagging that headers_from_request doesn't currently flow through the A2A path.

2. Add a local wrapper as a stopgap.
   - Replace the calls to `a2a_server::rest::rest_router(...)` and `jsonrpc::jsonrpc_router(...)` in `main.rs:240-243` with AURA-owned routers that mirror the upstream endpoints but take a `HeaderMap` extractor and build `ServiceParams` from it before calling `state.handler.send_message(&params, req)`. 
     - Some duplication, but it's the same ~30 lines we want in the upstream PR. 
     - Delete the wrapper once upstream merges.

I don't have a strong opinion — depends on how soon any caller actually needs A2A + MCP headers forwarded. If no one needs it immediately, **(1)** is fine; if anyone's going to try wiring up MCP with header forwarding **(2)** avoids the silent failure.

crates/aura-web-server/src/a2a/agent_executor.rs
```diff
@@ -0,0 +1,406 @@
...
+impl AgentExecutor for AuraAgentExecutor {
+    fn execute(
+        &self,
+        ctx: ExecutorContext,
+    ) -> BoxStream<'static, Result<StreamResponse, A2AError>> {
+        let config = self.resolve_config();
+        let stream_shutdown_token = self.app_state.stream_shutdown_token.clone();
+        let task_cancel_tokens = self.task_cancel_tokens.clone();
+
+        Box::pin(async_stream::stream! {
```
justintime4tea, line 138: A2A tasks bypass the active-request tracker - `chat_completions` increments via `ActiveRequestGuard` (`crates/aura-web-server/src/handlers.rs:422`), so the graceful-shutdown drain at `main.rs:295` waits for it. This stream never touches `app_state.active_requests`, so on `SIGTERM` phase 1 (`shutdown_token.cancel`) the drain considers the server idle and trips phase 2 (`stream_shutdown_token.cancel` at `main.rs:301`) immediately — terminating in-flight A2A tasks without honoring `--shutdown-timeout-secs`.

**Fix:** 
Hold an ActiveRequestGuard for the lifetime of this stream or document the asymmetry in `docs/a2a-implementation.md`.

crates/aura-web-server/src/a2a/agent_executor.rs
```diff
@@ -0,0 +1,406 @@
...
+}
+
+fn extract_text(parts: Vec<Part>) -> Result<String, A2AError> {
+    let mut strings: Vec<&str> = Vec::new();
+
+    for part_content in parts.iter() {
+        if let PartContent::Text(t) = &part_content.content {
+            strings.push(t)
+        } else {
+            return Err(A2AError::invalid_params(
```
justintime4tea, line 492: Right now this only accepts `Part::Text`. This limits Aura to string-shaped prompts and we lose out on some potential gains even with just minimally supporting `application/json`. Even if we only care about one or two of the bullets below, I feel like the effort is worth it for the consistency and improved ergonomics of accepting structured input.

I suggest just extend extract_text to handle PartContent::Data, serialize JSON parts deterministically into the prompt, and add application/json to default_input_modes + the skill's input_modes on the agent card.

**Accepting application/json (via the A2A spec's data parts) would be a boon because:**

1. Structured input from upstream - When another A2A agent calls Aura as a worker, it may have typed payloads — forcing it to stringify and hope the LLM re-parses correctly is lossy and brittle.

2. Tool-call-shaped inputs. MCP tools have JSON schemas; if a caller already knows the exact arguments, structured input lets Aura bypass LLM-mediated arg construction and route directly.

3. Schema-discoverable skills. Once skills advertise application/json with an input schema in the agent card, A2A clients can validate before they call — closer to RPC ergonomics than chat ergonomics.

      Don't necessarily need to do this now, or ever, but it's more of a "nice to have" that we could do and becomes possible once we support structured input.

4. Round-tripping. Aura already emits structured Artifact metadata (tool_call args, scratchpad usage, usage tokens). Symmetric input keeps the contract consistent.

  The output side of the A2A contract already speaks structured JSON. But the input side only accepts text/plain. That asymmetry has a few practical costs:

  - No replay. A consumer can't capture Aura's output, mutate one field, and feed it back in as the next request — they have to flatten everything into prose. Symmetric I/O makes "redo this with since='2h' instead" trivial.
  - Agent-to-agent chaining. When another A2A agent consumes Aura's structured artifacts and wants to call Aura back (or call a third agent with the same payload), it has to re-serialize to prose for input. Every hop in an agent chain loses fidelity.
  - Caller mental model. "If I see structured X coming back, I should be able to send structured X in" is the principle of least surprise for any RPC-shaped protocol. Mixing structured output with prose-only input is the kind of mismatch that produces footguns in client SDKs (which end up writing JSON.stringify(args) wrappers everywhere).
  - Logging / replay tooling. Recording A2A traffic for debugging or test fixtures is much cleaner when both directions are JSON. Right now you'd capture a JSON response and a string request, and you'd need a custom serializer for the request side.

  In short: it's not a new feature need — it's making the input contract match the output contract Aura already honors. Cheap consistency win.

crates/aura-web-server/src/a2a/agent_executor.rs
```diff
@@ -0,0 +1,406 @@
...
+            let request_id = format!("a2a_{}", task_id);
+            let cancel_token = stream_shutdown_token.child_token();
+            {
+                // keep track of this cancel token in case the a2a framework
+                // receives a cancel call and we need to propagate
+                let mut cancel_map = task_cancel_tokens.lock().await;
+                cancel_map.insert(task_id.clone(), cancel_token.clone());
+            }
+
+            let mut stream = match agent.stream(&text, vec![], cancel_token, &request_id).await {
```
justintime4tea: chat_history is hardcoded to vec![], so every A2A call hits the LLM with no prior turns. "Follow-up" requests for the same "session" appear to the agent like brand new inquiries - you can't ask "Tell me more about XYZ" because it doesn't have past knowledge of XYZ, there is no chat history. context_id → session_id only namespaces persistence paths — it doesn't replay messages into the prompt.

I think there could be two parts here we could talk through and maybe one part we could include in this initial implementation? The first part seems fairly trivial enough but the second part might need a bit more work?

The framework gives you the history via `ExecutorContext.stored_task` (populated on same-task_id continuations; `Task.history: Option<Vec<Message>>`).

### Fix (something like this?):

```rust
fn a2a_msg_to_aura(msg: &a2a::Message) -> Option<aura::Message> {
    let text = msg.parts.iter().find_map(|p| match &p.content {
        a2a::PartContent::Text(t) => Some(t.clone()),
        _ => None,
    })?;
    match msg.role {
        a2a::Role::User  => Some(aura::Message::user(&text)),
        a2a::Role::Agent => Some(aura::Message::assistant(&text)),
        _ => None,
    }
}

let chat_history: Vec<aura::Message> = ctx
    .stored_task
    .as_ref()
    .and_then(|t| t.history.as_ref())
    .map(|msgs| msgs.iter().filter_map(a2a_msg_to_aura).collect())
    .unwrap_or_default();

let mut stream = agent.stream(&text, chat_history, cancel_token, &request_id).await?;
```

**Caveats:**
- Only covers same-task_id continuations. 
- New-task-same-context_id threading leaves `stored_task = None` — needs a `TaskStore` lookup by context_id

### "new-task-same-context_id"
The upstream `TaskStore` trait gives you list(req) with `ListTasksRequest.context_id` as a first-class filter (_a2a-server/src/task_store/store.rs:29, a2a/src/types.rs:519-521)_. You just need to hold the store on the executor.

**Wiring:** `AgentExecutor::execute` doesn't receive the store, so inject it at construction. In `AuraAgentExecutor::new`, accept an `Arc<dyn TaskStore>` and share the same instance you pass to `AuraRequestHandler` in main.rs.

Rough idea:
```rust
let history_msgs: Vec<a2a::Message> = match self.task_store
    .list(&ListTasksRequest {
        context_id: Some(context_id.clone()),
        history_length: Some(50),       // cap; matches A2A spec field
        include_artifacts: Some(false),
        ..Default::default()
    })
    .await
{
    Ok(resp) => {
        let mut tasks = resp.tasks;
        tasks.retain(|t| t.id != task_id);          // exclude current task
        tasks.sort_by_key(|t| t.status.timestamp);   // oldest first
        tasks
            .into_iter()
            .flat_map(|t| t.history.unwrap_or_default())
            .collect()
    }
    Err(_) => Vec::new(),                            // fail open
};

let chat_history: Vec<aura::Message> =
    history_msgs.iter().filter_map(a2a_msg_to_aura).collect();
```

**Three things to note**:
- Pagination: `list()` is paged; if a context has >page_size tasks you'll need to loop on `next_page_token` or set a large `page_size`. The cap above (`history_length: 50`) limits per-task history, not task count.
- Ordering: the trait doesn't promise sort order, so sort client-side by `status.timestamp`. Within a task, `Task.history` is already in arrival order.
- Same-task case: `list()` excludes the current task after the filter, so combine this with the `stored_task` read for same-task_id continuations — otherwise you lose the in-progress task's own prior turns.

justintime4tea (changes requested):
I'm still working through this a bit but wanted to relay some initial feedback.

crates/aura-web-server/src/a2a/agent_executor.rs
dominic-mcallister-logdna: Good callout, let me join the values so I don't lose anything.

crates/aura-web-server/src/a2a/agent_executor.rs
dominic-mcallister-logdna: For this first iteration, this will definitely need to be a gap.

crates/aura-web-server/src/a2a/agent_executor.rs
dominic-mcallister-logdna: I think for a v1 implementation, accepting only text is fine. We should take the time to define a structured input schema for a better implementation.

crates/aura-web-server/src/a2a/agent_executor.rs
dominic-mcallister-logdna: I'll put this in place.

crates/aura-web-server/src/a2a/agent_executor.rs
```diff
@@ -0,0 +1,406 @@
...
+                context_id: context_id.clone(),
+                status: TaskStatus {
+                    state: TaskState::Working,
+                    message: None,
+                    timestamp: Some(chrono::Utc::now()),
+                },
+                metadata: None,
+            }));
+
+            let config = config.ok_or_else(|| A2AError::invalid_params("no agent configuration available"))?;
```
kristof-mattei, line 164: Think: can we move this to the top? That way we don't emit a working status, as there is nothing we can do regardless?

crates/aura-web-server/src/a2a/agent_executor.rs
```diff
@@ -0,0 +1,406 @@
...
+    }
+
+    fn cancel(&self, ctx: ExecutorContext) -> BoxStream<'static, Result<StreamResponse, A2AError>> {
+        let task_id = ctx.task_id.clone();
+        let context_id = ctx.context_id.clone();
+        let task_cancel_tokens = self.task_cancel_tokens.clone();
+        Box::pin(futures_util::stream::once(async move {
+            // a2a got a cancel call, so make sure rig cancels as well
+            // and remove the token from tracking
+            let mut cancel_map = task_cancel_tokens.lock().await;
```
kristof-mattei: Think: if the `task_id` isn't in `cancel_map`, what does it mean? Inconsistent state? Do we want to log a `WARN`? Do we want to tell the user?

crates/aura-web-server/src/a2a/agent_executor.rs
justintime4tea: Can we acknowledge the gap in the a2a-implementation doc? Some sort of footnote along the lines of "Currently, no headers are received from A2A clients due to a gap in an upstream a2a dependency - you'll need to include all desired headers, such as authorization, within AURA TOML configuration files directly."? We can link to the PR in the upstream project for: https://github.com/a2aproject/a2a-rs/pull/70

The biggest "real" impact here is lack of dynamic authorization headers from A2A client requests - that said, if MCP's expect any headers, they'll have to come from being statically defined in AURA config TOML until upstream addresses impl gap.

crates/aura-web-server/src/a2a/agent_executor.rs
dominic-mcallister-logdna: done.

crates/aura-web-server/src/a2a/agent_executor.rs
dominic-mcallister-logdna: I took at run at this and added a unit test to suit.

crates/aura-web-server/src/a2a/agent_executor.rs
dominic-mcallister-logdna: Added a note in the file. We can fast follow if/when your PR is approved. Perhaps I could have a separate commit to temporarily handle the situation so its easier to cleanup.

crates/aura-web-server/Cargo.toml
dominic-mcallister-logdna: I cleaned this up in favor of tags. Note, I pointed both at the server tag so we didn't have a bunch of weird referencing going on in our solution as both tags point to the same commit.

crates/aura-web-server/src/a2a/agent_executor.rs
dominic-mcallister-logdna: Good call, moved to the top of the async function.

crates/aura-web-server/src/a2a/agent_executor.rs
dominic-mcallister-logdna: yeah, I would not expect it to be null either. I added a warn so we can periodically scan logs if necessary and see if something is up. That'll be more sound than just ignoring it.

justintime4tea:
There were a couple issues I saw regarding active request tracking and unexpected cancelling behavior. I went ahead and proposed some changes [as a PR into your branch](https://github.com/mezmo/aura/pull/170), I hope you don't mind :pray: I felt it easier to discuss if we had an example/solution to discuss.

## Issues
Cancelling an A2A task returns `Canceled` to the caller, but the executor doesn't actually stop:

- `Agent::stream(...)` takes the cancel token as `_cancel_token` and ignores it. Token-cancel has no effect on the inner stream.
- The real MCP cancellation path is keyed off `request_id` and only fires when something calls `agent.cancel_and_close_mcp(request_id, reason)` — which the A2A executor never did. So remote MCP servers never received `notifications/cancelled` and kept running tools we'd already abandoned.
- After the stream loop, `execute()` unconditionally yields `Completed` if `success` is still true. That can race past the `Canceled` we just emitted from `cancel()` and clobber it back to Completed.
- There were also a couple of latent races around `active_requests` accounting — a fast cancel before the stream started, or a cancel mid-loop followed by natural loop-exit cleanup, could double-decrement and wrap the counter.

Net effect: the caller thinks the task is cancelled, the agent keeps streaming, the remote MCP server never knows, and the final task state can flip back to Completed.

## What changed in my proposed/suggested PR

You can find [the PR here](https://github.com/mezmo/aura/pull/170).

In `crates/aura-web-server/src/a2a/agent_executor.rs`:

- `cancel()` now does the work cancel is supposed to do: calls `agent.cancel_and_close_mcp(request_id, "A2A cancelTask")` to notify in-flight MCP calls, then fires the cancel token. To make that possible, the per-task cancel map now stores `TaskCancelEntry { token, agent, request_id }` instead of just a token.
- `execute()`'s stream loop is wrapped in `tokio::select!` against `cancel_token.cancelled()`, matching the shape used in `streaming/handlers.rs:153` and `orchestration/orchestrator.rs:251`. The loop exits the moment cancel fires rather than racing the inner stream to completion.
- The trailing `if success { yield Completed }` is gated on `!cancel_token.is_cancelled()` so a cancel landing during the final chunk doesn't get overwritten.
- `execute()` now registers/unregisters `request_id` with `aura::RequestCancellation` for parity with the OpenAI handler and so future code can address an in-flight A2A task by id.
- Replaced the manual `active_request_tracker.increment()` / `.decrement()` pair with an `ActiveRequestGuard` RAII guard living in the generator scope. Exactly one decrement per generator exit (loop break, early return, panic, consumer drop). Closes the double-decrement race that the initial impl hand-rolled version had with `cancel()`.
- Server shutdown now drives MCP cleanup for A2A tasks. The cancel token in `execute()` is a child of `stream_shutdown_token`, so a shutdown wakes the `select!` and breaks the loop — but no one was calling `cancel_and_close_mcp`. The post-loop now detects the shutdown path (cancel token fired AND our entry is still in the cancel map — `cancel()` removes its entry before firing, so a still-present entry means the parent token fired), calls `agent.cancel_and_close_mcp(request_id, "server shutdown")`, and yields a terminal `Canceled` status so the client sees a clean end rather than a silent stream close. Mirrors the OpenAI handler's `StreamTermination::Shutdown` arm.

crates/aura-web-server/src/a2a/agent_executor.rs
```diff
@@ -0,0 +1,571 @@
...
+                                ("name".into(), Value::String(tc.name.clone())),
+                                ("arguments".into(), Value::String(tc.arguments.clone())),
+                            ])),
+                            extensions: None,
+                        };
+                        yield Ok(StreamResponse::ArtifactUpdate(TaskArtifactUpdateEvent {
+                            task_id: task_id.clone(),
+                            context_id: context_id.clone(),
+                            artifact,
+                            append: Some(!first_chunk),
```
justintime4tea: I don't think `first_chunk` should gate whether we send `Some(true)` or `Some(false)` for append for this artifact or any of these artifacts except the first one above with artifact_id "response". This field (append) is used to indicate whether the update is an append for a given artifact but given that the only artifact that has a unique ID is the one with "response" for the ID, all other UpdateEvents should be sending `append: Some(false)`, right?

To determine what "append" should be, if you had other artifacts with consistent/non-unique ID's, you'd need a "first_chunk_for_this_artifact", some sort of HashMap<artifact_id, first_chunk> where you check if a given artifact itself has received a "first chunk" or not... that said, artifact_id is always unique except for the very first case above where artifact_id is always "response" so we can skip the HashMap<artifact_id, first_chunk> and should always send Some(false) for artifacts which are NOT "response", right?

crates/aura-web-server/src/a2a/agent_executor.rs
```diff
@@ -0,0 +1,571 @@
...
+            description,
+            version: VERSION.to_string(),
+            provider: None,
+            documentation_url: None,
+            icon_url: None,
+            capabilities: AgentCapabilities {
+                streaming: Some(true),
+                push_notifications: Some(false),
+                extensions: None,
+                extended_agent_card: None,
```
justintime4tea, line 101: This (extended_agent_card) is set to `None` however in `request_handler.rs` line 120-126 get_extended_agent_card is delegating to the "inner" get_extended_agent_card and may return some sort of "stub" from `DefaultRequestHandler`. We should either return a real extended agent card or drop the delegation so we aren't returning a stub and can be aligned with the `None` we're returning here for extended_agent_card.

crates/aura-web-server/src/a2a/agent_executor.rs
```diff
@@ -0,0 +1,571 @@
...
+        AgentCard {
+            name,
+            description,
+            version: VERSION.to_string(),
+            provider: None,
+            documentation_url: None,
+            icon_url: None,
+            capabilities: AgentCapabilities {
+                streaming: Some(true),
+                push_notifications: Some(false),
```
justintime4tea, line 99: `push_notifications` is set to false, but create_push_config, get_push_config, list_push_configs, and delete_push_config are delegated and routed via `request_handler.rs` lines 80-112. 

Should we flip the capability to true? or stub those _push_config methods to return unimplemented so the card matches reality?

crates/aura-web-server/src/a2a/agent_executor.rs
```diff
@@ -0,0 +1,571 @@
...
+impl AuraAgentExecutor {
+    pub fn new(app_state: Arc<AppState>, task_store: SharedTaskStore) -> Self {
+        Self {
+            app_state,
+            task_store,
+            task_cancel_state: Arc::new(Mutex::new(HashMap::new())),
+        }
+    }
+
+    fn resolve_config(&self) -> Option<aura_config::Config> {
```
justintime4tea, line 49: We may need to revisit how we approach/support multi-config for A2A so that A2A clients don't have to be "locked" to specific deployments of AURA that have specific/desired agent config. 

**Issue:** right now, as an A2A client, you are only able to use the default or first config AURA has been configured to use - you'd have to launch a whole new AURA specifically for an A2A client if you can't change the existing AURA instance and it isn't serving the desired config as default or first.

Perhaps for the first/initial implementation we leave this as is but we make sure to document this in the `a2a_implementation.md` letting A2A clients know of the limitation?

**Future thinking?**

I think the ideal solution is to have a one-to-one mapping between A2A AgentCard and AURA config where each AURA config served by an AURA web server has an AgentCard created for it and is served via it's own route - this lets an A2A client connect to a single AURA server similar to how the HTTP/chat_completions clients use "model" req field, and use any of the "configured agents".

- JSON-RPC at 
  - `/aura-config-alias/a2a/v1/rpc`
- REST at 
  - `/aura-config-alias/a2a/v1/message:send`
  - `/aura-config-alias/a2a/v1/tasks/`
  - etc...
- Agent card at 
  - `/.well-known/agent-card-for-one-config.json`
  - `/.well-known/agent-card-for-a-diff-config.json`
  - `/.well-known/agents.json`
    - Essentially just a "JSON array" of all the individual cards
    - This could be used as a "single" call to enumerate all known agent cards

There are two open issues/questions and a discussion on the `a2aproject/A2A` project related to multi-agent serving from single host/server:
- https://github.com/a2aproject/A2A/issues/221
- https://github.com/a2aproject/A2A/issues/178
- https://github.com/a2aproject/A2A/discussions/166

crates/aura-web-server/src/main.rs
```diff
@@ -231,6 +260,7 @@ async fn run() -> std::io::Result<()> {
         ))
         .with_state(app_state);
 
+    let app = aura_router.fallback_service(a2a_router);
```
justintime4tea: I don't think `Router::fallback_service` is what we want to use here - we should be able to use `Router::merge`. 

```rust
let app = Router::new() /// <-- change this to "let app" instead of "let aura_router"
        .route("/health", get(handlers::health))
        .route("/v1/models", get(handlers::list_models))
        .route("/v1/chat/completions", post(handlers::chat_completions))
        .layer(TraceLayer::new_for_http())
        .layer(middleware::from_fn_with_state(
            app_state.clone(),
            shutdown_guard,
        ))
        .with_state(app_state)
        .merge(a2a_router);
       /// ^--- notice the merge

/// you can then just omit this line below
/// let app = aura_router.fallback_service(a2a_router);
```

The other `a2a-rs-server` crate you had been using served its own `/health` route and the easy "workaround" was to use `Router::falback_service` but the crate you switched to, from a2aproject, does not serve a `/health` route and is safe to "merge".

After switching to using `merge` we can remove the comment mentioning the "/health" endpoint, conflict, and "fall through" since there isn't a conflict anymore.

crates/aura-web-server/src/a2a/agent_executor.rs
justintime4tea: Nevermind - the `DefaultRequestHandler` returns:
```rust
 Err(A2AError::unsupported_operation(
          "extended agent card not configured",
      ))
```
which makes sense. I think I was confused because we were delegating to `DefaultRequestHandler` and thought it may return some sort of default value.

I'll close as resolved :pray:

crates/aura-web-server/src/a2a/agent_executor.rs
justintime4tea: Nevermind, this was similar to the extended agent card delegation - `DefaultRequestHandler` will just reject the calls with the appropriate errors. Will close this as resolved as well :pray:

crates/aura-web-server/src/a2a/agent_executor.rs
```diff
@@ -0,0 +1,571 @@
...
+                    metadata: None,
+                }));
+            }
+
+            // _request_guard drops at end of generator scope → exactly one decrement.
+
+            // Skip Completed if cancel() or the shutdown path already emitted Canceled —
+            // yielding here would clobber it.
+            if success && !cancel_token.is_cancelled() {
+                yield Ok(StreamResponse::StatusUpdate(TaskStatusUpdateEvent {
```
justintime4tea, line 431: Might be worth documenting in `a2a_implementation.md` that there are _two_ "termination signals" or events that indicate termination - The final block emits both `last_chunk: true` (Lines: 336-343) artifact AND TaskState::Completed status (Lines: 394-403, here). 

Clients should consider that both should be received to indicate "true" termination/finality. Until they've received both, the stream is still "active".

crates/aura-web-server/src/main.rs
dominic-mcallister-logdna: Fixed.

crates/aura-web-server/src/a2a/agent_executor.rs
dominic-mcallister-logdna: I added these specific links into the documentation for note.

crates/aura-web-server/src/a2a/agent_executor.rs
dominic-mcallister-logdna: This is a good callout. I also discovered that "append" isn't really appending in the in-memory store, so you'll find an override in my override store that'll group the artifacts together.

Shearerbeard:
I've spent some time with the code but found the tooling for manually testing A2A to be a bit lacking. https://github.com/a2aproject/a2a-rs has a basic CLI bundled for trying these things that I'll vet in a few for better verification. I've also looked at putting together with a claude a light internal tool on top to get a bit better visibility as the CLI is a bit thin in functionality.

justintime4tea:
While testing with `a2acli` from a2aproject, used to debug/test a2a impls, I was unable to do anything other than request the agent card. This is because the agent card, which drives a2a clients, needs to have absolute URL's presented in the card.

Here's a PR that should resolve the issue and make the A2A impl work with a2acli and other A2A clients:
https://github.com/mezmo/aura/pull/173

Something else I spotted, when you do connect using `a2acli`, run `send` to interact with AURA and then run `list-tasks` to see the tasks, the token counts in metadata are showing 0's:

image

justintime4tea:
Can we cargo-feature-gate A2A so that AURA Web Server can be built without A2A? There may be deployments where people don't want to serve A2A at all and I feel it should be an "opt-in". We can include it in the docker image for AURA but people should be able to build/deploy AURA without A2A.

dominic-mcallister-logdna:
> unts in metadata are showing 0's:

Non-issue. Tokens are only included if provided from aura itself in the final events and when tokens are used. That prompt has no bones.

justintime4tea (approved):
Looking good @dominic-mcallister-logdna!

@Shearerbeard did you get a chance to run this through a2acli and your A2A inspection tool you mentioned?

Shearerbeard:
> Looking good @dominic-mcallister-logdna!
> 
> @Shearerbeard did you get a chance to run this through a2acli and your A2A inspection tool you mentioned?

Not yet, I'll see if I can between flights in a few.

Shearerbeard (approved):
This seems to work with the cli - we can re-evaluate it with the more intensive testing harness Im working on later.

Issue #114, opened 2026-05-12
aura-cli does not display aura.reasoning SSE events


Shearerbeard:
## Problem

The server emits `aura.reasoning` SSE events when both `AURA_CUSTOM_EVENTS=true` and `AURA_EMIT_REASONING=true` are set, but the CLI does not render them. Reasoning output is useful for understanding model thought process, especially during orchestration where the coordinator makes routing and synthesis decisions.

## Expected Behavior

When reasoning events are enabled, the CLI should display reasoning chunks in the output (e.g., in a distinct style or collapsible section).

Shearerbeard:
cc @justintime4tea

justintime4tea:
What about showing the reasoning at the "top-level" for coordinator reasoning and show in-line entries for reasoning, interleaved with tool calls, under the worker/task output?

<img width="1265" height="922" alt="Image" src="https://github.com/user-attachments/assets/fadb2eb3-5234-4441-93b1-94393e5344c6" />

PR #165, opened 2026-05-20
feat(aura-cli): display LLM reasoning output for coordinator and workers

justintime4tea:
Shows "Reasoning: ..." output both at the top-level of the output, for the coordinator, and in-line with a given task for worker reasoning.

Here is an example of running against an orchestration enabled AURA config using Anthropic `claude-sonnet-4-6` model.

image

Here is an example of running against a single-agent AURA config using Anthropic `claude-sonnet-4-6` model.

image

Additionally, event names are consolidated to a single source of truth.

Base aura.* event-name constants now live solely in aura-events::event_names, beside the existing orchestration::event_names (kept as separate modules). The duplicate copy in aura/src/stream_events.rs was removed, and AuraStreamEvent::event_name() returns these consts rather than inline literals.

aura-cli no longer hard-codes "aura.*" strings: the dispatch match, REPL handlers, and stream panel reference the consts via a local event_names re-export. The orchestrator handler now matches full event names directly instead of stripping the "aura.orchestrator." prefix and comparing the tail.

aura-web-server consumers import event_names straight from aura-events rather than the aura facade re-export, so the base names have one definition and one import path.

Ref: GH-114

### DIFF
--- a/crates/aura-cli/src/api/stream.rs
+++ b/crates/aura-cli/src/api/stream.rs
@@ -11,6 +11,7 @@ use reqwest::Response;
 use aura_events::AuraStreamEvent;
 
 use crate::api::types::{AccumulatedToolCall, ChatCompletionChunk};
+use crate::event_names;
 
 /// Result of processing a stream — either a text response or tool calls.
 #[derive(Debug)]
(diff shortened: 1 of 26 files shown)


esatterwhite:
In the commit message you can reference a github issue the same way `#` or `GH-` is legal

`Ref: #114`

`Fixes: GH-114`

both are legal / satisfy the commit linter. and still get all the deep linking benefits in all the places.

your commit message has `Ref: 114` with out an github identifier - thats why I'm pointing it out.
I'd recoment using `GH-` just because `#` is also a comment char in commit messages

crates/aura-cli/src/api/stream.rs
```diff
@@ -735,6 +769,65 @@ mod tests {
...
+    async fn aura_reasoning_callback_carries_worker_agent_id() {
+        // In orchestration mode the server emits aura.reasoning with the
+        // worker's agent_id so the CLI can attribute reasoning to a worker.
+        let data = serde_json::json!({
+            "content": "Analyzing the logs.",
+            "agent_id": "log_worker",
+            "parent_agent_id": "coordinator",
+            "session_id": "s1"
+        });
+        let events = vec![sse("aura.reasoning", &data.to_string()), sse("", "[DONE]")];
```
Shearerbeard: Thinking ahead - I wonder if we should canonicalize some of these strings as consts along with the SSE event structs

crates/aura-cli/src/api/stream.rs
```diff
@@ -802,6 +895,7 @@ mod tests {
             |_, _| {},
             |_, _, _, _| {},
             |_, _| {},
+            |_, _, _| {},
             |_, _| {},
             |_, _| {},
```
Shearerbeard: Just curiosity here - would this be a candidate for a non matching _ block since it seems like were only acting on certain combos?

crates/aura-cli/src/api/types.rs
```diff
@@ -298,6 +298,19 @@ pub enum DisplayEvent {
...
+    /// the worker name (e.g. `"log_worker"`) when emitted by an orchestration worker.
+    /// `content` holds the final accumulated text for the whole reasoning stretch
+    /// (one `DisplayEvent::Reasoning` per agent-contiguous stretch, not per delta).
+    /// `fields` carries the raw SSE payload (agent_id, content, parent_agent_id,
+    /// session_id, trace_id, ...) so /expand can render a complete fields tree.
+    Reasoning {
+        content: String,
+        agent_id: String,
+        #[serde(default)]
+        fields: BTreeMap<String, serde_json::Value>,
```
Shearerbeard, line 312: Btreemap because your preserving order?

crates/aura-cli/src/oneshot.rs
```diff
@@ -95,6 +95,11 @@ pub fn run_oneshot(
                     |_tool_id, _tool_name| {},
                     |_tool_id, _tool_name, _duration, _result: Option<&str>| {},
                     |_prompt_tokens, _completion_tokens| {},
+                    // on_reasoning — no-op in oneshot mode (reasoning chunks
+                    // would interleave with the rendered response; the user
+                    // asked for the answer, not the thought process)
+                    |_content, _agent_id, _fields| {},
+                    // on_raw_event — no-op in oneshot mode
```
Shearerbeard: Similar question as above. Again just an observation from a shallow diff

crates/aura-cli/src/ui/event_replay.rs
```diff
@@ -380,8 +419,9 @@ pub fn replay_event_log_global() {
                         Some(description.as_str())
                     });
                     if let Some(text) = reasoning_text {
-                        let display = if text.len() > 120 {
-                            format!("{}...", &text[..117])
+                        let display = if text.chars().count() > 120 {
+                            let prefix: String = text.chars().take(117).collect();
+                            format!("{prefix}...")
```
Shearerbeard: Only concerning piece I'm seeing here - is this truncation utf8/emoji safe? We have some special handling with safe_truncation somewhere in aura that we use a lot but you approach this one a bit differently so I'm not sure if it applies.

Shearerbeard (changes requested):
Mostly just questions and observations - one truncation concern we should look into pre merge. It runs absolutely amazing.

crates/aura-cli/src/api/types.rs
justintime4tea: Yes, preserving order. With BTreeMap the /expand fields tree renders consistently and tests aren't subject to HashMap's randomized iteration. If we ever needed to preserve the original JSON field order we could switch to IndexMap, but stable sorted order is what I was going for here.

crates/aura-cli/src/ui/event_replay.rs
justintime4tea: It's UTF-8 safe because I'm using `.chars().take()/.collect()`, which only ever cuts on character boundaries, so it can't split a multi-byte sequence and panic.

`safe_truncate` solves the byte-budget case (it needs `floor_char_boundary` because it slices by byte index) but here I'm truncating to a visible-character count for a preview, so the character-based approach is the right unit and `safe_truncate` doesn't quite fit.

There is something though that neither handles and that is multi-scalar emoji (ZWJ/skin-tone/flags) being cut mid-grapheme. I didn't come up with this, Claude did when I asked it to further look into my truncation to see if it could find issues with it that I hadn't and to answer your question of "is it emoji safe". This, it turns out, is purely cosmetic - no crash, but multi-scalar emoji's aren't safe (either with this or safe_truncate). I could switch (this, not safe_truncate) to unicode-segmentation::graphemes though if we want to be sure that we not cut off mid-emoji :)

crates/aura-cli/src/api/stream.rs
justintime4tea: I think we might have some consts already for these event names if I recall... let me do some cleanup on this.

crates/aura-cli/src/api/stream.rs
justintime4tea: There's definitely room for improvement here with regards to currently requiring multiple no-op handlers to fill the function params. I could switch to using a trait that has default no-ops and then you just supply the handler you wish to override. I'll do some cleanup here.

crates/aura-cli/src/oneshot.rs
justintime4tea: Similar to other comment - I'll switch to using a trait that has default no-ops. I'll clean this up.

crates/aura-cli/src/api/stream.rs
Shearerbeard: Sure - cleanup could always be prioritized as a GH too. Up to your discretion here

crates/aura-cli/src/ui/event_replay.rs
Shearerbeard: LGTM

crates/aura-cli/src/api/stream.rs
Shearerbeard: Sure - this could always tie into another GH issue if it becomes a bit of a rabbit hole.

justintime4tea: I was able to get some of this resolved but I think we can further resolve it in a new issue for refactoring events into aura-events and out of aura crate. Similar to the config structs in aura, we need to move things out of aura crate into the respective sub-crates.

Shearerbeard (approved):

PR #179, opened 2026-05-28
Draft: feat(telemetry): notice-gated three-state PostHog product telemetry

charlesjohnson:
Implements anonymous product telemetry as a **notice-gated, three-state
consent model** (not opt-out / send-by-default), built on a new
`aura-telemetry` workspace crate. The data *pipe* — sealed
`PropertyValue` allow-list + `#[derive(Event)]` anti-PII gate, install-id,
background PostHog `/batch` sink, JSONL inspection log, `/telemetry`
commands, container persistence — is shared across every surface; the
*gate* in front of it is the consent state machine described below.

## The three states

- **Unknown** — no preference recorded. Telemetry is **held**: events are
  written to the local inspection log only (`sent: false`,
  `not_sent_reason: "Unknown"`), **never queued to the sink and never
  backfilled** after a later transition. This satisfies the spec's hard
  constraint: *must not send in the Unknown state.*
- **Enabled** — a first-run notice was presented in an interactive
  surface and not opted out of, or an operator explicitly enabled via
  config/env. May send.
- **Disabled** — kill switch or explicit opt-out. Held, never sends.

## Consent flow

- **CLI REPL, first run, no preference** → a one-time notice is printed
  **before** the welcome banner (states what is collected, links to
  `docs/telemetry.md`, gives the disable mechanism). The user's **first
  input** is the gate: an explicit `/telemetry disable` persists
  **Disabled**; anything else transitions Unknown→Enabled at runtime,
  persists **Enabled**, and sends from there. Unknown-era events captured
  before the transition are held and **never backfilled**.
- **One-shot `--query`** → never participates: no notice, no telemetry
  handle is handed to the backend, and **no consent header is propagated**
  — even if the user previously opted in.
- **Server / any non-interactive surface** → no notice; Enabled only via
  explicit `[telemetry] enabled = true` / `AURA_TELEMETRY_ENABLED=true`,
  otherwise Unknown (events held, inspectable, never sent).
- **HTTP consent propagation (CLI → server)** → an **Enabled** CLI
  attaches `X-Aura-Telemetry-Consent: enabled` to its requests. A server
  in **Unknown** honors it and transitions Unknown→Enabled at runtime
  (process-lifetime only, **not persisted** — re-established per
  connection, avoids writing the often read-only server config). A
  **Disabled** server ignores the header, so an operator opt-out always
  wins. The two processes keep separate install-ids; the envelope
  `aura_source` (`cli` vs `web-server`) distinguishes them downstream.

## Resolution precedence (highest first)

1. **Hard holds → Disabled**, even over an explicit enable:
   `DO_NOT_TRACK`, `AURA_TELEMETRY_DISABLED`, CI vars, cargo-test markers.
   Keeps CI/dev out of the install count regardless of misconfiguration.
2. `AURA_TELEMETRY_ENABLED=false` / config `enabled = false` → **Disabled**.
3. `AURA_TELEMETRY_ENABLED=true` / config `enabled = true` / recorded
   preference `true` → **Enabled**.
4. Otherwise → **Unknown**.

The recorded preference is the existing `[telemetry] enabled` field as
`Option<bool>` (`None` = Unknown, `Some(true)` = Enabled,
`Some(false)` = Disabled), with an AND-merge across global/project
`cli.toml` layers so a user's `/telemetry disable` can't be silently
undone by a project-shipped `enabled = true`.

## Runtime state machine

- Init spawns the sink eagerly only when **Enabled**; **Unknown** stashes
  the sink params and spawns nothing. `TelemetryHandle::enable()`
  transitions Unknown→Enabled at runtime via a captured `runtime::Handle`
  and starts the sink on the fly.
- `/telemetry enable` / `disable` take effect **immediately** for the
  session and are symmetric — a runtime `/telemetry disable` can be undone
  by `/telemetry enable`. The one exception: a startup kill switch leaves
  no captured sink and can **never** be resurrected in-process; `enable()`
  returns an honest "held until next launch" outcome rather than falsely
  claiming the session is active.

## Reused substrate (privacy & audit)

- **Structural anti-PII gate**: a sealed `PropertyValue` allow-list and
  `#[derive(Event)]` proc-macro that compile-time-reject free-form fields
  (`String`, numeric, etc.) — 3 `trybuild` compile-fail snapshots lock it.
- **Install-id**: race-safe UUID publication via `std::fs::hard_link`,
  **source-aware** persistence — the server roots it under the mounted
  `memory_dir` (durable across container recreation), the CLI under
  `~/.aura/install-id` — with a one-time warning when telemetry is active
  but the id won't survive a restart.
- **Inspection log**: JSONL with line-based rotation, recording the
  **real POST outcome** (`sent: true` only after 2xx, otherwise
  `PostFailed(<category>)` / `ChannelFull`), so the audit trail never
  lies. Unknown-state rows are the "what *would* be sent" payloads, so
  users can inspect both what was sent **and** what is being held.
- **Sink**: background-task PostHog `/batch` POSTs with a `post_timeout`
  bounded strictly below the shutdown budget so in-flight events still
  reach their post-flush inspection-log writes. Network failures are
  silent at `debug` and never alter Aura's behaviour.
- **Surfaces**: `aura-web-server` emits `server_started` and exposes a
  loopback-only `GET /telemetry/recent`; `aura-cli` emits
  `cli_session_started` and adds `/telemetry status|recent|enable|disable`.

## Testing & verification

510 tests pass across `aura-telemetry`, `aura-cli`, `aura-web-server`,
and `aura-config` under default features (326 of the `aura-cli` suite
also under `--features standalone-cli`), including: the 3 `trybuild`
anti-PII snapshots; wiremock coverage of wire shape, disabled/held modes,
5xx failure, channel-full, and slow-endpoint shutdown; the **no-backfill**
guarantee (an Unknown-era event is not sent after `enable()`); the honest
`enable()` outcomes; source-aware install-id resolution; and the
one-shot-never-propagates-consent contract. `cargo clippy -D warnings`
(default features) and `cargo fmt --check` are clean. The full
user-facing contract is documented in `docs/telemetry.md`.

## Notes

- Rebased onto current `main` (includes the A2A feature); resolved the
  `axum 0.8` bump in `aura-web-server`.
- A `docs/adr/` convention plus an ADR for the notice-gated telemetry
  decision are held back for a separate follow-up PR to keep this one
  scoped to the implementation.

---

🤖 Prepared with Claude Code. Design, implementation, tests, and
review-iteration were done collaboratively with a human author who
reviewed and verified the work end-to-end against the product spec. No AI
co-authorship is claimed on the commits, per the project CLA.


esatterwhite:
It isn't necessary to reference a phantom issue. ( `Ref: LOG-000000 (no ref)` ). Rather than create confusion around a related issue that doesn't exist, it would be better to omit it, Or create a Github issue that describe the problem, and references that

`Fixes: #1`
`Resolves: GH-1`
`Ref: #1`
`Ref: LOG-1234`

these are all valid

crates/aura-cli/src/repl/commands.rs
```diff
@@ -71,6 +71,259 @@ pub(crate) fn handle_help() {
...
+pub(crate) fn format_telemetry_recent(
+    telemetry: &aura_telemetry::TelemetryHandle,
+    n: usize,
+) -> String {
+    let Some(log) = telemetry.inspection_log() else {
+        return "inspection log is disabled (AURA_TELEMETRY_LOG_EVENTS=0).".to_string();
+    };
+    match log.recent(n) {
+        Ok(events) => {
+            if events.is_empty() {
```
kristof-mattei: Use Ok([]) and Ok(list), that way we can move the empty check to the match statement.

crates/aura-cli/src/repl/commands.rs
```diff
@@ -71,6 +71,259 @@ pub(crate) fn handle_help() {
...
+                return "no telemetry events recorded yet.".to_string();
+            }
+            let mut out = format!("last {} event(s):", events.len());
+            for evt in events {
```
kristof-mattei, line 255: This is very heavy on the allocations. I think we can do better with the appending. 

We can move the to

```
                out.push_str(&format!(
                    "\n  {}  {}  {}",
                    evt.ts.format("%Y-%m-%dT%H:%M:%SZ"),
                    evt.event,
                    suffix
                ));
```

above the `if`, and then inside the if we can do `push_str` which avoids the allocation of `[sent]`, `[not sent]`, and for `format!("[not sent — {r}]")` we can do `out.push_fmt("[not sent — {}]", r)`.

charlesjohnson:
@esatterwhite This is in the CLAUDE.MD right now, because the commitlint required a Ref. Is that no longer mandatory? We can update to make the ref optional.

crates/aura-cli/src/config.rs
```diff
@@ -308,6 +354,105 @@ fn upsert_top_level_string(content: &str, key: &str, value: &str) -> String {
...
+///
+/// Backs `/telemetry disable` (and a future `/telemetry enable`). The
+/// upsert is sectioned: it locates `[telemetry]` and either replaces an
+/// existing `enabled = …` line or inserts one immediately under the
+/// header. If the section is missing, it is appended at end-of-file
+/// with a blank-line separator. Other sections, sibling keys, and
+/// comments are preserved.
+pub fn save_telemetry_enabled_to_global_cli_toml(enabled: bool) -> Result<()> {
+    let dir = global_aura_dir().ok_or_else(|| {
+        anyhow::anyhow!("could not determine ~/.aura/ (no home directory available)")
```
kristof-mattei: This will fail when `aura` is ran in a read-only container.

crates/aura-cli/src/config.rs
```diff
@@ -57,10 +62,42 @@ impl FileConfig {
                 .or(self.enable_final_response_summary),
             style: other.style.or(self.style),
             log_file: other.log_file.or(self.log_file),
+            telemetry: merge_telemetry(self.telemetry, other.telemetry),
         }
     }
 }
 
+fn merge_telemetry(
```
kristof-mattei, line 77: This seems highly custom code to merge 2 toml files. It might be prudent to find out if there are 3rd party crates available so we don't have to maintain this code.

crates/aura-cli/src/config.rs
```diff
@@ -308,6 +354,105 @@ fn upsert_top_level_string(content: &str, key: &str, value: &str) -> String {
...
+/// - If `[section]` is missing, append `\n[section]\nkey = <value>` at
+///   end-of-file (a blank line separates it from any prior content).
+fn upsert_section_bool(content: &str, section: &str, key: &str, value: bool) -> String {
+    let val_str = if value { "true" } else { "false" };
+    let new_line = format!("{key} = {val_str}");
+    let section_header = format!("[{section}]");
+
+    let mut lines: Vec<String> = content.lines().map(String::from).collect();
+
+    let section_idx = lines.iter().position(|l| l.trim() == section_header);
```
kristof-mattei: Nit: `_idx` suffix is a Python thing. I prefer `_index`.

crates/aura-cli/src/config.rs
```diff
@@ -308,6 +354,105 @@ fn upsert_top_level_string(content: &str, key: &str, value: &str) -> String {
...
+    match section_idx {
+        Some(start) => {
+            // Bound the search to this section: stop at the next
+            // `[other]` header or end-of-file.
+            let end = lines
+                .iter()
+                .enumerate()
+                .skip(start + 1)
+                .find(|(_, l)| l.trim_start().starts_with('['))
+                .map(|(i, _)| i)
```
kristof-mattei: Use https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map

crates/aura-cli/src/config.rs
```diff
@@ -308,6 +354,105 @@ fn upsert_top_level_string(content: &str, key: &str, value: &str) -> String {
...
+                    format!("{new_line}  {comment}")
+                };
+            } else {
+                lines.insert(start + 1, new_line);
+            }
+        }
+        None => {
+            // Append a fresh section. Blank-line separator improves
+            // round-tripping (matches what `toml` would produce).
+            if !lines.is_empty() && lines.last().map(|l| !l.is_empty()).unwrap_or(false) {
```
kristof-mattei: `lines.is_empty()` is redundant.

crates/aura-cli/src/repl/commands.rs
```diff
@@ -71,6 +71,259 @@ pub(crate) fn handle_help() {
...
+                .next()
+                .and_then(|s| s.parse::<usize>().ok())
+                .unwrap_or(20);
+            format_telemetry_recent(telemetry, n)
+        }
+        "disable" => format_telemetry_disable_result(
+            crate::config::save_telemetry_enabled_to_global_cli_toml(false),
+        ),
+        other => format!(
+            "Unknown /telemetry subcommand: {other}\nAvailable: status, recent [N], disable"
```
kristof-mattei: Using a list like this has the potential that isn't not updated in the future. 

Also I'd like to see the commands being converted into some enum instead of passing around bare text.

crates/aura-cli/src/repl/commands.rs
```diff
@@ -71,6 +71,259 @@ pub(crate) fn handle_help() {
...
+        ),
+    };
+    println!("{body}");
+    redraw_input_frame();
+}
+
+pub(crate) fn format_telemetry_disable_result(result: anyhow::Result<()>) -> String {
```
kristof-mattei: Better to create a custom error type and then implement std::fmt::Display for it.

crates/aura-cli/src/config.rs
charlesjohnson: Pushing back gently on swapping this for a third-party crate, for two reasons:

1. **It's pre-existing code.** `merge_over` (the field-wise `Option::or` layering) is already on `main` — this PR only adds the one `telemetry` field to it. Replacing the merge engine would be a refactor of existing behavior unrelated to telemetry, so I'd rather not fold it into this PR.

2. **The merge isn't actually generic — and a generic crate would reintroduce a bug.** `merge_enabled` is a deliberate *kill-switch AND*: a project `.aura/cli.toml` with `[telemetry] enabled = true` must **not** override a global `~/.aura/cli.toml` with `enabled = false`. That's the documented privacy guarantee — a user's `/telemetry disable` (which writes `enabled = false` globally) can't be silently undone by a project file. A drop-in deep-merge crate does last-writer-wins, which would flip that back. So a crate would have to special-case `enabled` anyway, which undercuts the "less code to maintain" goal.

If the team wants to move config layering onto a dedicated crate more broadly, `config` and `twelf` are already in the tree and that'd be a reasonable standalone refactor — happy to file it as a follow-up — as long as it preserves the `enabled` AND-merge semantic. I just don't think it belongs in this PR.

charlesjohnson:
superseded by #270.

PR #181, opened 2026-05-29
feat(mcp): bring stdio transport to parity with http streamable and sse

### LINKED ISSUE #163: Bring STDIO MCP transport to parity with HTTP streamable and SSE
## Summary

Bring the STDIO MCP transport to feature parity with the HTTP streamable
and SSE transports. STDIO currently uses a legacy code path that lacks
progress notifications, cancellation propagation, per-server tool tracking,
and scratchpad integration. Additionally, a connection lifetime bug causes
the child process to be killed after initialization, making tool execution
unreliable.

## Motivation

STDIO is the original MCP transport for local, process-based servers
(Python scripts, Node.js tools, CLI wrappers). While many MCP servers
have migrated to HTTP streamable, STDIO remains the lowest-friction
transport for local development and single-machine deployments.

Currently STDIO tools are registered via Rig's legacy `McpTool` path
(`add_rmcp_tools`) rather than AURA's unified `McpClient` / `McpToolAdaptor`
pattern. This means:

- No `aura.progress` SSE events for STDIO tools
- No `notifications/cancelled` propagation on client disconnect
- No `InFlightRequests` cancellation tracking
- No per-server scratchpad thresholds (tools stored in flat `Vec`, no
  server-name mapping)
- Child process killed after init due to `RunningService` going out of
  scope (commented-out `// self.clients.push(client);` at mcp.rs:591-592)

## Approach

STDIO already uses `rmcp::transport::TokioChildProcess`, which implements
`Transport<RoleClient>`. The `McpClient::from_transport()` method (added in
the SSE work) already accepts any `T: Transport<RoleClient>`. This means
STDIO can plug directly into the unified client path without a new transport
implementation -- the changes are wiring, not new transport code.

The fix:
1. Replace the legacy `client_info.serve(transport)` path in
   `try_connect_stdio()` with `McpClient::from_transport(transport, url)`
2. Add `stdio_clients` / `stdio_tools` storage to `McpManager` (parallel
   to `streamable_clients` / `sse_clients`)
3. Register STDIO tools via `McpToolAdaptor` instead of Rig's `McpTool`
4. Wire STDIO clients into cancellation, request tracking, and tool
   iteration loops
5. Remove the legacy `tool_definitions: Vec<(Tool, ServerSink)>` and
   `tools: Vec<McpTool>` fields once migrated

## Configuration

No configuration changes needed. Existing `transport = "stdio"` configs
continue to work:

```toml
[mcp.servers.my_stdio_server]
transport = "stdio"
cmd = ["python", "-m", "my_mcp_server"]
args = ["--verbose"]
env = { LOG_LEVEL = "debug" }
description = "Local STDIO MCP server"
```

STDIO servers will gain the same progress, cancellation, and scratchpad
capabilities as `http_streamable` and `sse` transports.

Shearerbeard:
Fixes #163 - stdio isn't a huge priority for big production backend systems but having it available makes evaluating and playing with AURA a lot more accessible

### DIFF
--- a/crates/aura/src/orchestration/orchestrator.rs
+++ b/crates/aura/src/orchestration/orchestrator.rs
@@ -1712,9 +1712,11 @@ Assign tasks to the worker whose tools best match the required operations."#,
             }
         }
 
-        // Collect from legacy tool definitions (rmcp::model::Tool)
-        for (tool, _) in &mcp_manager.tool_definitions {
-            names.push(tool.name.to_string());
+        // Collect from STDIO tools
+        for tools in mcp_manager.stdio_tools.values() {
+            for tool in tools {
+                names.push(tool.name.to_string());
+            }
         }
 
         // Remove duplicates while preserving order
@@ -1755,10 +1757,12 @@ Assign tasks to the worker whose tools best match the required operations."#,
             }
         }
 
-        // Collect from legacy tool definitions (same rmcp::model::Tool type)
-        for (tool, _) in &mcp_manager.tool_definitions {
-            let schema_value = serde_json::Value::Object((*tool.input_schema).clone());
-            schemas.insert(tool.name.to_string(), schema_value);
+        // Collect from STDIO tools
+        for tools in mcp_manager.stdio_tools.values() {
+            for tool in tools {
+                let schema_value = serde_json::Value::Object((*tool.input_schema).clone());
+                schemas.insert(tool.name.to_string(), schema_value);
+            }
         }
 
         schemas
@@ -1839,10 +1843,12 @@ Assign tasks to the worker whose tools best match the required operations."#,
                 }
             }
 
-            // Collect from legacy tool definitions
-            for (tool, _) in &mcp_manager.tool_definitions {
-                if let Some(ref desc) = tool.description {
-                    descriptions.insert(tool.name.to_string(), desc.to_string());
+            // Collect from STDIO tools
+            for tools in mcp_manager.stdio_tools.values() {
+                for tool in tools {
+                    if let Some(ref desc) = tool.description {
+                        descriptions.insert(tool.name.to_string(), desc.to_string());
+                    }
                 }
             }
         }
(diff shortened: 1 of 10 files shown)


crates/aura/src/orchestration/orchestrator.rs
```diff
@@ -1712,9 +1712,11 @@ Assign tasks to the worker whose tools best match the required operations."#,
             }
         }
 
-        // Collect from legacy tool definitions (rmcp::model::Tool)
-        for (tool, _) in &mcp_manager.tool_definitions {
-            names.push(tool.name.to_string());
+        // Collect from STDIO tools
+        for tools in mcp_manager.stdio_tools.values() {
+            for tool in tools {
+                names.push(tool.name.to_string());
```
dominic-mcallister-logdna, line 1718: I see there is a check for duplicate tool names below. Should we add some sort of warning that we have ambiguous tool names to invoke or should things be namespaced by a schema name? This situation may not be common, but certainly a foot-gun we can run into.

Shearerbeard: This is a great point and something I've considered tackling. The code your seeing is a pre-existing pattern with the other server types but I think the idea of namespacing could be interesting but isn't something that should be done without a bit of strategy. Many system prompts have verbatim tool usage guidelines "run 'x' before 'y'" so there is some backwards compat concerns. We also have mcp filter by explicit tool names that this would effect. The upshot is that it would prevent collisions and give us something easier to glob match on when writing mcp filters - 'mezmo_*' or 'k8_get_*' for example. The problem is that we might need to consider how the namespaces are defined in our config to keep the tool names from getting to long. My proposal is we leave this pattern as is for now and file an issue around  re-exploring the namespace idea. I'm going to audit to see if there is a good place to make this more clear in the README

Shearerbeard: Filed #186 and added a README warning about the conflict behavior

justintime4tea (approved):

dominic-mcallister-logdna (approved):

PR #185, opened 2026-06-02
fix(aura-cli): display reasoning events in single-agent mode

teriyakichild:
## Summary
- `aura.reasoning` SSE events were silently dropped in single-agent mode because the orchestrator event handler early-returns for any event without the `aura.orchestrator.` prefix
- Adds early dispatch for `aura.reasoning` (same pattern as `aura.progress` and `aura.worker_phase`) so the thinking animation sub-line shows reasoning content regardless of mode

## Test plan
- [ ] Run CLI against an OpenRouter reasoning model (e.g. `arcee-ai/trinity-large-thinking`) and verify thinking animation displays reasoning text
- [ ] Verify orchestration mode reasoning still works (no regression)
- [ ] `cargo clippy -p aura-cli`

### DIFF
--- a/crates/aura-cli/src/repl/loop.rs
+++ b/crates/aura-cli/src/repl/loop.rs
@@ -712,6 +712,13 @@ pub fn run_repl(
                                     return;
                                 }
 
+                                if event_name == "aura.reasoning" {
+                                    if let Some(content) = val.get("content").and_then(|v| v.as_str()) {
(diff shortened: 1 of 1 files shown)


Shearerbeard:
@justintime4tea tagging you to see if this one is relevant based on your big reasoning PR being close to merging

justintime4tea:
Running current main branch we do have reasoning even when using single agent - closing this ticket as no longer necessary.

image

Issue #193, opened 2026-06-04
feat(orchestration): support fallback_tool_parsing for coordinator and workers


charlesjohnson:
## Summary

In orchestration mode, `fallback_tool_parsing` is silently ignored: the coordinator and workers are constructed with it hardcoded to `false`, regardless of `[agent.llm].fallback_tool_parsing`. Single-agent mode honors the setting. Now that the quickstart defaults to orchestration mode, local/Ollama models that depend on fallback tool-call parsing cannot use the headline mode — they must opt out to single-agent mode.

## Evidence

- Worker agent built with `fallback_tool_parsing: false` — `crates/aura/src/orchestration/orchestrator.rs:749`
- Coordinator agent built with `fallback_tool_parsing: false` (and `mcp_manager: None`) — `crates/aura/src/orchestration/orchestrator.rs:2072`
- Single-agent path honors it — `crates/aura-web-server/src/handlers.rs:346` → `is_fallback_tool_parsing_enabled()`

## Impact

A local model that emits tool calls as text (rather than native tool-call structures) can fail or loop instead of answering when routed through the orchestrator, because neither the coordinator's routing nor the workers' tool invocations get the fallback text parser. Documented as a limitation for now (see below), but it means "orchestrator-by-default" excludes fallback-dependent local models from the default experience.

## Why it's non-trivial

- **Workers**: likely tractable — propagate the effective worker LLM's `fallback_tool_parsing` into the `Agent` and populate `fallback_tool_names` from the worker's discovered/filtered MCP tools (mirrors single-agent `builder.rs` logic).
- **Coordinator**: harder — it has `mcp_manager: None` and routes via plan-JSON parsing (`max_plan_parse_retries`), not MCP tool calls, so `fallback_tool_names` has nothing to bind to. Needs investigation into how routing / submit-result decisions are surfaced for local models that can't emit native tool calls.

## Proposed work

1. Propagate `fallback_tool_parsing` (+ `fallback_tool_names`) to worker agents from each worker's effective LLM config.
2. Investigate coordinator routing under fallback parsing (plan/route/submit-result) for models without native tool-calling; decide on a mechanism or document a hard boundary.
3. Add orchestration + Ollama integration coverage.

## Acceptance criteria

- With `[agent.llm].fallback_tool_parsing = true` and an Ollama model that emits text tool calls, an orchestrated request completes (workers invoke their tools; coordinator routes) without manual fallback to single-agent mode.

## Current workaround (documented)

[...]

PR #194, opened 2026-06-04
feat(quickstart): default to cli-driven orchestrator mode

### LINKED ISSUE #242: Quickstart
coordinate, don't duplicate

charlesjohnson:
- The quick start now leads with the bundled CLI (`docker exec -it aura ./aura-cli`)
- the default `quickstart.toml` ships in orchestration mode with a coordinator plus `researcher`/`writer` workers. The writer stays tool-free via a non-matching `mcp_filter` (an omitted/empty filter would grant all tools).
- `docker-compose.yml` enables `AURA_CUSTOM_EVENTS` so the CLI renders plan/worker events
- bumps LibreChat v0.7.8 → v0.8.6, which skips unknown SSE event types instead of crashing on them (danny-avila/LibreChat#12919); 
- `quickstart/librechat.yaml` points the default model at "Aura Orchestrator". 
- The docs (`docs/quickstart.md`, `docs/ollama-guide.md`) are aligned to the CLI-first orchestrator flow and now state that vector stores must be attached to a worker or coordinator to be queryable, and that `fallback_tool_parsing` applies only in single-agent mode. 
- The orchestration `fallback_tool_parsing` gap is tracked in #193.

### DIFF
--- a/docs/ollama-guide.md
+++ b/docs/ollama-guide.md
@@ -36,6 +36,8 @@ Known handled styles include:
 
 This improves tool reliability with local models that do not consistently emit structured function-calling payloads.
 
+> **Orchestration limitation.** `fallback_tool_parsing` applies only to single-agent mode (`[orchestration].enabled = false`). In orchestration mode the coordinator and workers are built with fallback parsing **disabled**, so a local model that can't emit native tool calls may fail to drive worker tools. Run such models in single-agent mode, or use a model with reliable native tool-calling for orchestration.
+
 ## Practical Guidance
 
 - Prefer instruction-tuned variants (`*-instruct`) when you need reliable tool execution.
--- a/docs/ollama-guide.md
+++ b/docs/ollama-guide.md
@@ -36,7 +36,7 @@ Known handled styles include:
 
 This improves tool reliability with local models that do not consistently emit structured function-calling payloads.
 
-> **Orchestration limitation.** `fallback_tool_parsing` applies only to single-agent mode (`[orchestration].enabled = false`). In orchestration mode the coordinator and workers are built with fallback parsing **disabled**, so a local model that can't emit native tool calls may fail to drive worker tools. Run such models in single-agent mode, or use a model with reliable native tool-calling for orchestration.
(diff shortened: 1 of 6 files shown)


docs/ollama-guide.md
```diff
@@ -36,6 +36,8 @@ Known handled styles include:
 
 This improves tool reliability with local models that do not consistently emit structured function-calling payloads.
 
+> **Orchestration limitation.** `fallback_tool_parsing` applies only to single-agent mode (`[orchestration].enabled = false`). In orchestration mode the coordinator and workers are built with fallback parsing **disabled**, so a local model that can't emit native tool calls may fail to drive worker tools. Run such models in single-agent mode, or use a model with reliable native tool-calling for orchestration.
```
teriyakichild: where did this come from?   I don't think this is accurate

charlesjohnson: Admittedly this is a generated PR, but as I understand it from 🤖 :

> This is from how orchestration builds its agents. `fallback_tool_parsing` is only honored on the single-agent path:
> 
> Single-agent enables it from `config: handlers.rs:346 → with_fallback_tool_parsing`.
> Coordinator hardcodes it off: `orchestrator.rs:2072` (`fallback_tool_parsing: false, mcp_manager: None`).
> Workers too: `orchestrator.rs:749` (`false, empty fallback_tool_names`); the Ollama worker arm sets up no fallback executor (`orchestrator.rs:2401`).
> 
> The wrapper no-ops unless the flag and tool names are set: `builder.rs:1099`.
> The coordinator routes by calling a routing tool; if it doesn’t, planning retries and then errors (“Coordinator did not call a routing tool”, `orchestrator.rs:1343-1427`), with no text-JSON fallback. 
> 
> A model emitting tool calls as text won’t register a routing decision, and the coordinator has fallback parsing disabled, so those aren’t recovered. Workers likewise rely on native tool calls (incl. `submit_result`). Net: a fallback-dependent local model works single-agent but stalls in the orchestrated default. 

> Filed #193 to close the gap in code.

charlesjohnson: Confirmed it is a real bug, not intended behavior. Orchestration builds the coordinator and workers with fallback parsing off (`orchestrator.rs:749` and `:2072`; the wrapper no-ops at `builder.rs:1099`), and the coordinator routes via a native routing-tool call (`orchestrator.rs:1343-1427`) — so a model that emits tool calls as text stalls instead of routing. Reframed the docs from "limitation" to a known issue tracked in #193, keeping the single-agent-mode workaround until the fix lands (d81576d).

teriyakichild (approved):

Issue #154, opened 2026-05-18
feat(mcp): surface degraded MCP server status instead of silently reporting 0 tools


ericlake:
## Summary

When an MCP server connection fails (auth errors, connectivity issues, etc.), Aura currently swallows the failure and reports `Connected successfully, 0 tools discovered`. This makes a degraded MCP server indistinguishable from one that simply has no tools, leaving users unaware of the problem.

## Current Behavior

When an MCP server fails to connect (e.g., 401 auth error), the logs show the error but then contradict it:

```
WARN  aura::mcp: HTTP Streamable MCP connection failed: MCP initialization error: ...
ERROR rmcp::transport::worker: worker quit with fatal: Transport channel closed, when Client(reqwest::Error { kind: Status(401, None), ... })
INFO  aura::mcp: pagerduty - Connected successfully, 0 tools discovered
```

The server appears fully functional with no tools to the end user.

## Expected Behavior

- The `Connected successfully` log should not be emitted when the connection actually failed — this is a bug
- Aura should track and surface degraded MCP server status to the user (e.g., via an `aura.mcp_status` SSE event or similar mechanism)
- The status should include the failure reason when possible (auth failure, connection refused, timeout, etc.)
- Users should be able to distinguish between "server has no tools" vs "server is configured but unavailable"

Baijack-star:
I traced this through the current MCP init path. The contradictory log comes from a real error being converted into a zero-tool success before `server_info` is written.

Relevant path:

- `crates/aura/src/mcp.rs` `initialize_from_config()` logs `"{server_name} - Connected successfully, {tools_count} tools discovered"` for any `Ok(tools_count)` returned by `connect_and_discover_tools()`.
- `connect_http_streamable()` calls `try_connect_http_streamable()`. For 401-ish errors it returns `Err(McpInitError(...))`, but for other failures it does:
  - `warn!("  HTTP Streamable MCP connection failed: {}", e);`
  - `Ok(0)`
- That `Ok(0)` is indistinguishable from a valid MCP server that connected and legitimately returned an empty `tools/list` response. So `initialize_from_config()` records `ConnectionStatus::Connected`, `tools_count: 0`, and emits the misleading success line.

[...]

PR #196, opened 2026-06-05
feat(mcp): surface and report failed MCP server connections

justintime4tea:
MCP connection failures were swallowed as Ok(0) and logged as "Connected successfully, 0 tools discovered", hiding degraded servers.

- mcp: bubble transport/discovery errors instead of Ok(0); record per-server status (Connected/Failed/NotAttempted) with reason
- capture the real HTTP status (404/401) in CustomHttpClient so the reason is precise, not a generic "channel closed"
- render the full anyhow cause chain ({e:#}) for richer reasons
- events: add aura.mcp_status SSE event projecting server status; emit at stream start (single-agent) and after the shared manager connects (orchestration, via StreamItem::McpStatus)
- cli: show failed/empty servers in a persistent status section and immediately in scrollback (themed ⏺ Error/Warning); condense verbose HTTP chains to "HTTP MCP server 'x': 404 Not Found"; warn on stderr in one-shot mode
- tests + streaming-api-guide docs

Ref: GH-154

### DIFF
--- a/crates/aura-cli/src/api/mcp_status.rs
+++ b/crates/aura-cli/src/api/mcp_status.rs
@@ -1,0 +1,302 @@
+//! Shared rendering of the `aura.mcp_status` SSE event into user-facing
+//! notices.
+//!
+//! Both the REPL (status area, styled) and one-shot mode (stderr) surface MCP
+//! connection problems. This module centralises the parsing and message
+//! wording so the two stay consistent — callers only decide presentation
+//! (color/prefix vs. plain stderr line).
+
+/// Marker the `aura` transport layer prepends to an HTTP status in the failure
+/// reason (e.g. `"server returned HTTP 404 Not Found"`). Shared from
+/// [`aura_events`] so producer and consumer match on a single source of truth —
+/// because it's **our** string and not a Rig/rmcp error format, it's stable
+/// across Rig fork bumps. The only upstream-owned fragment we parse is reqwest's
+/// `HTTP status ... (<status>)` `Display` output, used as a fallback in
+/// [`extract_http_status`].
+use aura_events::HTTP_STATUS_MARKER;
+
+/// A single user-facing notice derived from an `aura.mcp_status` event.
+///
+/// The variant carries the severity; the payload is the human-readable message
+/// body, which does **not** include an `error:`/`warning:` prefix — each caller
+/// adds its own prefix and styling.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum McpNotice {
+    Error(String),
+    Warning(String),
+}
+
+impl McpNotice {
+    /// The message body, regardless of severity.
+    pub fn message(&self) -> &str {
+        match self {
+            McpNotice::Error(message) | McpNotice::Warning(message) => message,
+        }
+    }
+}
+
+/// Human-friendly label for an MCP transport identifier from the event.
+fn transport_display(transport: &str) -> &str {
+    match transport {
+        "http_streamable" => "HTTP",
+        "sse" => "SSE",
+        "stdio" => "STDIO",
+        other => other,
+    }
+}
+
+/// Extract a concise HTTP status (e.g. "404 Not Found") from a verbose error
+/// chain, if one is present.
+///
+/// The server collapses the transport error chain into the reason string, which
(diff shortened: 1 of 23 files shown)


crates/aura-cli/src/api/mcp_status.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+/// A single user-facing notice derived from an `aura.mcp_status` event.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct McpNotice {
+    pub level: McpNoticeLevel,
+    /// Human-readable message body. Does **not** include an `error:`/`warning:`
+    /// prefix — each caller adds its own prefix and styling.
+    pub message: String,
+}
```
Shearerbeard: This could probably be condensed into a single parametrized enum

crates/aura-cli/src/api/mcp_status.rs
```diff
@@ -0,0 +1,281 @@
...
+
+/// Human-friendly label for an MCP transport identifier from the event.
+fn transport_display(transport: &str) -> &str {
+    match transport {
+        "http_streamable" => "HTTP",
+        "sse" => "SSE",
+        "stdio" => "STDIO",
+        other => other,
+    }
+}
```
Shearerbeard, line 46: Not a blocker. I assume we get transport as a string from either rig or from our older code. Might be worth adding it to the liat of magic strings that need to be hard typed.

crates/aura-cli/src/api/mcp_status.rs
```diff
@@ -0,0 +1,281 @@
...
+            && open < close
+        {
+            let status = rest[open + 1..close].trim();
+            if !status.is_empty() {
+                return Some(status.to_string());
+            }
+        }
+    }
+    None
+}
```
Shearerbeard, line 82: I also imagine this is another spot we get a string only from Rig.

crates/aura-cli/src/api/mcp_status.rs
```diff
@@ -0,0 +1,281 @@
...
+                level: McpNoticeLevel::Warning,
+                message: format!(
+                    "{disp} MCP server '{name}' connected but reported no tools available"
+                ),
+            });
+        }
+    }
+    notices
+}
+
```
Shearerbeard: You can probably just use .fold with a vec accumulator and remove the outer mut. Those unwrap_or can probably just become unwrap_or_defaut

crates/aura-cli/src/api/mcp_status.rs
```diff
@@ -0,0 +1,281 @@
...
+/// The server collapses the transport error chain into the reason string, which
+/// for an HTTP failure looks like:
+///   "... server returned HTTP 404 Not Found: Failed to establish ...
+///    Client error: HTTP status client error (404 Not Found) for url (...)"
+///
+/// We prefer the `server returned HTTP <status>` fragment (injected by AURA at
+/// the transport layer, so it's clean), then fall back to reqwest's
+/// `HTTP status ... (<status>)` form.
+fn extract_http_status(reason: &str) -> Option<String> {
+    const MARKER: &str = "server returned HTTP ";
```
Shearerbeard: Maybe we can move the const assignment out of the function and if it's matching up to library string matcing (type erased errors from Rig) we should probably make a note of which commit of our fork we are matching against

Shearerbeard (changes requested):
Minor things

crates/aura-cli/src/api/mcp_status.rs
justintime4tea: I think right now we don't have any shared transport enum or type and you're right these are magic strings. They are our magic strings though I think - http_streamable/sse/stdio are string literals across aura,
  aura-events, aura-config, and the transport field rides the SSE wire as a String.

crates/aura-cli/src/api/mcp_status.rs
justintime4tea: This string is ours - I moved the const to aura-events so it can be shared between aura and aura-cli.

Shearerbeard (dismissed):

Shearerbeard (dismissed):

Shearerbeard (approved):

PR #200, opened 2026-06-09
feat(streaming): surface upstream provider error detail

charlesjohnson:
Upstream model-provider errors were collapsed into a fixed generic message ("the request could not be completed") that hid the real cause and made failures impossible to diagnose, even though the actual error string was already in hand at the stream error branch. This change parses that error string for the HTTP status, provider error `type`, and `message` and surfaces a readable line instead (e.g. `429 requests: Rate limit reached for gpt-4o`). When nothing parses, it falls back to the raw error so detail is never silently swallowed, and it caps the surfaced detail at 500 chars so a large provider body can't flood the response stream. Context-overflow handling and the full raw error in logs/OTel are unchanged. Covered by new unit tests for the OpenAI/OpenRouter and Anthropic error shapes, the raw fallback, the length cap, and the end-to-end stream-collection wiring.


crates/aura-web-server/src/streaming/handlers.rs
```diff
@@ -1228,6 +1227,42 @@ fn is_context_overflow_error(error_str: &str) -> bool {
...
+/// Build the user-facing message for an upstream provider error.
+///
+/// `error_str` is the stringified `StreamError` coming off the provider stream
+/// (see `provider_agent::map_stream_item`). Its format is owned by the upstream
+/// provider/rig and not something we control, so we do not parse it.
+///
+/// When `debug` is true (dev-only `AURA_DEBUG_PROVIDER_ERRORS` flag) the raw
+/// error is surfaced verbatim — length-capped — so operators can diagnose the
+/// failure. When false (the default, required for public-facing deployments)
+/// only a generic message is returned; the raw error still reaches logs/OTel.
```
Shearerbeard: These are pretty claude typical to over document here - type system is strong and comments shouldn't restate the type system  and only explain necessary non obvious behavior.

crates/aura-cli/src/backend/direct.rs
```diff
@@ -70,6 +70,9 @@ impl DirectBackend {
             streaming_buffer_size: 400,
             aura_custom_events: true,
             aura_emit_reasoning: true,
+            // Standalone CLI runs locally for a single user (not public-facing),
+            // so surface raw upstream provider errors to aid local diagnosis.
```
Shearerbeard: To many comments also covered elsewhere

Shearerbeard (changes requested):
Mostly just verbose comments.

Shearerbeard (dismissed):

justintime4tea (dismissed):

justintime4tea (approved):

Issue #174, opened 2026-05-27
Refactor to make aura-config crate the canonical place for config structs


justintime4tea:
## Summary
Right now there are several config structs which are duplicated between crates - WorkerConfig is one such example. This duplications means that we have to keep multiple “semantically same” structs in sync, have funky boilerplate code “converting” from one to the other, and the relationship between aura and aura-config is reversed - aura-config depends on aura instead of the other way around. 

We should refactor and move all config structs into aura-config. As a product of this, we will remove the unnecessary duplication. When refactor is complete, aura crate should depend on aura-config crate - aura-config crate should _not_ depend on aura crate.

PR #201, opened 2026-06-10
refactor(config): move pure config types into aura-config

justintime4tea:
Invert the aura/aura-config dependency so all serializable TOML config types are owned by aura-config and re-exported from aura.

- Move LlmConfig, AgentSettings, orchestration/worker config, and scratchpad config out of the aura crate into aura-config, split into new orchestration.rs and scratchpad.rs modules.
- Relocate lenient_int into aura-config, where the moved types use it.
- Move RigBuilder the other way: from aura-config into a new aura::rig_builder module, since it composes runtime agents.
- Rename the old aura AgentConfig to AgentRuntimeConfig, the runtime build context pairing pure config with non-serializable fields (tool wrappers, persistence handles, session id); the TOML [agent] table is now the distinct aura_config::AgentConfig.
- Update all call sites and tests to the new module paths.

Ref: GH-174


Shearerbeard (changes requested):
Looks good to me other than there are a few out of date node paths in our docs - I think promptless flagged them but I haven't vetted every recommendation yet

Shearerbeard (approved):

PR #222, opened 2026-06-11
fix(ci): rewrite install script to avoid json parsing

teriyakichild:
## Summary
- **Fixed unbound variable bug**: `local tmpdir` in `main()` was invisible to the EXIT trap after the function returned, causing `set -u` to error
- **Replaced JSON parsing with plain-text checksums**: the grep regex couldn't match across nested JSON objects (the `uploader` field contains `}`), silently failing checksum verification. Now downloads a `checksums.txt` release asset instead — no jq, no API calls, no JSON
- **Version detection via HTTP redirect**: uses the Starship/k3s pattern (`curl -I .../releases/latest` → parse Location header) instead of GitHub API JSON
- **Generates `checksums.txt` in CI**: `sha256sum` in the `build-release-binaries` make target, uploaded alongside binaries by semantic-release

## Test plan
- [x] `bash scripts/install.sh` installs both binaries successfully
- [x] Gracefully skips verification when `checksums.txt` is missing (current releases)
- [x] `AURA_CHECKSUMS=<local file>` with valid checksums → "Verified checksum: OK"
- [x] `AURA_CHECKSUMS=<local file>` with bad checksums → "checksum mismatch", exits non-zero
- [x] `AURA_VERSION=1.22.0` pinned version works
- [x] Latest version auto-detection works

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -15,19 +15,24 @@ REPO="mezmo/aura"
 VERSION="${AURA_VERSION:-latest}"
 INSTALL_DIR="${AURA_INSTALL:-${HOME}/.local/bin}"
 COMPONENT="${AURA_COMPONENT:-all}"
-GITHUB_API="https://api.github.com"
+BASE_URL="https://github.com/${REPO}/releases"
 
 main() {
     detect_platform
     resolve_version
-    fetch_release_metadata
 
     echo "Installing AURA ${VERSION} (${OS}/${ARCH}) to ${INSTALL_DIR}"
     mkdir -p "${INSTALL_DIR}"
 
-    local tmpdir
     tmpdir=$(mktemp -d)
     trap 'rm -rf "${tmpdir}"' EXIT
+
+    if [[ -n "${AURA_CHECKSUMS:-}" ]]; then
+        cp "${AURA_CHECKSUMS}" "${tmpdir}/checksums.txt"
+    else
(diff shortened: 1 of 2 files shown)


Shearerbeard (approved):

scripts/install.sh
```diff
@@ -69,38 +74,48 @@ detect_platform() {
 
 resolve_version() {
     if [[ "${VERSION}" == "latest" ]]; then
-        VERSION=$(curl -fsSL -H "Accept: application/vnd.github.v3+json" \
-            "${GITHUB_API}/repos/${REPO}/releases/latest" \
-            | grep -o '"tag_name" *: *"[^"]*"' | head -1 | sed 's/.*"v\([^"]*\)".*/\1/')
+        local location
+        location=$(curl -fsSI "${BASE_URL}/latest" 2>/dev/null \
```
jmoses, line 78: The workers should have jq installed (although it might need to be added to the OSS workers)

jmoses: unless this wasn't the json regex parsing that was removed :D

teriyakichild: This runs on the user's workstation, so I'm trying to minimize dependencies.  I did consider adding jq, but found that it was not a pattern used by popular open-source repos.

jmoses: I missed it was a user-facing script

jmoses: time to make jq-as-a-service

PR #247, opened 2026-06-12
fix(mcp): bound tool error content before it reaches an agent

justintime4tea:
MCP tools can return multi-KB stack traces or provider error blobs in their is_error responses. Unlike successful outputs, errors are never diverted to the scratchpad, so an oversized error flowed straight into a worker's context window — triggering rapid replan loops that timed out.

Earlier work bounded these only at the coordinator's continuation prompt; a worker still saw the raw error inline. Bound errors at the tool-result source instead, so every consumer (worker, single-agent, coordinator) is protected before the error reaches an LLM context.

- add bound_error_content: head+tail truncation preserving the top-level error and the deepest cause, with an omitted-bytes notice; the byte budget is a target, not a strict cap
- apply it in extract_tool_result (text + structured), in from_jsonrpc_error, and on the HTTP transport-error path
- extract bound_transport_error so the changed path is unit-tested
- successful outputs are untouched (handled by the scratchpad)

Ref: GH-18

### DIFF
--- a/crates/aura/src/mcp_response.rs
+++ b/crates/aura/src/mcp_response.rs
@@ -84,16 +84,46 @@ impl CallOutcome {
     pub fn from_jsonrpc_error(code: i32, message: &str) -> Self {
         if SCHEMA_ERROR_CODES.contains(&code) {
             Self::SchemaError {
-                content: message.to_string(),
+                content: bound_error_content(message.to_string(), MAX_TOOL_ERROR_BYTES),
                 code,
             }
         } else {
             Self::GeneralToolError {
-                content: message.to_string(),
+                content: bound_error_content(message.to_string(), MAX_TOOL_ERROR_BYTES),
                 code: Some(code),
             }
         }
     }
+}
+
+/// Target budget (in bytes) for an MCP tool *error* surfaced to an agent.
+pub const MAX_TOOL_ERROR_BYTES: usize = 8 * 1024;
+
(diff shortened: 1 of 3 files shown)


crates/aura/src/mcp_response.rs
```diff
@@ -84,18 +84,68 @@ impl CallOutcome {
...
+/// Bound an MCP error message to roughly `max_bytes`, preserving head and tail.
+///
+/// Stack traces carry the high-level error at one end and the root cause at
+/// the other, so head-only truncation can discard the single most actionable
+/// line. This keeps ~75% from the head (top-level error and any
+/// failure-categorization keywords) and ~25% from the tail (deepest cause),
+/// joined by a notice stating how many bytes were omitted. Returns the input
+/// unchanged when it already fits.
+///
+/// `max_bytes` is a target, not a strict cap — the rendered result may exceed
```
Shearerbeard: comments seem like they explain the whole thing - we can cut them back a bit to the "non obvious" bits

Shearerbeard (changes requested):
One one minor thing

crates/aura/src/mcp_response.rs
justintime4tea: I wasn't sure where to draw the line on verbosity - I think I hit a better balance, give it a check?

Shearerbeard (approved):

PR #249, opened 2026-06-13
Fix transitive dependency context and failure propagation in orchestration

### LINKED ISSUE #221: sequential plans lose transitive dependency context
## Problem

Tasks in a sequential plan only receive the previous step's result. `flatten_sequential` (crates/aura/src/orchestration/types.rs) sets each task's dependencies to the step immediately before it, and `build_task_context` (crates/aura/src/orchestration/orchestrator.rs) only walks direct dependencies. Failure propagation breaks the same way: `blocked_tasks` only blocks direct dependents when a task fails, so transitive dependents stay pending and the executor's deadlock breaker silently drops them. The test suite documents the limitation as-is ("Only direct dependents are blocked").

Real plan from the SRE hard benchmark ([answerbook/ai-experiments](https://github.com/answerbook/ai-experiments) `aura-e2e`, branch `mshearer/aura-e2e-tests`), restart-investigation prompt on gpt-5.2:

```
plan:
  T0  query prometheus for pods with restart_count > 5
  T1  fetch alertmanager rules
  T2  for each pod found (T0), determine which rule (T1) would fire

dependencies the plan needs:    T2 <- {T0, T1}
dependencies the runtime built: T2 <- {T1}      (frontier replaced per step)

execution:
  T0  ok    15s   4 tool calls
  T1  ok    11s   2 tool calls
  T2  FAIL  300s  0 tool calls   prompt has T1's rules but not T0's pod list,
                                 worker has nothing to act on, sits to timeout
```

Single agent answers the same question in under 80s.

The failure-propagation half, same benchmark (multi-category-findings, gpt-5.2):

```
  T0  ok
  T1  FAIL  timeout
  T2  blocked            (direct dependent of T1)
  T3  pending forever    never started, silently dropped from the final answer
```

## Evidence

Scanning all stored SRE hard runs with `aura-stalls` (83 result directories, 9+ models): 92 affected task instances. 3 hard stalls (zero tool calls then timeout), 85 planned tasks dropped after an upstream failure, and 4 tasks re-fetching data an earlier task already had with identical arguments. Another 56 tasks succeeded but ran 5x+ slower than their plan siblings while re-deriving the missing context. That's 12% of all worker task time corpus-wide and 15% in canonical runs. In the run above, the starved task is 92% of the prompt's task time.

Classifying the affected plans: 86 of 90 non-infra cases were valid decompositions, mostly forced by tool contracts (you can't `k8s_get_workload` before listing workloads) or synthesis steps that combine several earlier results. The plans ask for transitive context, the runtime doesn't deliver it.

## Notes

- SSE events don't include task dependency lists, so none of this is visible from streamed data. Affected runs just look like slow models.
- Only observed in orchestration mode with sequential plans of 3+ steps. Parallel fan-out is unaffected, the synthesis step directly depends on every branch.
- A parallel plan dodges the simple case (T0/T1 above are independent, models that fan out rarely hit this). Chained steps like list -> get -> correlate have no plan shape that delivers both earlier results though, so we can't just prompt our way out.
- Scan is reproducible: `uv run aura-stalls sre-hard-results-*` in ai-experiments `aura-e2e` on `mshearer/aura-e2e-tests`.

Shearerbeard:
### Summary
Previously, when tasks in a plan depended on each other (e.g., T0 → T1 → T2), a task only received the output of its immediate predecessor. T2 would see T1's results but miss T0's context entirely. This caused two major issues:
1. Workers missing earlier context got stuck in loops, causing 300s timeouts and wasting about 12% of total task time.
2. If an early task failed, downstream tasks were left in limbo and silently dropped instead of being properly flagged as blocked.

This PR fixes these issues by:
- Passing all previous ancestor results (e.g., both T0 and T1) to the current task.
- Adding a 32KB limit for this dependency context to prevent prompt bloat. If the limit is reached, older tasks fall back to a 500-byte summarized preview.
- Propagating failures correctly. If an upstream task fails, the replanner now sees the entire chain of blocked tasks.
- Adding a `dag` field to `plan_created` events and a `dependencies` field to `task_started` events so API consumers can see the full execution graph.
- Consolidating duplicate orchestration events into a single source (`aura-events`) to fix data drift issues.

### Validation
- The timeout stall is completely gone. Executions are 9x faster on our canonical prompt, and wasted time dropped from 46.9% to 2.5% in testing.
- The fix causes zero regressions across multiple models (GPT-5.1/5.5, Sonnet/Opus 4.6, Qwen 3.6).
- Healthy runs show no extra latency or token usage.
- Live tests confirmed the replanner successfully detects and re-executes blocked chains.
- The new 32KB context limit coexists with token budgets without throwing `ContextOverflow` errors.

### Design Notes
- The new `dag` array maps 1:1 with the `tasks` array by index. 
- New SSE payload fields use `#[serde(default)]` to ensure backward compatibility with older clients.
- The `dependency_context_budget` is configurable in `[orchestration.artifacts]` (defaults to 32,000 bytes).

Fixes #221


crates/aura-config/src/orchestration.rs
```diff
@@ -173,6 +173,14 @@ pub struct ArtifactsConfig {
     #[serde(default = "default_result_summary_length")]
     pub result_summary_length: usize,
 
+    /// Total byte budget for dependency context injected into a worker prompt.
+    /// Direct dependencies always render in full; transitive ancestors render
+    /// in full until this budget is reached, then degrade to compact previews.
+    /// Default: 32000 (8x result_artifact_threshold, sized to the same
+    /// continuation-prompt envelope that fits all 200K-window models).
+    #[serde(default = "default_dependency_context_budget")]
+    pub dependency_context_budget: usize,
```
jakedipity: Defining this budget in terms of bytes feels a little inconsistent as I'd expect this to be in terms of tokens given it is impacting the prompt size and context usage. At a minimum I think adding the suffix `_bytes` would make it clear, although I'd personally lean towards token counting.

Shearerbeard: Thats a good catch - most of the rest of what we do its now caught with token counting even though our implementation isn't quite 100% accurate. We mostly use tiktoken for openai style counting which I think undercounts for Claude.

jakedipity (dismissed):
Besides my 1 comment which isn't a blocker, looks good

crates/aura/src/orchestration/frame_validation_tests.rs
```diff
@@ -1127,7 +1128,10 @@ fn test_worker_task_empty_context() {
 // ========================================================================
 
 #[test]
-fn test_continuation_running_task_renders_as_blocked() {
+fn test_continuation_running_task_renders_as_not_executed() {
+    // A pending/running task with no failed ancestor is NOT blocked — it
+    // just never ran. It must not be mislabeled "blocked (dependency
+    // failed)" (the pre-#221 behavior this test used to document).
```
justintime4tea: I notice these history mentioning comments slip in from time to time:
"(the pre-#221 behavior this test used to document)"

might be worth trimming out? I realize it references #221 but not sure the value post-human review?

Shearerbeard: Good catch

Shearerbeard: fixed

crates/aura-config/src/orchestration.rs
Shearerbeard: Pushed fixes so I could use another review

crates/aura/src/orchestration/orchestrator.rs
```diff
@@ -2695,56 +2706,38 @@ Assign tasks to the worker whose tools best match the required operations."#,
...
+    fn token_counter_for_task(
+        &self,
+        worker_name: Option<&str>,
+    ) -> Arc<dyn scratchpad::TokenCounter> {
+        let effective_llm = worker_name
+            .and_then(|name| self.config.workers.get(name))
+            .and_then(|w| w.llm.as_ref())
+            .unwrap_or(&self.agent_config.llm);
+        let (provider, model) = effective_llm.model_info();
+        scratchpad::token_counter_for_provider(provider, model)
```
jakedipity, line 2722: nit: feels like this should be moved to some shared crate specifically for tokens, maybe we do that when we do #257

jakedipity (dismissed):

examples/reference.toml
```diff
@@ -187,6 +187,29 @@ model = "gpt-5.2"
...
+enabled = false
+max_planning_cycles = 3
+allow_direct_answers = true
+allow_clarification = true
+tools_in_planning = "summary"
+max_tools_per_worker = 10
+
+[orchestration.artifacts]
+# Base directory for execution persistence, plan storage, and result artifacts.
+memory_dir = "/tmp/aura-orchestration"
```
justintime4tea: Should be a top-level configuration now - used by several things.

crates/aura/src/orchestration/stream_events.rs
justintime4tea, line 1: If we don't mind the bigger diff, we could move all of this into aura-events and do away with `stream_events.rs`? Looks like we just have some comments and tests left here and then in aura-events where this PR moves most of the other stuff, there aren't any tests at all.

justintime4tea (commented):
I added a couple small points of feedback.

The one other thing I could think of was that the budgeting doesn't seem to account for the pointer previews (125 tokens each), lineage headers, or separators - in a wide fan-out do we think this could pose a potential risk worth concerning ourselves with?

Also, I think these two comments might be conflicting? Thoughts?

- crates/aura-config/src/orchestration.rs:176:
`/// Total token budget for dependency context injected into a worker prompt.`

- crates/aura/src/orchestration/orchestrator.rs:4197:
`/// budget_tokens caps the total tokens of inlined full results.`

justintime4tea (approved):

Shearerbeard:
Closing - analyzing a different approach for the same fix and augmentation based on testing feedbac

PR #250, opened 2026-06-15
feat(cli): add init subcommand for guided starter-config generation

charlesjohnson:
Adds `aura-cli init`, an interactive command that generates a starter configuration so new users get to a working agent quickly.

## What it does
- Senses conventional API-key env vars; pick a provider from a validated numeric menu.
- Verifies the key against the provider's live model list, then shows a curated, best-first model shortlist (newest version per family, dated snapshots and non-chat variants hidden). Pick by number, accept the suggested default, or type any id.
- Writes a minimal `config.toml` whose provider/model/key are referenced from a sibling, gitignored `.env` via `{{ env.LLM_* }}` — no secret in the toml. An existing `.env` is merged non-destructively.
- Both binaries auto-load `.env` at startup so the generated config resolves without manual exporting.

## Foundation
- Adds the optional `[bootstrap]` config table (`Config.bootstrap`) so the generated config validates. The bootstrap agent itself lands separately.

## Tests
Provider/model selection, `.env` render/merge, and a dotenv→config-resolution integration test. Builds, clippy, and fmt clean across default and `standalone-cli` features.


crates/aura-cli/Cargo.toml
```diff
@@ -28,6 +28,7 @@ anyhow = { workspace = true }
 uuid = { workspace = true }
 regex = { workspace = true }
 clap = { workspace = true, features = ["env"] }
+dotenv = { workspace = true }
```
teriyakichild: It seems that dotenv is abandoned and we should be using dotenvy instead.

crates/aura-cli/src/init.rs
```diff
@@ -0,0 +1,1792 @@
...
+        if !self.interactive {
+            return Ok(detected_value.map(String::from));
+        }
+        match detected_var {
+            Some(v) if detected_value.is_some() => print!("API key [{v}]: "),
+            _ => print!("API key: "),
+        }
+        std::io::stdout().flush()?;
+        let mut line = String::new();
+        self.stdin.read_line(&mut line)?;
```
teriyakichild, line 542: `read_line` prints characters to the terminal. Standard practice for credential prompts is to suppress this. Possible solutions:
* crossterm is already a dependency
* rpassword::read_password() would work too

crates/aura-cli/Cargo.toml
charlesjohnson: Migrated the workspace from `dotenv` to the maintained `dotenvy` fork — manifests and every call site (including the pre-existing `aura-config` loader). 6e92b47

crates/aura-cli/src/init.rs
charlesjohnson: The API-key prompt now reads with echo suppressed via `rpassword` on a real terminal, so the key no longer appears on screen. An `is_tty` flag keeps the scripted unit tests reading their injected stdin. 6e92b47

crates/aura-cli/src/init.rs
```diff
@@ -0,0 +1,1822 @@
...
+         {llm}",
+        name = toml_escape(&spec.name),
+    )
+}
+
+/// Render a `.env` file for a user-provided API key. Only called when the
+/// user typed a key that isn't already in their environment.
+fn render_env(env_var: &str, value: &str) -> String {
+    format!(
+        "# Generated by `aura-cli init`. Gitignored — do not commit.\n\
```
Shearerbeard: I didn't catch this on my own but a claude once over points out that this statement is incorrect. Were writing an env file for them and we have no idea if its in their gitignore.

crates/aura-cli/src/init.rs
```diff
@@ -0,0 +1,1822 @@
...
+        None
+    } else {
+        match lister.list(&provider, verify_key, base_url.as_deref()) {
+            Ok(ModelList::Verified(models)) => {
+                println!(
+                    "Verified: {provider} answered with {} model(s).",
+                    models.len()
+                );
+                Some(models)
+            }
```
Shearerbeard, line 861: Claude catch that I wouldn't' have seen Openrouter may not actually be using auth at all on its models endpoint - could be thousands of models too

crates/aura-cli/src/init.rs
```diff
@@ -0,0 +1,1822 @@
...
+        "gemini" => &["gemini-2.5-pro", "gemini-2.5-flash", "gemini-2.0-flash"],
+        "openrouter" => &[
+            "openai/gpt-5",
+            "anthropic/claude-sonnet-4",
+            "google/gemini-2.5-pro",
+            "openai/gpt-4o",
+        ],
+        "ollama" => &["qwen3", "llama3", "mistral"],
+        _ => &[],
+    }
```
Shearerbeard: Pretty stale recommendations and far from best results or anything we would bother with in our harness

openai 5.1-5.5 and I wouldn't feel comfortable adding the o series for log analysis
claude 4.5+ on any of the 3 base models
gemini - 2.5 series barely runs tools well, stick to only 3.1 pro and the most recent 3.5ish flash model (verified working)

openrouter: same recommendations as above - add Arcee Trinity, Qwen 3.6 plus, and GLM 5.1 (well tested)

ollama - this we need to be more specific on - the strings aren't the right match and chances are you wouldn't see great results with local models older than qwen 3.6 (35b A4b), qwen 3.6 (27b), GLM 4.7 (30b), Gemma 4 27bA4b all work great but I partially wonder if we should even advertise ollama as a quick start - thats a rough path to go down unless you really know your shit with the open self hosted models

crates/aura-cli/src/init.rs
Shearerbeard, line 1: candidate for a quick module split

crates/aura-cli/src/init.rs
```diff
@@ -0,0 +1,1822 @@
...
+
+/// Default provider order (also the no-keys-found display order).
+const PROVIDERS: &[&str] = &[
+    "openai",
+    "anthropic",
+    "bedrock",
+    "gemini",
+    "ollama",
+    "openrouter",
+];
```
Shearerbeard: Make this a finite type (enum) and let the compiler/macros to the heavy lifting for the stringification/clap concerns and it will probably remove a bunch of code.

```
  #[derive(Debug, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
  #[clap(rename_all = "lowercase")]
  pub enum Provider {
      OpenAI, Anthropic, Bedrock, Gemini, Ollama, OpenRouter,
  }
```

^ claude snippet example of what this might look like. Removes a lot of down stream logic and checks

crates/aura-cli/src/init.rs
```diff
@@ -0,0 +1,1822 @@
...
+//! 5. **Model**: rank the fetched list into a short, best-first shortlist via
+//!    a per-provider table of stable family roots (newest version per family,
+//!    snapshots hidden). The user picks by number, accepts the suggested
+//!    default, or types any id — never a silent pick.
+//! 6. Write a minimal **complete** config referencing the provider's native
+//!    env vars.
+//!
+//! Verification is best-effort: network or key failures warn and continue
+//! (`--offline` skips the attempt entirely); init never hard-blocks on the
+//! network. Output is deterministic given the same choices.
```
Shearerbeard, line 26: Light read on my part but some caution: vet for accuracy and verbosity - these comments get hard to maintain as changes get made

crates/aura-cli/src/init.rs
Shearerbeard: Further notes here is that once you query ollama you don't necessarily get the canonical names of the models - you get whatever guff they installed (includes the quant - very messy name) or the user aliases. Its probably just safe to take whatever they have there as options but mention in the scripting what has been tested. TLDR ollama has some modes we know work (worth mentioning as a disclaimer in the flow) and that the list that comes back from /models is by nature user curated - not canonicalized in any way

Shearerbeard (changes requested):
Looks good other than some quick concerns and cleanup listed along with noticing we didn't have a readme update. Might be worth a pass with the rust-quality skill as you go if you haven't already to catch up any straggling nits I didn't get to in this review.

crates/aura-cli/src/init.rs
charlesjohnson: Fixed in ed5bdea — the generated .env header no longer claims to be gitignored; it now states the file holds a secret and to add it to .gitignore.

crates/aura-cli/src/init.rs
charlesjohnson: Right — OpenRouter's /models is unauthenticated (and can return thousands). Fixed in ed5bdea: `Verified` is now printed only for providers whose list endpoint authenticates the key (OpenAI/Anthropic/Gemini); OpenRouter and Ollama just report the count. The curated shortlist already keeps the displayed list short.

crates/aura-cli/src/init.rs
charlesjohnson: Vetted — fixed the one concrete inaccuracy (the .env "gitignored" claim) in ed5bdea, and trimmed the wording. Keeping comments lean; the planned init.rs module split will re-home/trim the rest.

crates/aura-cli/src/init.rs
charlesjohnson: To update the per-provider recommendations I need the exact id strings as each provider's models endpoint returns them — `family_roots` matches version-free *prefixes* against the start of those ids, then picks newest-by-natural-sort within a family. Specifically:

- **OpenAI**: which gpt-5.x ids to prefer (exact `gpt-5.1`…`gpt-5.5`), and confirm dropping the o-series.
- **Anthropic**: the 4.5 base-model ids (sonnet/opus/haiku) as `/v1/models` returns them.
- **Gemini**: the 3.x pro id + the current 3.5-flash id (the table is on 2.5 today).
- **OpenRouter**: the org/model ids for Arcee Trinity, Qwen 3.6, GLM 5.1 — and keep or replace the existing openai/anthropic/google entries?

Ollama is handled in ed5bdea per your note: it now shows whatever `/api/tags` returns (no family-root curation, and no chat-marker filter so `*-instruct` survives) plus a 'local results vary — see docs for tested models' disclaimer. If you drop the canonical tested ids here (qwen3.6 35b-a4b, GLM 4.7 30b, Gemma 4 27b, …) I'll put them in the disclaimer/docs.

crates/aura-cli/src/init.rs
charlesjohnson: Done in f16d79e — `Provider` is now a `clap::ValueEnum`, so `--provider` is validated at parse time (rejects unknown values, lists the choices in `--help`) and every per-provider match is exhaustive/compiler-checked. Dropped the PROVIDERS table, the manual validation bail, and its test.

crates/aura-cli/src/init.rs
charlesjohnson: Agreed — splitting in a follow-up. After the provider-enum change init.rs is ~1.9k lines, so the split is clearly worth doing, but as its own focused reorg PR: doing it here would balloon this PR's diff mid-review (and collide with the still-open model-id updates). Tracking it as the next PR.

crates/aura-cli/src/init.rs
Shearerbeard: This might work - my concern on this is that we haven't "verified" anything with the connection when we get models back - the key they provided isn't tested on that endpoign.

charlesjohnson: Agreed — and that's the case now in 39c84cf: for a key-using provider whose list endpoint doesn't authenticate (OpenRouter), the message says explicitly it "did not validate your key — that happens on first use," rather than implying anything was verified.

charlesjohnson:
Addressed this round (all pushed, branch green — clippy/rustfmt/tests):

- **README**: added an `aura-cli init` section to the CLI README (usage, what it does, generated `config.toml`/`.env`, full flag reference). 39c84cf
- **"Verified" wording**: for a key-using provider whose list endpoint doesn't authenticate (OpenRouter), the message now says explicitly it *did not validate your key — that happens on first use*. 39c84cf
- **Quality pass**: ran a reuse/simplification/efficiency/altitude sweep over `init.rs` and applied the behavior-preserving nits (dropped a dead no-op substitution, de-duplicated the 'model not in list' warning, simplified an unused binding). 5df1977

Still open / deferred:
- **Model recommendations**: waiting on the exact provider model-id strings (asked in the family_roots thread) before updating the table.
- **Module split** of `init.rs` (~1.9k lines): tracked as a focused follow-up PR, per the thread.

The altitude pass also noted per-provider behavior is spread across a few `match Provider` sites — I'm folding that consolidation into the module-split follow-up rather than churn this PR further.

crates/aura-cli/src/init.rs
charlesjohnson: Wired your recommendations into family_roots (bcdaad4):

- **OpenAI**: gpt-5.5 (default), gpt-5.4; o-series dropped.
- **Anthropic**: claude-sonnet-4-6 (default), claude-opus-4-8, claude-haiku-4-5.
- **Gemini**: gemini-3.5-flash (default), gemini-3.1-pro.
- **OpenRouter**: no recommendation; you type the id.
- **Ollama**: no recommendation; shows whatever /api/tags returns.

These are matched as id prefixes, so a dated snapshot (e.g. `gpt-5.5-2025-...`) still resolves to the clean id, and the first entry is the suggested default. One call I made for you: you didn't specify OpenAI's default, so I led with gpt-5.5 (newest); say the word if you'd rather gpt-5.4 lead.

Shearerbeard: 5.5 is good and IMO its hard to find a reason to recommend 5.4 over it in my testing.

charlesjohnson: Dropped gpt-5.4; OpenAI now recommends just gpt-5.5. (66bd21e)

Shearerbeard (approved):

PR #252, opened 2026-06-16
refactor: consolidate slash-command handling into one registry

### LINKED ISSUE #177: aura-cli: slash-command registration is brittle (4 sites, 3 silent failure modes)
## Summary

Adding a new REPL slash command currently requires touching four
different files, three of which fail **silently** if missed — tests
pass, the binary runs, but the command misbehaves in the live REPL in
ways that only surface during interactive use.

## The four sites

For a new `/foo` command, a contributor must update all of:

1. **`crates/aura-cli/src/repl/commands.rs`** — the `handle_foo`
   handler function.
2. **`crates/aura-cli/src/repl/loop.rs`** — the dispatch arm
   (`else if let Some(arg) = input.strip_prefix(\"/foo\") { ... }`).
3. **`crates/aura-cli/src/ui/state.rs::COMMANDS`** — the
   `(\"/foo\", \"description\")` tuple in the static slice.
4. **`crates/aura-cli/src/ui/event_replay.rs::print_help`** — the
   row printed by \`/help\`.

## The silent failure modes

| Missing site | Symptom | How it surfaces |
|---|---|---|
| `commands.rs` handler | Compile fails | Caught immediately — not silent. |
| `loop.rs` dispatch | \"Unknown command\" at runtime | Visible on first use. |
| `state.rs::COMMANDS` | Autocomplete hint disappears as you type; Enter silently does nothing because `validate_command_input` falls through to a final `COMMANDS.iter().any(...)` and rustyline treats `false` as \"input incomplete, do not submit.\" | **No tests catch this**; only interactive REPL use surfaces it. |
| `print_help` | Command is invisible in `/help` | Quiet discoverability gap; users assume the command doesn't exist. |

## Concrete recent example

In #_ (the PostHog telemetry PR), the new `/telemetry` command landed
with sites 1 and 2 done. Three follow-up commits were needed once the
pattern's brittleness surfaced:

- [`0186da4`](https://github.com/mezmo/aura/commit/0186da4) added the
  `/help` row.
- [`0a42018`](https://github.com/mezmo/aura/commit/0a42018) added the
  `COMMANDS` entry, which was what actually broke Enter in the live
  REPL.

Both were caught only when a maintainer tried `/telemetry` interactively
and reported that the autofill disappeared and Enter was silently
inert. No unit test, integration test, or piped REPL subprocess test
would have caught either: the dispatch fires fine when input arrives
via stdin pipe (which bypasses rustyline's validator).

## Suggested direction

Two tiers, ordered by scope:

**Tier 1 — single source of truth for the list** (smaller refactor,
catches the class of bug above):

- Make `state.rs::COMMANDS` the registry. Add fields beyond name +
  description if needed — e.g. an optional validator hook.
- Have `print_help` iterate `COMMANDS` instead of a hand-maintained
  list. (Most rows are mechanical; the four that take arguments
  could carry a `usage_hint` string.)
- Nudge `loop.rs` dispatch toward a `match` over a name resolved from
  the registry — at minimum, error out on a registry-entry-without-
  handler at compile time.

This is ~50 lines of refactor and the registry change alone would have
caught the `/telemetry` regression because adding a handler without
adding a `COMMANDS` row would have stayed compiling but failed a
trivial integration test (\"every registered command should validate
without prefix\").

**Tier 2 — uniform handler signatures via a `Context` struct** (bigger
refactor, defers as separate work):

Handlers today take heterogeneous parameters (`&mut conversation`,
`&ConvStore`, `&mut input_reader`, `&TelemetryHandle`, etc.). Bundling
into a `CommandContext` would let the dispatch be a single
`match` arm with a `fn(&Context, &str)` handler shape, fully closing
the registration gap. Worth doing later but not strictly necessary to
fix the silent failure modes.

## Why this matters

The pattern has been growing — `/telemetry` is the 12th slash command —
and each new addition has the same trap. The brittleness compounds:
every contributor learns about the four sites by stepping on the rake.

A short ADR (we discussed kicking off a `docs/adr/` convention; the
telemetry PR has a stashed `0001-record-architecture-decisions.md` for
that) capturing the chosen approach would also serve as the contributor
docs for adding the next slash command.

## Reproduction (post-PR for the regression)

This is fixed on `main` post-`0a42018`, but the structural smell remains.
Reproduce the bug pattern on any new command by:

1. Add `handle_foo` in `commands.rs`.
2. Add the dispatch arm in `loop.rs`.
3. **Do not** touch `state.rs` or `event_replay.rs`.
4. Build and run \`./target/debug/aura-cli\`. Type `/foo` interactively.
5. Observe: autofill disappears, Enter inert, no error message.

jakedipity:
Slash commands were defined across five hand-synced sites: the
dispatch chain, midstream execution, a command table, the
Enter-submission validator, and the /help text. These drifted
(e.g. /stream and /style were absent from /help) and the validator
and dispatcher matched command names by different rules.

Introduce repl::registry as the single source of truth. Each Command
holds its name, description, usage hint, handler, and an optional
submission gate; dispatch, validation, /help, and autocomplete all read
from one COMMANDS table, so adding a command is one entry plus a
handler.

- replace the if/else dispatch chain with a registry lookup returning a
  CommandOutcome (handled, reinject input, or exit)
- match commands by exact name in the dispatcher, validator, and
  mid-stream handler, dropping the duplicated prefix-resolution helper;
  abbreviations are still completed via Tab
- generate /help and the autocomplete hint from the registry
- add unit tests for lookup, submission gates, validation, and argument
  splitting
- drive midstream execution from the registry

User-visible changes: /help now lists every command (including /stream
and /style); unknown or partial commands now submit and report "unknown
command" instead of being silently ignored.

Fixes: https://github.com/mezmo/aura/issues/177

### DIFF
--- a/crates/aura-cli/src/repl/loop.rs
+++ b/crates/aura-cli/src/repl/loop.rs
@@ -21,11 +21,12 @@ use crate::repl::commands;
 use crate::repl::conversations::ConversationStore;
 use crate::repl::history::ConversationHistory;
 use crate::repl::input_reader::{self, HISTORY_COUNT, HISTORY_DEPTH, LAST_READLINE_INPUT};
+use crate::repl::registry::{self, CommandContext, CommandOutcome};
 use crate::tools;
 use crate::ui::markdown::render_markdown;
 use crate::ui::prompt::{
-    PendingCommand, WaveAnimation, cleanup_terminal, clear_display_events, clear_input_hint,
-    drain_stdin, erase_input_frame, extend_display_events, frame_lines, get_cumulative_tokens,
+    WaveAnimation, cleanup_terminal, clear_display_events, clear_input_hint, drain_stdin,
+    erase_input_frame, extend_display_events, frame_lines, get_cumulative_tokens,
     get_selected_model, handle_ctrlc, install_sigint_handler, is_expanded_output,
     last_mid_stream_history_entry, load_and_restore_sse_events, lock_term,
     overwrite_orch_task_header_unlocked, prepare_input_line, print_fields_tree,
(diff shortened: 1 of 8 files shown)


Shearerbeard (approved):
approved pending on a second opinion on testing

justintime4tea (approved):

teriyakichild (approved):
LGTM

Issue #191, opened 2026-06-03
HITL approval gates for MCP tool calls


Shearerbeard:
## Problem

Agents execute tool calls unconditionally. There is no way to require a human to sign off before a destructive call runs, which blocks SRE dogfooding of the orchestrated agent: incident remediation needs an operator's approval before anything irreversible happens.

Two deployment shapes need approval and differ structurally:

- Unattended (A2A, background): no human on the stream, so an external service answers.
- Attended (CLI, dev, dogfooding): the operator on the SSE stream is the approver. SSE is one-way, so the decision needs a path back up.

## Decision

Recorded in the ADR (`docs/adr/2026-06-16-hitl-approval-architecture.md`) with the type-level design in `docs/design/hitl.md`: dual-channel routing chosen by config, fail-closed, per-process registry.

Routes, fixed per deployment in a `[hitl]` table:

- Webhook (unattended): one synchronous round-trip; decision in the response body.
- Conversational (attended): the gated call parks in-process while the SSE stream stays open; the decision returns via `POST /v1/approvals/{decision_id}`. The park lasts as long as the connection, and a dropped stream fails closed.

Where a single-agent client advertises its own `request_approval` tool, the attended surface reuses the existing turn-ending client-tool cycle.

```toml
[hitl]
require_approval = ["kubectl_*", "restart_*"]

[hitl.route]
mode = "conversational"   # or: mode = "webhook", url = "https://..."
timeout_secs = 300
```

## Plan

Each phase has a workspace build/test/clippy gate and an e2e smoke against a local mock webhook:

1. Domain plus module split, webhook arm only.
2. Conversational route: registry, decision ingress, disconnect/shutdown cancellation.
3. Attended client plus orchestration verification.

Landed: the id newtypes (RunId/SessionId/TaskIdentity) and the `aura::hitl` typed skeleton (signatures complete, `todo!()` bodies).

## Out of scope (tracked elsewhere)

Durable park-and-resume and exactly-once binding (#209), ingress auth, A2A InputRequired wiring, confidence-based auto-approve, multi-webhook fan-out, an Auto route, coordinator-mediated surfaces.

## Notes

- Research context: LOG-23108
- Interacts with #187 (long-silent-call detection must treat a parked approval as alive)
- Builds on the config-crate inversion (#174 / PR #201)

rpelevin:
[...]

PR #256, opened 2026-06-16
doc(adr): hitl approval architecture decision and design note

### LINKED ISSUE #255: HITL Dual Channel ADR
Quick ADR and design doc (type based design) around AURA HITL implementation.

Shearerbeard:
Aura's agents run tool calls unconditionally, there is no way to require human
sign-off before a destructive call. That blocks SRE dogfooding of the
orchestrated agent (#191). This is the design decision for human-in-the-loop
approval gating written slightly before and modified during the first spike. 

- `docs/adr/2026-06-16-hitl-approval-architecture.md`: the decision, in MADR
  format. Two approval routes selected per deployment in config (a synchronous
  webhook for unattended A2A/background; a conversational route that holds the
  SSE stream open and takes the decision back through `POST /v1/approvals/{id}`
  for attended CLI and dogfooding), the fail-closed lifecycle, and a per-process
  registry with named single-pod and durability boundaries.
- `docs/design/hitl.md`: the companion design note. Domain types, module layout,
  config and wire formats, SSE event gating, and a keep/rework/discard map over
  the unmerged spike (`mshearer/hitl-v1-config-gate`).

Type-level spec lives in the design notes to avoid polluting the main ADR. Implementation phasing is tracked in #191.

Closes #255

### DIFF
--- a/docs/adr/2026-06-16-hitl-approval-architecture.md
+++ b/docs/adr/2026-06-16-hitl-approval-architecture.md
@@ -1,0 +1,186 @@
+<!-- markdownlint-disable MD033 -->
+# Human-in-the-loop approval gating for agent tool calls
+
+- Status: **proposed**
+- Deciders: Mike Shearer
+- Date: 2026-06-16
+
+Technical Story: [#191](https://github.com/mezmo/aura/issues/191)
+
+## Context and Problem Statement
+
+Aura's agents execute tool calls unconditionally. There is no way to require a
+human to sign off before a destructive call runs, which blocks SRE dogfooding of
+the orchestrated agent ([#191](https://github.com/mezmo/aura/issues/191)):
+incident remediation needs an operator's approval before anything irreversible
+happens.
+
+Two deployment shapes need approval, and they differ structurally:
+
+- **Unattended** (A2A, background): no human is on the stream, so an external
+  service answers approval requests.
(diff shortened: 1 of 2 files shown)


charlesjohnson:
Design comments:
1) Are we thinking about fallback on deny or timeout? Should there be retry loops? An opportunity to provide an alternate instruction rather than just deny? If we timeout, what to do next?
2) Based on reports of people who have built similar systems, requests quickly outstripped timeouts, and they wound up needing to queue. That's probably YAGNI but worth calling out for the future.

Shearerbeard:
> Design comments:
> 
> 1. Are we thinking about fallback on deny or timeout? Should there be retry loops? An opportunity to provide an alternate instruction rather than just deny? If we timeout, what to do next?
> 2. Based on reports of people who have built similar systems, requests quickly outstripped timeouts, and they wound up needing to queue. That's probably YAGNI but worth calling out for the future.

1. Deny has an optional reason string that gets fed back to the LLM similar to what you would expect in a coding chat agent. Timeout similarly fed back to the agent - the agent turn gets to clarify or decide to try something else
2. There is a mild nod to this here but its out of scope for this document and would have to come post session state implementation. Consequences line. The follow up proposal for V2 will be closely tied with session state and you'll have a chance to review this.

jakedipity (approved):

charlesjohnson (approved):

PR #261, opened 2026-06-18
fix(mcp): suppress stdio MCP server stderr output

teriyakichild:
rmcp's TokioChildProcess::new defaults stderr to Stdio::inherit(), which lets MCP server debug output (often raw JSON-RPC frames) leak to the host terminal. Use the builder with Stdio::null() instead.


Shearerbeard (approved):
Good find @teriyakichild

PR #269, opened 2026-06-23
feat(cli): default to standalone mode when --api-url is absent

### LINKED ISSUE #268: make standalone-cli the default feature for aura-cli
## Summary

The `standalone-cli` cargo feature on `aura-cli` is currently opt-in, which means release binaries and the Docker image are built as HTTP-only clients. Users must pass `--standalone --config <path>` every time they want to run agents in-process.

## Proposal

1. Make `standalone-cli` a default cargo feature so all standard builds include it
2. Default to standalone mode when `--api-url` / `AURA_API_URL` is not set
3. Default `--config` to `config.toml` in the current directory when omitted
4. Only require `--standalone` explicitly when the user wants standalone mode *and* `--api-url` is set

This simplifies the most common invocation from:
```
aura-cli --standalone --config config.toml
```
to just:
```
aura-cli
```

HTTP mode is still available via `--api-url`.

teriyakichild:
## Summary

- Make `standalone-cli` a default cargo feature so release binaries and the Docker image include standalone support without extra build flags
- Default to standalone mode when `--api-url` / `AURA_API_URL` is not set — simplifies the common invocation from `aura-cli --standalone --config config.toml` to just `aura-cli`
- `--config` defaults to `config.toml` in the current directory when omitted
- `--standalone` is only required when the user wants standalone mode *and* `--api-url` is set (e.g. for debugging)

## Changes

- **`Cargo.toml`**: `default = ["standalone-cli"]`
- **`cli.rs`**: New `resolve_standalone()` replaces `validate_standalone_args()` — decides mode based on whether `--api-url` is present. Non-feature builds error early when `--api-url` is absent
- **`backend/mod.rs`**: `from_config` takes `is_standalone: bool` instead of reading `args.standalone`; defaults config path to `config.toml`
- **`main.rs`**: Wired up to new `resolve_standalone()`
- **`tests/cli_args.rs`**: Updated tests for new behavior + new test for `--config` + `--api-url` warning

## Test plan

- [x] `cargo build -p aura-cli` (default features, includes standalone)
- [x] `cargo build -p aura-cli --no-default-features` (HTTP-only build)
- [x] `cargo test -p aura-cli` — 380 tests pass (366 unit + 14 integration)
- [x] `cargo clippy -p aura-cli` — clean
- [x] Verify `aura-cli` with no args enters standalone mode and tries `config.toml`
- [x] Verify `aura-cli --api-url http://... ` enters HTTP mode
- [ ] Verify Docker image build still works

Fixes #268

### DIFF
--- a/.makefiles/rust.mk
+++ b/.makefiles/rust.mk
@@ -54,6 +54,11 @@ nextest: $(DOCKER_ENV) $(NEXTEST_BIN) $(REPORT_DIR)
 lint-rust: | $(DOCKER_ENV) $(REPORT_DIR)  ## lint rust code via clippy
 	$(RUN) cargo clippy $(if $(IS_CI),-q,) --all-targets --all-features $(if $(IS_CI),--message-format=json,) -- -D warnings $(if $(IS_CI),> $(REPORT_DIR)/clippy.json,)
 
+.PHONY: check-cli-http-only
+check-cli-http-only: $(DOCKER_ENV) ## Verify the HTTP-only (no-default-features) aura-cli still builds
+	$(RUN) cargo clippy -p aura-cli --no-default-features --all-targets -- -D warnings
+	$(RUN) cargo test -p aura-cli --no-default-features
+
 .PHONY: update-lockfile
 update-lockfile: $(DOCKER_ENV) ## Regenerate Cargo.lock after version changes
 	$(RUN) cargo update --quiet --workspace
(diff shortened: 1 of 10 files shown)


Shearerbeard (dismissed):
Looks good to me - would be cool to get @justintime4tea second opinion though

justintime4tea (changes requested):
Two things to consider:

1. We probably need to update docs somewhere so that users who do want a lightweight HTTP only client have clear instructions on how to build one.
    - Before, lightweight was default and standalone required `--features=standalone-cli` but now lightweight builds requires `--no-default-features`
    - We may need to do some doc cleanup/updates where it mentions "standalone-cli" feature.
2. None of the CI/tests build the lightweight client now that the default is "standalone" - we may want to add a `--no-default-features` build check for aura-cli to the Makefile/CI so the HTTP-only path doesn't silently rot

justintime4tea (approved):

PR #275, opened 2026-06-24
Feature/191 hitl

### LINKED ISSUE #191: HITL approval gates for MCP tool calls

Shearerbeard:
Fixes #191

### DIFF
--- a/docs/hitl.md
+++ b/docs/hitl.md
@@ -1,23 +1,28 @@
 # Human-in-the-loop approval gates
 
-Human-in-the-loop (HITL) approval gates let an orchestration worker ask a
-webhook for permission before running selected MCP tools. Use them for
-operations that need a human decision before execution, such as production
-changes or destructive actions.
+Human-in-the-loop (HITL) approval gates let an orchestration worker ask for
+permission before running selected MCP tools. Use them for operations that need
+a human decision before execution, such as production changes or destructive
+actions.
 
 Current behavior:
 
-- Webhook routing works for orchestration workers.
-- Matching worker tool calls are blocked until the webhook approves them.
+- Webhook routing works for unattended orchestration-worker approvals.
+- Conversational routing works for attended orchestration-worker approvals over
+  an open SSE stream. The AURA CLI in HTTP mode is the first attended client.
+- Matching worker tool calls are blocked until the configured route approves
+  them.
 - Human denials are returned to the model as normal tool feedback, so the worker
   can explain the denial without treating it as a transport failure.
 - Timeouts, cancellation, and webhook channel failures still fail closed as tool
   errors.
-- Conversational approval over SSE is not wired yet. Use `mode = "webhook"`.
-- Single-agent config gates are not composed in this phase. The webhook gate is
-  for orchestration workers.
-- Webhook approval lifecycle events emit as `aura.approval_requested` and
-  `aura.approval_completed` on streaming responses.
+- Conversational HITL requires `stream=true`; non-streaming requests are
+  rejected because approval prompts are delivered over SSE.
+- Single-agent config gates are not composed in this phase. HITL gates are for
+  orchestration workers.
+- Approval lifecycle events emit on streaming responses. Webhook emits
+  `aura.approval_requested` and `aura.approval_completed`; conversational also
+  emits `aura.approval_pending` while the tool call is parked.
 
 ## Configure a webhook gate
 
@@ -39,7 +44,7 @@ request to the webhook before the MCP tool runs.
 
 Patterns are matched in **config order, first-match wins**. The first glob that
 matches a tool name triggers the approval request; later patterns are not
-checked. This means a broad early pattern shadows a later, more specific one:
+checked. A broad early pattern shadows a later, more specific one:
 
 ```toml
 # k8s_* matches first for every k8s tool, including k8s_get_*.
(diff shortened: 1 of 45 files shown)


justintime4tea:
Initial testing is looking good! 

Still working through the code but a couple things I've found so far:
- `docs/hitl.md` L#21–22 says "Single-agent config gates are not composed in this phase." but [this commit](https://github.com/mezmo/aura/pull/275/changes/c5a70268b39f311559328510e5e1e1689a268e53) adds single-agent support, yes? 
  - Seems promptless picked this up too? just seeing the comment from it.
- HITL and scratchpad have two different glob matching behaviors
  - HITL uses "first-match-wins" (config order)
  - scratchpad uses "longest-match-wins" (specificity)
  - Maybe we should consider a single "glob match" behavior/pattern and use that for any/all glob matching we do?

Shearerbeard:
> Initial testing is looking good!
> 
> Still working through the code but a couple things I've found so far:
> 
> * `docs/hitl.md` L#21–22 says "Single-agent config gates are not composed in this phase." but [this commit](https://github.com/mezmo/aura/pull/275/changes/c5a70268b39f311559328510e5e1e1689a268e53) adds single-agent support, yes?
>   
>   * Seems promptless picked this up too? just seeing the comment from it.
> * HITL and scratchpad have two different glob matching behaviors
>   
>   * HITL uses "first-match-wins" (config order)
>   * scratchpad uses "longest-match-wins" (specificity)
>   * Maybe we should consider a single "glob match" behavior/pattern and use that for any/all glob matching we do?

Good finds - I think we should just unify a glob module and rules - cut down the code sprawl that way too.

justintime4tea (dismissed):

docs/hitl.md
```diff
@@ -0,0 +1,289 @@
...
+timeout_secs = 300
+```
+
+`require_approval` is a list of glob patterns matched against MCP tool names.
+When an orchestration worker calls a matching tool, Aura posts an approval
+request to the webhook before the MCP tool runs.
+
+Patterns are matched in **config order, first-match wins**. The first glob that
+matches a tool name triggers the approval request; later patterns are not
+checked. A broad early pattern shadows a later, more specific one:
```
jmoses: Does this really matter? We're not matching here for anything other than "does it match", so an earlier pattern overshadowing a later pattern results in the same behavior (tool matches, hitl required).

docs/hitl.md
```diff
@@ -0,0 +1,289 @@
...
+
+```toml
+# k8s_* matches first for every k8s tool, including k8s_get_*.
+# The k8s_get_* pattern never fires.
+require_approval = ["k8s_*", "k8s_get_*"]
+
+# Reorder so read-only tools are NOT gated. k8s_get_* matches first
+# and short-circuits; k8s_* gates the rest.
+require_approval = ["k8s_get_*", "k8s_*"]
```
jmoses: Also this is also entirely incorrect, the order of these has no impact, and there's no "short circuting" here. The statement that `k8s_get_*` short circuits so "read only" tools are not gated is false.

docs/hitl.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+This differs from the scratchpad subsystem, which uses longest-match (most
+specific glob wins). HITL uses first-match because approval gating is a
```
jmoses: Again, this is incorrect.

docs/hitl.md
Shearerbeard: Pattern does not mater here - good catch updating docs as well as the follow up ticket to consolidate our glob semantics. Pushing update

jmoses (dismissed):
The doc stuff looks good to me.  I have no opinion on the actual code changes.

justintime4tea (dismissed):

justintime4tea (approved):

PR #281, opened 2026-06-24
feat(agent): ability to hide agent

### LINKED ISSUE #280: Hidden Agents
All configs loaded are presented in two manners:

- the /v1/models endpoint of the webserver 
- /model autocomplete in standalone aura-cli

Moreso the /v1/models, it would be nice to hide agents while they are being developed and/or not ready for consumption by all when hosted by the webserver.  For actual usage, if I know the name of the model, I'd like to be able to use it while "hidden"

dominic-mcallister-logdna:
summary: add hidden attribute to agents to allow them to be hidden from the models endpoint of the webserver and the model command of the cli. Still allows the model to be used by the caller if known, but to be hidden during development and/or testing phases

fixes: #280

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,6 +323,15 @@ system_prompt = "You are a DevOps expert."
 model_owner = "mezmo"        # override owned_by in /v1/models (defaults to LLM provider)
 ```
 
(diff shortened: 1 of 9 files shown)


Shearerbeard (changes requested):
Looks good, probably worth a mention in the kitchen sink toml that has all possible options

dominic-mcallister-logdna:
> Looks good, probably worth a mention in the kitchen sink toml that has all possible options

Updated both the readme and quickstart to display the new agent configuration value.

Shearerbeard (approved):

PR #282, opened 2026-06-25
fix(cli): word-wrap assistant summary headline

charlesjohnson:
The assistant summary headline was printed with a raw `println!` and relied on the terminal to wrap it, so long headlines broke mid-word at the right edge (e.g. `routin`/`g`) at narrow widths. This adds a display-width-aware, word-boundary plain-text wrapper (`ui::wrap::wrap_plain`) and routes the summary through it with a 2-space hanging indent so continuation lines align under the headline, in both the live and replay render paths. The markdown body and the summary now also reserve a 1-column right margin so text never butts against the edge. Verified visually at 80/40/160 columns, with unit tests for the wrapper and the 80-column summary case.

### DIFF
--- a/crates/aura-cli/src/repl/loop.rs
+++ b/crates/aura-cli/src/repl/loop.rs
@@ -23,7 +23,7 @@ use crate::repl::history::ConversationHistory;
 use crate::repl::input_reader::{self, HISTORY_COUNT, HISTORY_DEPTH, LAST_READLINE_INPUT};
 use crate::repl::registry::{self, CommandContext, CommandOutcome};
 use crate::tools;
-use crate::ui::markdown::render_markdown;
+use crate::ui::markdown::{render_markdown, render_summary};
 use crate::ui::prompt::{
     WaveAnimation, cleanup_terminal, clear_display_events, clear_input_hint, drain_stdin,
     erase_input_frame, extend_display_events, frame_lines, get_cumulative_tokens,
@@ -1756,13 +1756,7 @@ pub fn run_repl(
                         });
 
                         if !summary.is_empty() {
-                            println!(
-                                "{} {}",
-                                "●"
-                                    .with(random_bullet_color())
-                                    .attribute(crossterm::style::Attribute::Bold),
-                                summary.attribute(crossterm::style::Attribute::Bold),
-                            );
+                            render_summary(&summary);
                         }
                         if !displayed_text.is_empty() {
                             println!();
--- a/crates/aura-cli/src/repl/loop.rs
+++ b/crates/aura-cli/src/repl/loop.rs
@@ -42,30 +42,11 @@ use crate::ui::prompt::{
 };
 use crate::ui::welcome::WelcomeState;
 
-/// Event-driven watcher that repaints the frame after a terminal resize while
-/// the REPL is idle in `readline()`.
+/// Repaints the frame after a terminal resize while idle at the prompt.
 ///
-/// rustyline wakes on SIGWINCH but only repaints its own input line, and only
-/// when that line is long enough to need it — so at an idle/short prompt the
-/// frame borders and status row stay drawn at the old width until the next
-/// keystroke. We get our own SIGWINCH notification by blocking on a
-/// `signal_hook` `Signals` stream (no polling): because rustyline is built with
-/// its `signal-hook` feature, it registers SIGWINCH via signal_hook's chaining
-/// pipe, so our consumer fires on the same signal even while readline is
-/// blocked.
-///
-/// On each resize we rebuild the status bar at the new width then do a full,
-/// absolute repaint: clear the screen + replay the conversation + redraw the
-/// frame (the same primitive `/style` uses), wrapped in a synchronized update
-/// so it doesn't flicker. We deliberately avoid the incremental
-/// `handle_resize_frame` used by the keystroke path: it relies on
-/// `SavePosition`/`RestorePosition`, and when a resize reflow scrolls the
-/// screen the saved absolute row is stale, so the frame drifts (ending up near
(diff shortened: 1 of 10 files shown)


charlesjohnson:
Screencaps! All 80x24; if you need wider / narrower I can do those too.

### Before
Screenshot 2026-06-25 at 9 57 17 AM

### After
Screenshot 2026-06-25 at 9 58 25 AM

justintime4tea (changes requested):
I tested locally and it appears to do what the PR says it does.

The one thing I'd like to see cleaned up is the large volume of "what and how" comments. The code should tell us what is happening and how, comments should be reserved for "why", if something isn't obvious at face value. There is a ton of comments referencing things like screenshot in an issue, previous behavior, new behavior (which may change and is why comments shouldn't describe what/how) and in general just way too verbose on the comments. This is common for AI generated code - AI tends to write comments defending the code itself wrote VS the desired lightweight "why" comments.

It's fine to have light/minimal comments at the top of a struct or function describing what it is or its purpose but I feel we need to cleanup the massive amount of AI generated comments in this PR.

The rule of thumb to apply line-by-line: if removing the comment and reading just the code would make a reasonable Rust reader come to the same conclusion, cut it. Keep a comment only where the code's intent could be misread or where it represents a decision (rejected alternative, race condition, external constraint) that isn't visible in the diff itself.

crates/aura-cli/src/repl/loop.rs
```diff
@@ -23,25 +23,112 @@ use crate::repl::history::ConversationHistory;
...
+/// so it doesn't flicker. We deliberately avoid the incremental
+/// `handle_resize_frame` used by the keystroke path: it relies on
+/// `SavePosition`/`RestorePosition`, and when a resize reflow scrolls the
+/// screen the saved absolute row is stale, so the frame drifts (ending up near
+/// the top with cumulative scroll). The absolute clear+replay can't drift, and
+/// for an idle single-line prompt rustyline's forced refresh re-anchors to the
+/// cursor we leave at the input line. We only act while idle in `readline()`
+/// with no request processing, so it never interleaves with mid-turn output or
+/// the streaming animation (which has its own resize handling).
+struct ResizeWatcher {
```
justintime4tea, line 50: An example for how to cleanup these comments could be simply:

```rust
/// Repaints the frame after a terminal resize while idle at the prompt.
///
/// Uses an absolute clear+replay+redraw rather than the incremental
/// resize path, which drifts when a reflow scrolls the screen (the
/// saved cursor position becomes stale).
```

crates/aura-cli/src/ui/state.rs
```diff
@@ -338,6 +338,26 @@ pub fn set_processing(active: bool) {
...
 
+/// Whether a request is currently being processed.
+pub fn is_processing() -> bool {
+    PROCESSING.load(Ordering::Relaxed)
+}
+
+/// Whether the REPL is currently blocked in `readline()` waiting for input
+/// (i.e. idle at the prompt). The resize watcher only redraws the frame while
+/// this is true, so it never interleaves with mid-turn output or streaming.
+pub(crate) static READLINE_ACTIVE: AtomicBool = AtomicBool::new(false);
```
justintime4tea, line 348: Example of cleaning up this comment:

```rust
/// Whether the REPL is idle in `readline()`. Used to gate the resize
/// watcher so it doesn't repaint mid-turn.
```

crates/aura-cli/src/ui/markdown.rs
```diff
@@ -64,10 +68,80 @@ pub fn render_markdown(text: &str) {
...
+
+/// Print the assistant summary headline: a colored "● " marker followed by the
+/// summary in bold, word-wrapped to the terminal width so it never breaks
+/// mid-word. Continuation lines use a 2-space hanging indent so they align
+/// under the headline text (matching the markdown body's indent). A
+/// [`RIGHT_MARGIN`] column is reserved on the right.
+///
+/// The marker and indent are both 2 columns wide, so every line shares the
+/// same available text width.
+pub fn render_summary(summary: &str) {
```
justintime4tea, line 84: Example of cleaning up this verbose comment:

```rust
/// Print the assistant summary headline, word-wrapped to terminal width.
```

The code contained within describes how things are printed, we can update the code later and not need to continuously sync the comments with the code.

crates/aura-cli/src/repl/loop.rs
charlesjohnson: Tightened to the drift rationale + idle gating, dropped the mechanism walkthrough. 605b48e

crates/aura-cli/src/ui/state.rs
charlesjohnson: Trimmed to two lines (idle-in-`readline()`, gates the watcher). 605b48e

crates/aura-cli/src/ui/markdown.rs
charlesjohnson: Cut to a one-line summary, keeping the word-wrap intent. 605b48e

crates/aura-cli/src/repl/loop.rs
charlesjohnson: Applied your exact wording. f6bf70b

crates/aura-cli/src/ui/markdown.rs
charlesjohnson: Cut to the single line you suggested (and dropped the helper's doc too). a4032ab

Shearerbeard:
@justintime4tea are you satisfied on the revisions for this one?

justintime4tea (approved):

justintime4tea:
> @justintime4tea are you satisfied on the revisions for this one?

I gave it an approval just now 👍 LGTM

PR #283, opened 2026-06-25
feat(scratchpad): provider-accurate token counting for gemini and claude

### LINKED ISSUE #257: Provider-accurate token counting for Gemini and Anthropic
## Summary

AURA's local token counting is exact for OpenAI models but falls back to OpenAI's
`o200k_base` BPE for every other provider. This issue tracks adding
provider-accurate counting, starting with Gemini and Anthropic.

## Background

All token counting funnels through a single trait and a single provider-dispatch
function, so this is a well-contained change:

- `TokenCounter` trait — `crates/aura/src/scratchpad/context_budget.rs:20`
- `TiktokenCounter` (the only production impl) — `context_budget.rs:44`
- `token_counter_for_provider(provider, model)` — `context_budget.rs:96`

The dispatch function is where accuracy is decided. Today every non-OpenAI
provider hits the catch-all arm and is approximated with `o200k_base`:

```rust
pub fn token_counter_for_provider(provider: &str, model: &str) -> Arc<dyn TokenCounter> {
    match provider {
        "openai" => Arc::new(TiktokenCounter::for_model(model)),
        // Future: add dedicated tokenizers for other providers here
        // "anthropic" => Arc::new(ClaudeCounter::new()),
        _ => Arc::new(TiktokenCounter::for_model(model)), // -> o200k_base
    }
}
```

Anthropic and Gemini tokenize differently from `o200k_base`; the error commonly
runs 10–30% on code, whitespace-heavy, and non-English text. The fix is to
implement `TokenCounter` for each provider and add match arms here. New counters
are then picked up automatically at the call sites in
`crates/aura/src/builder.rs` and `orchestration/orchestrator.rs`.

## Gemini — clean local solution

Use the [`gemini-tokenizer`](https://crates.io/crates/gemini-tokenizer) crate.
It embeds the Gemma 3 SentencePiece model (262,144-token vocab) via
`include_bytes!`, runs fully local with no network or external files, and
produces counts identical to Google's official Python SDK (`local_tokenizer.py`).
It also handles function calls/responses and tool schemas, not just plain text.
This is a drop-in `GeminiCounter` implementing `TokenCounter`.

## Anthropic — no accurate public local tokenizer

Claude 3+ models do not ship a public tokenizer. Note the
[`claude-tokenizer`](https://crates.io/crates/claude-tokenizer) crate exists but
wraps the legacy Claude 1/2 BPE — it is **not** accurate for current Claude
models, so we should not use it. Realistic options:

1. **Anthropic Count Tokens API** (`/v1/messages/count_tokens`) — exact, and the
   only authoritative source. Cost is a network round-trip, so a `ClaudeCounter`
   built on it needs response caching and must degrade to a local fallback when
   offline or rate-limited. (Also available via Amazon Bedrock for the Bedrock
   provider path.)
2. **Calibrated approximation** — keep using a local BPE (`cl100k_base` or
   `o200k_base`) scaled by an empirically measured correction factor (community
   implementations commonly use `cl100k_base × ~1.1`). No network, no exactness,
   but materially better than the raw `o200k_base` fallback and a good interim
   `ClaudeCounter` / offline fallback for option 1.

The approach for Anthropic is intentionally left open pending a decision on the
network-dependency vs. accuracy trade-off.

## Fallback

Keep `o200k_base` as the final arm for unrecognized models — it is the safety
floor that prevents a missing or garbled provider usage field from collapsing the
estimate to zero.

jakedipity:
## Summary

Local token counting was exact only for OpenAI; every other provider fell back to `o200k_base`, which runs 10–30% off on code, whitespace-heavy, and non-English text. This adds provider-accurate counting for Gemini and Claude.

All counting still funnels through `token_counter_for_provider(provider, model)` in `crates/aura/src/scratchpad/context_budget.rs`, so the call sites in `builder.rs` and `orchestration/orchestrator.rs` pick up the new counters automatically.

## Changes

- **Gemini** — `GeminiCounter` wraps the [`gemini-tokenizer`](https://crates.io/crates/gemini-tokenizer) crate (embedded Gemma 3 SentencePiece, 262k vocab, fully local, exact — matches Google's official SDK). The configured model id is passed through to the crate; an id it doesn't recognize degrades to a current Gemini model rather than erroring.
- **Anthropic / Bedrock-Claude** — `ClaudeApproxCounter` scales a local `cl100k_base` count by a calibrated `1.1` factor, since Claude ships no public local tokenizer. The factor is an isolated, documented `const`. Bedrock hosts multiple vendors, so only its Claude model ids route here; non-Claude Bedrock models stay on the fallback.
- **Fallback** — `o200k_base` remains the conservative floor for everything else (Ollama, OpenRouter, non-Claude Bedrock, unknown providers).

## Tests

`cargo test -p aura --lib` — all pass. New tests cover the dispatch table (per-provider counter invariants), Bedrock model-id routing, that Bedrock-Claude counts identically to native Anthropic, and that the Claude approximation never falls below its `cl100k_base` baseline.

Closes #257


Shearerbeard (dismissed):

jakedipity:
Simplified the Dockerfile a bit, needs a re-review.

Shearerbeard (approved):

PR #284, opened 2026-06-25
feat(cli): show version and server info in startup banner

charlesjohnson:
Replaces the working-directory line in the aura-cli startup banner with the CLI version and the connected server: the remote URL plus its version when reachable (e.g. `http://localhost:8080 (server v1.30.0)`), or `standalone` in standalone mode. The web server `/health` endpoint now reports its own `aura_version` so the CLI can display it, fetched best-effort with a short timeout that falls back to just the URL if the server is unreachable or older. The banner template's legacy `{{PWD}}` token is migrated to a new `{{AURA_STATUS}}` token, so existing seeded `~/.aura/.welcome` files switch over automatically. Verified manually across standalone, HTTP-reachable, and HTTP-down cases; `cargo fmt`, `clippy`, and the lib test suite (including 3 new welcome tests) pass.


Shearerbeard (approved):
Took me a bit to understand the semantics of what the .welcome file is (pre existing) but this looks good.

PR #286, opened 2026-06-26
fix(otel): decode agent turn prompts

### LINKED ISSUE #273: Otel Traces Worker -> agent.turn prompt is double escaped and hard to read
If you run phoenix and export our OpenInference otel in orchestration mode you will see some very hard to read initial worker prompts. This makes it pretty hard to debug the coordinator to worker delegation.

Example:

```
{"role":"user","content":[{"type":"text","text":"BACKGROUND (read-only, do not act on this): Create /app/hello.txt with the content \"Hello, world!\"\n\nYOUR TASK: Run the following command to create the file /app/hello.txt with the content \"Hello, world!\":\n\n```\necho 'Hello, world!' > /app/hello.txt\n```\n\nAfter running the command, capture the terminal output to confirm it completed without errors.\n\n\n\nUse your tools when the task requires fetching, generating, or computing new data. If your task is to analyze, summarize, or transform results already provided above, reason directly. When a tool exists for the operation, always prefer the tool. Large tool outputs may be saved to artifact files — if you see a `[Tool output saved to artifact: ...]` reference, use `read_artifact` to retrieve the full content. When you have your answer, call `submit_result` with a concise summary, your complete findings, and your confidence level.\n"}]}
```

Shearerbeard:
Fixes orchestration.worker agent turn showing
raw prompt json instead of agent.turn Input
Messages so its readable

Fixes: #273

### DIFF
--- a/crates/aura/src/openinference_exporter.rs
+++ b/crates/aura/src/openinference_exporter.rs
@@ -138,8 +138,10 @@ fn transform_span(mut span: SpanData) -> SpanData {
 
     // 2. Translate gen_ai.* → llm.* / tool.* (additive)
     let mut extra_attrs: Vec<KeyValue> = Vec::new();
+    let mut deferred_turn_prompt: Option<String> = None;
     let mut deferred_response: Option<String> = None;
     let mut deferred_reasoning: Option<String> = None;
+    let mut has_input_messages = false;
 
     for kv in &span.attributes {
         let key = kv.key.as_str();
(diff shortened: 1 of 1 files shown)


crates/aura/src/openinference_exporter.rs
```diff
@@ -278,32 +290,154 @@ fn truncate_string_values(attrs: &mut [KeyValue], limit: usize) {
...
+                    (part.get("type").and_then(|value| value.as_str()) == Some("text"))
+                        .then(|| part.get("text").and_then(|value| value.as_str()))
+                        .flatten()
+                })
+                .collect();
+            (!text_parts.is_empty()).then(|| text_parts.join("\n\n"))
         }
+        _ => None,
     }
 }
```
jakedipity, line 442: This drops non text message types which the previous code did not. Do we want to have a fallback for those types?

Shearerbeard: This works for our use cases - we don't support rigs non text functionality underneath and when we cut over to a better provider most of this code module gets deleted anyway

jakedipity (approved):

Issue #288, opened 2026-06-26
[BUG]: Quickstart doesn't work as described


gjanco:
### Summary

The three step quickstart in the readme does not work.  It fails with this error:
```
Error: Failed to load agent config

Caused by:
    0: IO error: No such file or directory (os error 2)
    1: No such file or directory (os error 2)
```

### Reproduction

**Steps**
1. clone the AURA repo
2. follow the steps in the quickstart section of the readme (including setting your LLM and key

### Relevant log output

```shell
Error: Failed to load agent config

Caused by:
    0: IO error: No such file or directory (os error 2)
    1: No such file or directory (os error 2)
```

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #289, opened 2026-06-26
fix: quickstart instructions

gjanco:
Adjusting quickstart instructions so they point to running AURA webserver.

Ref: #288

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,9 @@ Key capabilities:
 ## Quick Start
 
 ```bash
-cp .env.example .env            # set your LLM provider, model, and API key
-docker compose up -d            # starts Aura (orchestrator mode) + LibreChat + Phoenix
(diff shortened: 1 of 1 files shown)


charlesjohnson:
Now that aura-cli defaults to standalone, I don't think we need this.

That said we do need to make sure the init --> run path works with QuickStart.

charlesjohnson:
Also we might not need a full compose anymore for QuickStart, just a single container?

gjanco:
> Now that aura-cli defaults to standalone, I don't think we need this.
> 
> That said we do need to make sure the init --> run path works with QuickStart.

That is all I am doing with this change - is fixing the flow if someone is following the quickstart instructions at the top of the readme.  If there is a new way, lets talk about that.

gjanco:
Started #290 to discuss.

jakedipity (approved):

PR #292, opened 2026-06-28
doc: add xquik mcp example

kriptoburak:
## What

Adds `examples/complete/x-intelligence.toml`, a complete Aura agent config that connects to the Xquik Streamable HTTP MCP endpoint with `XQUIK_API_KEY`.

Updates the examples index so users can discover the new config next to the existing complete agent examples.

## Why

Aura already supports remote MCP servers with environment-based auth. This gives teams a ready-made public X research example without changing runtime defaults or requiring local sidecars.

## Validation

- `taplo check examples/complete/x-intelligence.toml examples/README.md`
- `aura-config` package tests passed: 120 tests
- Changed-file whitespace check passed
- Xquik live MCP manifest checked at `https://xquik.com/.well-known/mcp.json`


andrewelizondo:
Thanks for the contribution, but we're not going to take this one. AURA is an agent harness built specifically for SRE work: incident investigation, root cause analysis, and production operations. A social media research example doesn't fit that scope, and we want the examples directory to reflect the use cases the project is actually built for. If you want to rework this using an SRE-adjacent MCP endpoint, happy to review that.

Closing this as out of scope.

PR #298, opened 2026-06-29
feat(build): add macos release binaries

### LINKED ISSUE #297: [FEATURE]: Provide macOS binaries
## Summary

Publish prebuilt macOS binaries for AURA so users on macOS can run `aura-web-server` and `aura-cli` without building from source.

## Motivation

Release artifacts are currently only available for Linux. macOS users have no prebuilt option and must compile the project themselves, which is a meaningful barrier to getting started and to local development.

## Request

Provide downloadable macOS binaries for both `aura-web-server` and `aura-cli`, covering:
- Apple Silicon (arm64)
- Intel (x86_64)

## Acceptance criteria

- macOS binaries for `aura-web-server` and `aura-cli` are published alongside the existing Linux artifacts in releases.
- Both binaries run on a clean macOS machine (Apple Silicon and Intel).

jakedipity:
Add darwin amd64/arm64 to the release set, cross-compiled per target with the runner image's linux toolchains and the macOS SDK.

Guard each target against the wrong build host so an unsupported combination fails fast instead of emitting a mislabeled binary: linux builds require an amd64 linux host, darwin builds require macOS.

Closes: #297

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -60,7 +60,7 @@ pipeline {
         jiraSendBuildInfo site: 'logdna.atlassian.net'
         sh 'ls -alh -R report'
         archiveArtifacts allowEmptyArchive: true, artifacts: 'report/ci/**', caseSensitive: false, followSymlinks: false
-        sh: 'make clean'
+        sh 'make clean'
         if (env.SANITY_BUILD == 'true') {
           notifySlack(
             currentBuild.currentResult,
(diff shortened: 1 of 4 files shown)


release.config.js
```diff
@@ -29,10 +26,9 @@ module.exports = {
   branches: ['main'],
 
   // https://github.com/semantic-release/exec
-  prepareCmd: `./scripts/set-version.sh \${nextRelease.version}; sleep 2; ${buildReleaseBinariesCmd}`,
-  // On a dry run (no prepare step), build the release binaries during verify so
-  // a broken cross-compile is caught before the change reaches the release branch.
-  verifyReleaseCmd: `if [ "\${options.dryRun}" = "true" ]; then ${buildReleaseBinariesCmd}; fi`,
+  // Release binaries are built and stashed by the pipeline before this runs,
+  // then unstashed into dist/release for the github plugin to attach.
```
justintime4tea: Is it unstashed into dist/release or just dist? This is definitely not my wheel-house so I may be missing something, but it looks like things unstash into just dist/ no?

Jenkinsfile
```diff
@@ -313,19 +374,70 @@ pipeline {
...
+
+        stage('Semantic Release') {
+          steps {
+            sh 'rm -rf dist'
+            sh 'mkdir -p dist'
+
+            unstash 'linux-release-artifacts'
+            unstash 'darwin-release-artifacts'
+
+            sh 'make build-checksums'
```
justintime4tea: This only runs build-checksums, not verify-binaries — so unlike the dry-run path (L282) it never asserts the full binary matrix is present before publishing. 
- allowEmpty: false on the stashes guards against empty darwin artifacts, but not a partial set (e.g. web-server present, cli missing).
- Since verify-binaries depends on build-checksums, this can just become sh 'make verify-binaries' 
   — strict superset, same checksums.txt, plus the presence check. 
   - Cheap insurance against shipping an incomplete release.

Shearerbeard (dismissed):

Shearerbeard (approved):

PR #304, opened 2026-07-01
feat(cli): show orchestration workers in the startup banner

charlesjohnson:
The CLI startup banner now lists the active orchestration agent's workers and their descriptions in both standalone and HTTP modes, annotating any worker that overrides the coordinator model and wrapping long descriptions onto hang-indented continuation lines. `aura-config` gains `Config::worker_overview()`, and `/v1/models` entries carry a `workers` array plus a top-level `default`/`routes_any_model` routing hint that is gated on the default being visible so a hidden agent is never revealed by discovery. The CLI resolves workers for the selected model — falling back to the backend default when none is selected, or any model under single-config passthrough — via a bounded `/v1/models` fetch that cannot hang REPL launch.


charlesjohnson:
Screenshots!

Screenshot 2026-07-01 at 9 11 09 AM

Screenshot 2026-07-01 at 9 11 47 AM

andrewelizondo (approved):
lgtm 👍

Shearerbeard (changes requested):
Formalizing a review - adding workers with their models to the /v1/models endpoint is a compatibility break for the openai spec endpoint we expose. This might not be a huge issue for some looser clients but implicitly creates a maintenance burden to ensure popular clients can comply. I'll work out an alternative strategy that utilizes our internal SSE events channel and accomplishes the same UI.

Shearerbeard:
Attached issue for the non /v1 implementation https://github.com/mezmo/aura/issues/313 (in progress)

charlesjohnson:
Superseded by #319

Issue #325, opened 2026-07-09
Session storage (cross-pod state)


justintime4tea:
AURA needs session storage. The official/supported backend to support that session storage, and how session storage will work, needs to be decided on. An ADR needs to be created which addresses the need for session storage and then a first iteration is needed which satisfies this ADR.
 
- AURA today keeps every piece of cross-request state **in one pod's RAM**. Behind a
  load balancer with no session affinity, that state is invisible to the other pods.
- Two subsystems break in a multi-pod deployment **today**: **HITL conversational
  approvals** and the **A2A task registry**. Both are explicitly flagged in code as
  "not durable / TBD resilient location".
- The fix is a small set of **AURA-owned storage traits**. We can ship a specific implementation, such as one backed by Redis/Valkey, and the existing in-memory behavior stays as the default `impl`; anyone
  can write their own backend against the traits.
- A shared store alone is not sufficient. Two mechanisms are needed together:
  1. a **durable store** (survives restart, readable by any pod), and
  2. an **event bus** (pub/sub) to **wake** the pod that owns a suspended request and to fan out streaming updates across pods. An implementation using Redis/Valkey would give us both from one connection but there may be other choices we'd like to entertain so this is open for the ADR to explore.

## Phased rollout (Sub-issues)
Each phase is a self-contained issue/PR. Phase 1 is the prerequisite for everything else;
phases 2 and 3 are independent and can land in either order; phase 4 needs both.

1. **Traits + in-memory refactor (no behavior change).** Introduce `SessionStore`,
   `ApprovalStore`, `EventBus`; move today's structures behind them; A2A already has
   `TaskStore`. Ship with `backend = memory`. This is a pure refactor — existing tests are
   the guard.
2. **Redis/Valkey backend — A2A task store.** Smallest, highest-value cross-pod win
   (`message:send` + poll works across pods). No bus needed for the poll path.
3. **Redis/Valkey backend — HITL approvals (store + wake bus).** Cross-pod conversational
   approvals via `approval:{id}` pub/sub. Fail-closed preserved.
4. **A2A streaming/cancel over the bus.** Cross-pod `message:stream` / `subscribe` /
   `cancel` via the `a2a:task:{id}` (fan-out) and `a2a:cancel:{id}` (routed cancel) topics.
   Depends on the shared task store (phase 2) and the `EventBus` impl (phase 3).
5. **Helm/Valkey packaging + docs.** Values, Secret wiring, `/health`, runbook.

PR #326, opened 2026-07-09
doc(adr): add session-storage cross-pod state decision

justintime4tea:
Record the decision to put cross-pod session state behind Aura-owned traits, with an in-process default and one shipped networked backend (Redis/Valkey), recognizing a durable store and an event bus as distinct capabilities.

Add a companion design note under docs/design/ carrying the trait method sets, backend key/topic schema, config and wire formats, wiring changes, and phased rollout. The ADR holds the durable decision; the note follows the implementation.

Covers HITL conversational approvals and the A2A task surface, the two subsystems that break multi-pod today.

Ref: GH-325

### DIFF
--- a/docs/design/session-storage.md
+++ b/docs/design/session-storage.md
@@ -1,0 +1,519 @@
+<!-- markdownlint-disable MD033 -->
+# Session storage (cross-pod state): design and implementation note
+
+Companion to the ADR [2026-07-08-session-storage](../adr/2026-07-08-session-storage.md).
+The ADR records the decision (cross-pod state behind Aura-owned traits, in-process
+default, one shipped networked backend, store plus event bus as distinct
+capabilities). This note carries the trait method sets, the backend key/topic
+schema, the config and wire formats, the wiring changes, and the phased rollout.
+
+**Status:** living design note, current as of 2026-07-08. The ADR holds the
+durable decision. Nothing here is built yet — the Rust signatures, key schema, and
+diagrams below describe intended design, not code that exists. References to
+existing aura code (file:line, module paths) are where-to-look pointers and move as
+the code does.
+
+**Goal of iteration 1:** make the state AURA _already has_ work across pods, so a
+load-balanced multi-replica deployment (Helm prod runs 3–10 replicas, no session
+affinity) behaves the same as a single process. Do it behind **traits** so the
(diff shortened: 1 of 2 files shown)


docs/design/session-storage.md
```diff
@@ -0,0 +1,519 @@
...
+   useful for polling paths), but let one backend satisfy both — Redis/Valkey does.
+4. **Serialize the record, never the runtime handle.** Only plain data crosses the pod
+   boundary (`ParkedApproval`, A2A `Task`). Live handles (`oneshot::Sender`, tokio
+   `broadcast`, `CancellationToken`) stay pod-local and are re-established via the bus.
+5. **Fail closed, degrade to today.** If the store/bus is unavailable, HITL still fails
+   closed (deny), and a single-pod deployment still works. No new hard dependency for the
+   default path.
+6. **Small blast radius.** Iteration 1 touches only the two subsystems that are broken
+   multi-pod. Request-scoped brokers and on-disk artifacts are explicitly out of scope
+   (see §3).
```
jakedipity, line 103: A lot of this just feels like a rehashing of the ADR

justintime4tea: The main difference is that the ADR doesn't have design/impl details, the `docs/design/session-storage.md` doc does. The design/impl doc will help guide the implementation and can be updated during development. This allows us to keep the "what we want to build" with the "how we want to build it" separate. Finally, we'll need a `docs/session-storage.md` like HITL and telemetry, that is user-facing and all about what it means for a user and how they would use it.

This mirrors how the HITL and telemetry (to some degree) ADR and implementation work was done.

HITL
docs/adr/2026-06-16-hitl-approval-architecture.md - **ADR**
docs/design/hitl.md - **Design**
docs/hitl.md - **User-facing**

Telemetry
docs/adr/2026-06-23-cli-product-telemetry.md - **ADR**
(missing design doc)
docs/telemetry.md - **User-facing**

Session Storage
docs/adr/2026-07-08-session-storage.md - **ADR**
docs/design/session-storage.md **Design**
docs/session-storage.md **User-facing**

The telemetry is a little different in that there wasn't a design doc formally included.

After the session-storage work is completed, we can create a `docs/session-storage.md` and make sure it gives consumers the information they need about what session-storage is, how it works, and how to take advantage of it.

Shearerbeard (approved):

PR #333, opened 2026-07-09
fix(cli): show welcome banner when no .welcome file on disk

### LINKED ISSUE #294: [FEATURE]: CLI "Welcome Banner" when using docker
### Summary

Right now when I start the CLI, all I get is a prompt:
Image

I would like to see:
- the AURA logo (this is missing when running the quickstart instructions
- the current version of the CLI
- the current version of the aura web server
- at a minimum, the current config I have running
- tips for getting started

Bonus
- which specific LLM I am using
- a list of "whats new" since the last time I ran AURA

### Additional Context

_No response_

### Upload screenshots

_No response_

### Searched Issues

- [x] No similar issues found

### Code of Conduct

charlesjohnson:
The welcome banner content is embedded in the CLI binary, but it was only ever read back from disk after seeding `~/.aura/.welcome`. In the Docker release image the runtime user (`appuser`) has no writable home directory, so seeding fails silently and, with no `.welcome` in the working dir either, the banner was skipped entirely. This change falls back to the embedded content when neither the cwd nor the seeded home file is readable, so the banner renders in the container and on any read-only filesystem. The resolution logic is extracted into `resolve_welcome_content` with unit tests covering the source precedence.

Fixes #294

### DIFF
--- a/crates/aura-cli/src/ui/welcome.rs
+++ b/crates/aura-cli/src/ui/welcome.rs
@@ -51,7 +51,7 @@ impl WelcomeState {
     /// bright, high-contrast complement of the art's dominant hue.  Otherwise
     /// a random palette color is used.
     pub fn pick() -> Option<Self> {
-        let content = pick_welcome_file()?;
+        let content = resolve_welcome_content()?;
         let (text_color, art_color) = if content.contains('\x1b') {
             let complement = complementary_text_color(&content);
             let (_, fallback) = random_color_pair();
@@ -605,21 +605,21 @@ fn interpolate_tokens(content: String, status: &str) -> String {
 }
 
 /// Load the `.welcome` file from the current directory (preferred) or
-/// `~/.aura/` (fallback), and interpolate env vars.
-fn pick_welcome_file() -> Option<String> {
+/// `~/.aura/` (fallback), and interpolate env vars. For read-only filesystems,
+/// fall back to the color built-in welcome message.
+fn resolve_welcome_content() -> Option<String> {
     let aura_dir = aura_home_dir();
     seed_home_welcome(&aura_dir);
 
-    let cwd_welcome = Path::new(".welcome");
-    let home_welcome = aura_dir.join(".welcome");
+    let cwd_content = fs::read_to_string(Path::new(".welcome")).ok();
+    let home_content = fs::read_to_string(aura_dir.join(".welcome")).ok();
+    let non_empty = |c: String| Some(c).filter(|c| !c.trim().is_empty());
 
-    let content = if cwd_welcome.exists() {
-        fs::read_to_string(cwd_welcome).ok()
-    } else {
-        fs::read_to_string(&home_welcome).ok()
-    };
+    let content = cwd_content
+        .and_then(non_empty)
+        .or_else(|| home_content.and_then(non_empty))
+        .unwrap_or_else(|| BUILTIN_WELCOME_FILES[0].1.to_string());
 
-    let content = content.filter(|c| !c.trim().is_empty())?;
     Some(interpolate_tokens(content, &super::state::startup_status()))
 }
 


crates/aura-cli/src/ui/welcome.rs
```diff
@@ -27,9 +27,11 @@ const MAX_STAGGER: f32 = 0.55;
 const MIN_EASE: f32 = 1.2;
 const MAX_EASE: f32 = 4.0;
 
+const BUILTIN_WELCOME: &str = include_str!("../../.welcome");
```
justintime4tea: The name is kind of a misnomer because there isn't "a" built-in welcome file, there are two. That said, let's leave this as part of the `BUILTIN_WELCOME_FILES` and not pull it out into another const.

crates/aura-cli/src/ui/welcome.rs
```diff
@@ -604,22 +606,22 @@ fn interpolate_tokens(content: String, status: &str) -> String {
     result
 }
 
-/// Load the `.welcome` file from the current directory (preferred) or
-/// `~/.aura/` (fallback), and interpolate env vars.
+/// Ensure banner still displays on read-only filesystems.
+fn resolve_welcome_content(cwd: Option<String>, home: Option<String>) -> String {
```
justintime4tea: We can go ahead and do this in-line in the `pick_welcome_file` function.

crates/aura-cli/src/ui/welcome.rs
```diff
@@ -604,22 +606,22 @@ fn interpolate_tokens(content: String, status: &str) -> String {
...
+    let non_empty = |c: String| Some(c).filter(|c| !c.trim().is_empty());
+    cwd.and_then(non_empty)
+        .or_else(|| home.and_then(non_empty))
+        .unwrap_or_else(|| BUILTIN_WELCOME.to_string())
+}
+
 fn pick_welcome_file() -> Option<String> {
```
justintime4tea: Since we're not actually picking a file here but rather resolving welcome content, let's go ahead and change the name of this function to `resolve_welcome_content` - also, let's in-line the function added on line 610 and use the existing BUILTIN_WELCOME_FILES instead of the separate (new) BUILTIN_WELCOME.

All together the refactor should look something like this:
```rust
/// Load the `.welcome` file from the current directory (preferred) or
/// `~/.aura/` (fallback), and interpolate env vars. For read only
/// filesystems, fallback to the color built-in welcome message.
fn resolve_welcome_content() -> Option<String> {
    let aura_dir = aura_home_dir();
    seed_home_welcome(&aura_dir);

    let cwd_content = fs::read_to_string(Path::new(".welcome")).ok();
    let home_content = fs::read_to_string(aura_dir.join(".welcome")).ok();
    let non_empty = |c: String| Some(c).filter(|c| !c.trim().is_empty());

    let content = cwd_content
        .and_then(non_empty)
        .or_else(|| home_content.and_then(non_empty))
        .unwrap_or_else(|| BUILTIN_WELCOME_FILES[0].1.to_string());

    Some(interpolate_tokens(content, &super::state::startup_status()))
}
```

justintime4tea (changes requested):

crates/aura-cli/src/ui/welcome.rs
charlesjohnson: Done — dropped the separate `BUILTIN_WELCOME` const; the fallback now uses `BUILTIN_WELCOME_FILES[0].1`.

crates/aura-cli/src/ui/welcome.rs
charlesjohnson: Done — inlined into `resolve_welcome_content`, no separate helper.

crates/aura-cli/src/ui/welcome.rs
charlesjohnson: Done, applied essentially your snippet: renamed to `resolve_welcome_content`, inlined the resolution, and used `BUILTIN_WELCOME_FILES[0].1` for the read-only-fs fallback.

One note: I dropped the 3 unit tests I had added for the old standalone resolver. They passed `cwd`/`home` as args; now that resolution reads the filesystem inline, the builtin-fallback branch is not reachable from a unit test (`seed_home_welcome` writes `~/.aura/.welcome` on the test machine). I re-verified the fallback by driving the binary with `HOME=/dev/null` and no `.welcome` in the cwd — logo and footer both render. Happy to reintroduce a seam if you would prefer to keep that path unit-tested.

justintime4tea (approved):

PR #334, opened 2026-07-09
doc: split readme into development and contributing docs

charlesjohnson:
Moves architecture, build, Makefile, and testing documentation out of README.md and CONTRIBUTING.md into a new DEVELOPMENT.md, trims CONTRIBUTING.md down to the contribution process itself, and adds a which-docs-go-where map to CLAUDE.md. 

The README also gets a new capability-focused header, loses its redundant table of contents (GitHub will build us one), and its badges gain alt and title attributes. 

Review-driven accuracy fixes are included: 
- the FIRST_CHUNK_TIMEOUT_SECS default is 90 (not 30)
- the web-server integration test recipe now includes the required feature flags and A2A environment variables, 
- contributors are pointed at cargo test --workspace since the make test hook is empty. 

Stale references to the removed docs/rig-tool-execution-order.md (including two Rust doc comments) now point at docs/rig-fork-changes.md. 

The Ref: footer is documented as optional now that commitlint no longer requires it.


charlesjohnson:
Since this was docs-only I updated the docs to reference current behavior, rather than correct behavior to match the docs. 

For example, while I don't think replacing `make test` with running cargo tests directly is the right move, the point was docs accuracy, so that's what this does.

Shearerbeard (approved):

PR #335, opened 2026-07-09
build: include aura-cli in default tests and guard integration runs

charlesjohnson:
Wires the workspace so root `cargo`/`make` commands actually cover
aura-cli, and hardens the integration-test entry points against silent
gaps.

## Changes

- `aura-cli` joins `default-members`, so root `cargo build` / `cargo
  test` cover the whole workspace (CLI regressions now show up in a
  plain workspace run).
- `make test` aggregates `cargo test --workspace` via a new `test-rust`
  hook, so `make ci` runs the tests its help text already claims.
- Renames the self-contained STDIO suite to `test-stdio-local` so it no
  longer implies the Compose / up / down siblings the Docker-backed
  suites carry.
- Adds an ignored marker test in aura-web-server that surfaces when no
  integration feature is set, instead of a misleading green zero-test
  run.

## Benchmark: cost of adding aura-cli to default-members

Clean debug builds of the `default-members` set, with and without
aura-cli:

| Build                    | Crate-units | Distinct crates |
| ------------------------ | ----------- | --------------- |
| clean, without aura-cli  | 403         | 359             |
| clean, with aura-cli     | 436         | 390             |
| delta                    | +33 (~8%)   | +31             |

Every net-new crate is a small terminal-UI / REPL lib (crossterm,
rustyline, termimad, crokey, ...). There is no new async runtime and no
extra AWS / LLM / MCP stack: that heavy stack is already shared.
Wall-clock delta was within run-to-run noise, so crate count is the
reliable signal.

Keeping aura-cli in the default set also aligns the default feature set
with `cargo test --workspace` (which always includes it), avoiding a
feature-unification recompile of ~85 shared crates when alternating the
two.

## Verification

`cargo test --workspace`: 1568 passed, 16 ignored, 0 failed. The marker
test shows as 1 ignored with no feature and 0 tests with a suite
feature enabled.

Refs #334

### DIFF
--- a/.makefiles/aura.mk
+++ b/.makefiles/aura.mk
@@ -54,8 +54,8 @@ test-integration-local:: $(REPORT_DIR) ## Start local aura infra, run integratio
 
 # --- STDIO integration tests (no Docker infra needed; local child process) ---
 
-.PHONY:test-integration-stdio-local
-test-integration-stdio-local: $(REPORT_DIR) ## Run STDIO MCP integration tests locally
+.PHONY:test-stdio-local
+test-stdio-local: $(REPORT_DIR) ## Run STDIO MCP integration tests locally
 	@echo "Running STDIO MCP integration tests..."
 	@cargo test --package aura --features integration-stdio --no-fail-fast -- --test-threads=1
 
--- a/.makefiles/rust.mk
+++ b/.makefiles/rust.mk
@@ -1,6 +1,7 @@
 lint:: lint-rust fmt-rust
 clean:: clean-rust clean-node clean-report
 build:: build-rust
+test:: test-rust
 
 CARGO_BIN_DIR ?= .bin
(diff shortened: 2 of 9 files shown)


charlesjohnson:
All based on issues that were surfaced by the docs formerly claiming untrue-but-should-be-true things in #334, so first I fixed the docs there, and now this PR changes behavior to match former doc assertions.

Shearerbeard:
This is a great find and a great start but I'm already in the middle of cleaning up docker and CI with the caching so its not going to be worth our time to take this and just run into conflicts. Is there a bug ticket I can reference in my work?

charlesjohnson:
Filed #340 with all the findings as a checklist you can pull from — each item links back to the diff here as a reference implementation, and it notes the CI-lint gap (clippy over `default-members` without `--workspace` never lints standalone aura-cli today). Feel free to fold whatever's useful into the Docker/CI/caching cleanup.

Issue #321, opened 2026-07-09
[FEATURE]: Test and deps buildtime reduction


Shearerbeard:
### Summary

Build times are high in CI and its been a long time since we've audited test performance, unused deps, or build perf optimizations. This is a blanked ticket to catch in repo work to help reduce build times and deps bundle sizes.

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #343, opened 2026-07-11
chore(deps): replace native-tls with rustls across the workspace

### LINKED ISSUE #342: Drop the native-tls/openssl TLS stack from release builds
reqwest pulls native-tls by default, so Linux arm64 release builds cross-compile openssl-sys, adding slow C compilation to the artifact critical path. aura-cli already avoids it via rustls; the rest of the workspace does not.

Ref: #321

Shearerbeard:
## What

rig-core's default feature enabled reqwest/default, which links
native-tls (openssl-sys on linux) into the build. This points rig-core
at its reqwest-rustls feature and unifies the workspace reqwest on
rustls-tls, so the arm64 cross-compile no longer builds openssl in C.

Only the TLS backend changes (native to rustls-tls, bundled webpki
roots). http2, charset, and system-proxy stay enabled.

Fixes #342


Shearerbeard:
@jakedipity @justintime4tea - this is part of our ci build reduction (cuts out the C build). Tagging you so you have an idea of what I'm doing - this will roll up into the larger #321 effort

justintime4tea (approved):

PR #349, opened 2026-07-14
chore(ci): Only run CI for PRs, on main, or when manually triggered

jmoses:
This change updates CI to not run for commits on branches that are not
also PRs. This is to reduce contention of the OSS workers with PRs, as
typically branches without PRs are under heavy development.


Shearerbeard (approved):
this looks right to me

PR #355, opened 2026-07-14
feat(session-store): add storage traits with in-memory default

### LINKED ISSUE #327: Traits + in-memory refactor (no behavior change)
Introduce `SessionStore`, `ApprovalStore`, `EventBus`; move today's structures behind them; A2A already has `TaskStore`. Ship with `backend = memory`. This is a pure refactor — existing tests are the guard.

justintime4tea:
Phase 1 of docs/design/session-storage.md. Today, parked HITL approvals and A2A tasks live in one pod's memory, so an approval decision or task poll that lands on a different pod can't find them. This change introduces the storage traits that will let that state move to a shared backend, and puts the existing in-memory structures behind them. In-memory stays the default, so single-pod servers, CLI standalone, and tests behave exactly as before.

What's new:

- Two capability traits in aura::session_store: ApprovalStore, a durable home for parked approvals, and EventBus, pub/sub used to wake the process that parked a call when its decision arrives. In-memory implementations of both ship as the default backend.
- PendingApprovals rides on those traits: the serializable approval record goes to the store, decisions travel over the bus, and only the non-serializable wake handle stays in-process. With a shared backend, an approval parked on one pod can be resolved from another (covered by a two-registry test).
- The web server builds one session-store backend at startup and wires HITL approvals and the A2A task store from it.

Semantics worth knowing as a reviewer:

- Resolving is persistence-first: a 204 means the decision was recorded and consumed at most once. Delivering the wake to the parked call is best-effort; a lost wake fails closed when the parked call times out.
- Teardown is preserved: request cleanup still cancels parked approvals synchronously (even mid-shutdown), and the store cleanup follows in a background task.
- The new background tasks are instrumented with the current tracing span, so OTel trace parenting is unchanged despite the added async spawns.

Also adds tests around cancellation, expiry, bus fan-out/lag, and the 204 contract. The design doc marks phase 1 implemented and records the open questions this surfaced for the networked backend: fail-fast registration, resolve atomicity across store and bus, teardown latency behind store I/O, trace context over the bus, and metrics for swallowed backend faults.

Fixes: #327

### DIFF
--- a/crates/aura-web-server/src/session_store.rs
+++ b/crates/aura-web-server/src/session_store.rs
@@ -1,0 +1,75 @@
+//! The session-store factory: one backend handing out the capability handles
+//! for cross-pod session state ([`ApprovalStore`] and [`EventBus`] from
+//! `aura::session_store`, plus the upstream `a2a_server::TaskStore`).
+//!
+//! See `docs/design/session-storage.md` and
+//! `docs/adr/2026-07-08-session-storage.md`.
+
+use std::sync::Arc;
+
+use a2a_server::{InMemoryTaskStore, TaskStore};
+use async_trait::async_trait;
+use aura::session_store::{
+    ApprovalStore, EventBus, InMemoryApprovalStore, InMemoryEventBus, SessionStoreError,
+};
+
+/// A pluggable backend for cross-pod session state, handing out one handle
+/// per capability.
+#[async_trait]
+pub trait SessionStore: Send + Sync {
+    /// Durable parked HITL approvals.
+    fn approvals(&self) -> Arc<dyn ApprovalStore>;
+
+    /// Durable A2A tasks (the upstream `a2a_server::TaskStore` trait).
+    fn tasks(&self) -> Arc<dyn TaskStore>;
+
+    /// Cross-pod pub/sub.
+    fn bus(&self) -> Arc<dyn EventBus>;
+
+    /// Cheap liveness check.
+    async fn ping(&self) -> Result<(), SessionStoreError>;
+}
+
+/// The default backend: every capability is process-local, so state is scoped
+/// to one pod.
+pub struct InMemorySessionStore {
+    approvals: Arc<InMemoryApprovalStore>,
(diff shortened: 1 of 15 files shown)


crates/aura-web-server/src/session_store.rs
```diff
@@ -0,0 +1,75 @@
...
+/// per capability.
+#[async_trait]
+pub trait SessionStore: Send + Sync {
+    /// Durable parked HITL approvals.
+    fn approvals(&self) -> Arc<dyn ApprovalStore>;
+
+    /// Durable A2A tasks (the upstream `a2a_server::TaskStore` trait).
+    fn tasks(&self) -> Arc<dyn TaskStore>;
+
+    /// Cross-pod pub/sub.
```
Shearerbeard, line 26: Probably misleading - only cross pos on some stores

crates/aura/src/hitl/registry.rs
```diff
@@ -1,47 +1,77 @@
...
+//! The registry that parks conversational approvals (Route B) and the decision
+//! a later request resolves them with.
 //!
 //! This is the first mutable chat-path state that crosses request boundaries:
 //! an approval is registered during one request's stream and resolved by a
 //! `POST /v1/approvals/{id}` arriving as a different request. It follows the A2A
-//! `SharedTaskStore` idiom — a `Clone` newtype over an `Arc`, constructed once
-//! in `main`, not a global static.
+//! `SharedTaskStore` idiom — a `Clone` newtype over an `Arc` rather than a
+//! global static.
```
Shearerbeard, line 8: This is probably my work but its an unnecessary meta comment

Shearerbeard (approved):
I didn't find anything concerning here - looks good. Approved pending we get a few eyes on manual testing

justintime4tea:
Closing so that the feature/GH-327 can be used as a feature build for SRE testing.

This is new branches PR with feature build:
https://github.com/mezmo/aura/pull/361

PR #359, opened 2026-07-15
feat(telemetry): emit cli_session_ended on repl exit

charlesjohnson:
Adds a `cli_session_ended` PostHog telemetry event with a sealed `ExitReason` property.

The CLI now emits the event for quit, EOF, double interrupt, and input-error exits through the existing telemetry crate. Obeys all existing opt-in/opt-out behavior.

Validated with the full `aura-cli` and `aura-telemetry` test suites, CLI build, formatting, and clippy with the existing `large_enum_variant` baseline allowance.

### DIFF
--- a/crates/aura-cli/src/repl/loop.rs
+++ b/crates/aura-cli/src/repl/loop.rs
@@ -22,6 +22,7 @@ use crate::repl::conversations::ConversationStore;
 use crate::repl::history::ConversationHistory;
 use crate::repl::input_reader::{self, HISTORY_COUNT, HISTORY_DEPTH, LAST_READLINE_INPUT};
 use crate::repl::registry::{self, CommandContext, CommandOutcome};
+use crate::repl::telemetry_notice::{FirstMessageConsent, consent_on_first_message};
 use crate::tools;
 use crate::ui::markdown::{render_markdown, render_summary};
 use crate::ui::prompt::{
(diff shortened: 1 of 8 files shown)


crates/aura-cli/src/repl/loop.rs
```diff
@@ -2020,20 +2021,64 @@ pub fn run_repl(
...
+
+    /// Process the first submitted chat message, returning whether the newly
+    /// enabled preference should be persisted by the caller.
+    fn on_first_message(&mut self) -> bool {
+        if !self.consent_pending {
+            return false;
+        }
+        self.consent_pending = false;
+
+        use crate::repl::telemetry_notice::{FirstMessageConsent, consent_on_first_message};
```
Shearerbeard: Odd inline import - I would prefer this just use a standard top level import or if this is here due to a collision you can do a top level  module import and a qualified path

crates/aura-telemetry/src/events.rs
```diff
@@ -67,3 +69,13 @@ pub struct ChatRequestCompleted {
     /// user; `false` if it errored or was cancelled.
     pub success: bool,
 }
+
+/// Best-effort session-end counterpart emitted on REPL exit while telemetry remains `Enabled`; opt-out or transport failure can break pairing with [`CliSessionStarted`].
+/// **Why we track it / how we use it:** derive session length and turn volume from `session_id` timestamps and classify exit paths without payload counts or durations.
```
Shearerbeard: I don't think this belongs in this spot - maybe its covered in the top level module or docs

crates/aura-telemetry/src/properties.rs
```diff
@@ -89,6 +89,27 @@ impl DeploymentMethod {
...
+impl ExitReason {
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::Quit => "quit",
+            Self::Eof => "eof",
+            Self::Interrupt => "interrupt",
+            Self::Error => "error",
+        }
+    }
+}
```
Shearerbeard, line 111: Only used for printing/serialization or are we matching on strings where enum match statement would be a clearer dispatch?

Shearerbeard (changes requested):
Looking only at the diffs this looks close to ready. I have marked specific areas to handle with a follow up question. Can the tests be made less verbose? Seems like a lot of setup code and maybe you already have some setup this is duplicating.

charlesjohnson:
> Looking only at the diffs this looks close to ready. I have marked specific areas to handle with a follow up question. Can the tests be made less verbose? Seems like a lot of setup code and maybe you already have some setup this is duplicating.

Thanks for taking the time to review! I've pushed another version up to address these.

Shearerbeard (approved):

PR #360, opened 2026-07-15
fix(mcp): normalize bare boolean nodes in sanitized tool schemas

### LINKED ISSUE #358: MCP tool schemas with bare boolean fields break local llama.cpp tool calling
Aura's `sanitize_schemas = true` option is supposed to make MCP tool schemas
broadly compatible across providers, but it doesn't handle schema nodes that
are literally the JSON boolean `true`/`false` (valid JSON Schema shorthand for
"anything goes", which is how Pydantic v2 renders `Dict[str, Any]`-typed
fields). Those pass straight through unchanged.

OpenAI and Anthropic don't mind. llama.cpp's `--jinja` mode does: it builds a
tool-call grammar directly from the schema, and a bare boolean node breaks
that outright:

    400 "Unable to generate parser for this template. ... Unrecognized schema: true"

Anyone running a local llama.cpp-served model with tool calling enabled hits
this the moment an MCP tool has a loosely-typed (`Dict[str, Any]`-style)
parameter. Cloud providers are unaffected since they don't build a grammar
from the raw schema.

Minimal repro against any llama.cpp `--jinja` server:

    {"type": "object", "properties": {"extra": true}}   # 400s
    {"type": "object", "properties": {"extra": {}}}       # works

Same meaning, different shape - one breaks the grammar generator, one doesn't.

This lives in the sanitizer path in `crates/aura/src/schema_sanitize.rs` -
happy to send a PR.

Shearerbeard:
MCP tool schemas can render Dict[str, Any]-typed fields as the bare JSON Schema boolean `true` (valid shorthand for "any value", how Pydantic v2 emits it). Add a pass that normalizes bare true/false schema nodes to `{}` / `{"not": {}}` before the other sanitization passes run, so they only ever see object schemas.

Fixes: #358

### DIFF
--- a/crates/aura/src/mcp.rs
+++ b/crates/aura/src/mcp.rs
@@ -792,7 +792,8 @@ impl McpManager {
     /// Returns Ok(()) if successful, or Err(reason) if schema is invalid and should be rejected.
     fn sanitize_schema_for_openai(schema: &mut serde_json::Value) -> Result<(), String> {
         use crate::schema_sanitize::{
-            fix_empty_root_required, recursive_set_additional_properties_false,
+            fix_empty_root_required, normalize_bare_boolean_schemas,
+            recursive_set_additional_properties_false,
         };
 
         // VALIDATION: OpenAI requires tool schemas to have type: "object" at root level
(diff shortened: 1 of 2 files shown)


jmoses (commented):
If we're using an existing library to handle the JSON schemas, it might be worth an upstream ticket. I did no research into what we're using.

Shearerbeard:
> If we're using an existing library to handle the JSON schemas, it might be worth an upstream ticket. I did no research into what we're using.

Not really a jsonschema issue - lots of tweaky tool schema compatibility issues across providers. Whats being emitted (in this case from our MCP server) is valid - we just have extra processing for quirks based on the templates different models use.

justintime4tea (approved):

PR #376, opened 2026-07-16
feat(install): prefer homebrew on macos with direct-download fallback

charlesjohnson:
On macOS the install script now installs via the `mezmo/tap/aura` Homebrew tap when `brew` is available. It falls back to the existing direct GitHub Release download when brew is missing, when the new `AURA_NO_BREW=1` opt-out is set, or when `AURA_VERSION`/`AURA_COMPONENT` request something the formula cannot honor (in which case it prints why and downloads directly). Linux is unaffected.

Ref: #377


Shearerbeard (dismissed):

charlesjohnson:
Depends on #348

Shearerbeard (approved):

PR #380, opened 2026-07-16
feat(orchestration): nudge agents approaching the turn depth limit

### LINKED ISSUE #379: [FEATURE]: Nudge agents approaching the turn depth limit to submit partial results
### Summary

When a worker task hits the turn depth limit mid-investigation, rig raises `MaxDepthError`, the task fails, and everything the worker gathered is thrown away. The coordinator then retasks the worker, which starts over from zero. Thorough investigation prompts hit this routinely.

The worker never learns it is running out of turns, so it has no chance to cut its losses and submit partial findings. A partial `submit_result` is almost always more useful than a failed task plus a full restart.

### Proposed solution

Warn the agent inside the conversation as it approaches the limit, using two independent `[agent]` flags:

```toml
[agent]
nudge_last_turn = true       # on the final turn, instruct the agent to submit results now
nudge_turns_remaining = 3    # start wrap-up notices when 3 or fewer turns remain
```

The nudge text rides on MCP tool output (a `ToolWrapper` appends it), since rig feeds tool results back as the next turn's prompt. Turn position comes from counting per-turn `StreamItem::TurnUsage` events, so no rig fork changes are needed. Workers get `submit_result`-specific wording; single-agent mode says to answer in text.

This is a prototype to gauge whether nudging reduces task failures and coordinator retasking.

### Related code

- `crates/aura/src/turn_nudge.rs` (new)
- `crates/aura/src/orchestration/orchestrator.rs` (`create_worker`)
- `crates/aura/src/builder.rs` (`Agent::new`, stream methods)
- `crates/aura-config/src/config.rs` (`[agent]` flags)

teriyakichild:
Fixes #379

## Problem

A worker that hits the turn depth limit mid-investigation fails with `MaxDepthError` and loses everything it gathered. The coordinator retasks it and the investigation starts over from zero. Prompts that ask for thorough investigation hit this routinely.

## Change

Warn the agent in-conversation as it approaches the limit, so it can wrap up and submit partial results instead of losing the task. Two independent `[agent]` flags:

```toml
[agent]
turn_depth = 10
nudge_last_turn = true       # on the final turn: submit results now
nudge_turns_remaining = 3    # start wrap-up notices when 3 or fewer turns remain
```

Workers inherit the flags from `[agent]` and get wording that tells them to call `submit_result`. Single-agent mode says to answer in text. The coordinator is excluded. A worker that calls `submit_result` on its literal last turn still succeeds, since the orchestrator already treats `MaxDepthError` as success when a decision was captured.

## How it works

Rig's multi-turn loop lives in the fork, so the nudge rides on two aura-owned seams instead:

1. `Agent::count_turns` tracks the current turn by counting the per-turn `StreamItem::TurnUsage` events on the stream.
2. A new `TurnNudgeWrapper` (`ToolWrapper`) appends the notice to MCP tool output. Rig feeds tool results back as the next turn's prompt, so the model reads the notice at the start of its next turn.

The wrapper runs last in the composed chain, after scratchpad pointer rewriting, so the notice lands on the text the LLM sees and is never persisted as raw tool output. At most one nudge is emitted per turn. The turn math accounts for rig 0.28 executing up to `max_depth + 2` turns before `MaxDepthError` (its loop breaks on `counter > max_depth + 1`).

## Scope and known limits

This is a prototype to gauge impact on task failure and retask rates.

- Only MCP tool outputs carry the nudge; native tools (scratchpad reads, `read_artifact`, `submit_result`) are not wrapped by `tool_wrapper`. A turn that calls no MCP tool gets no nudge that turn.
- Ollama fallback tool parsing executes outside rig's loop and is not nudged.
- Nudge state is per `Agent`. All current callers build agents per request or per task, so concurrent streams on one agent do not occur today.

## Testing

- 6 unit tests in `turn_nudge.rs` covering the turn math, once-per-turn dedup, threshold behavior, worker vs single-agent wording, and counter reset between streams
- Parse test in `aura-config` for the new flags and their defaults
- `cargo test --workspace` passes, clippy clean

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,6 +337,8 @@ Tool names are not namespaced by server. If two servers register a tool with the
 
 `turn_depth` controls how many tool-calling rounds can happen in a single turn. Higher values allow multi-step tool workflows before final response generation. This acts as a failsafe to prevent models from spinning out in unbounded tool-call loops.
 
+`nudge_last_turn` and `nudge_turns_remaining` enable turn-limit nudging: as the agent approaches the `turn_depth` limit, a warning is appended to tool output telling it to wrap up and deliver its answer (orchestration workers are told to call `submit_result`) rather than lose all gathered work when the limit terminates the run. `nudge_last_turn = true` warns on the final turn; `nudge_turns_remaining = N` starts wrap-up warnings when N or fewer turns remain. Both default to off and can be enabled independently.
+
 `context_window` sets the context window size (in tokens) for the agent, used for usage percentage reporting in `aura.session_info` streaming events.
 
 The complete starter configuration is in [examples/reference.toml](examples/reference.toml). Minimal per-provider configs are in `examples/minimal/` and complete agent examples are in `examples/complete/`.
--- a/crates/aura-config/src/config.rs
+++ b/crates/aura-config/src/config.rs
@@ -757,6 +757,15 @@ pub struct AgentConfig {
     /// `[orchestration.worker.<name>.skills]`.
     #[serde(default)]
     pub skills: SkillsConfig,
+    /// Nudge the agent on its final turn to submit its results instead of
+    /// calling more tools (default: false).
+    #[serde(default, deserialize_with = "lenient_bool::deserialize_bool")]
+    pub nudge_last_turn: bool,
+    /// Nudge the agent to start wrapping up when this many turns remain
+    /// before the turn-depth limit. `None` disables wrap-up nudging.
+    #[serde(default)]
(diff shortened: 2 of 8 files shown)


Shearerbeard (dismissed):

Shearerbeard (changes requested):
Looks good - just needs a docs added to the reference.toml or somewhere else that makes sense

teriyakichild:
> Looks good - just needs a docs added to the reference.toml or somewhere else that makes sense

@Shearerbeard done

Shearerbeard (approved):

PR #393, opened 2026-07-22
feat(session-store): redis-backed A2A and HITL

### LINKED ISSUE #328: Redis/Valkey backend — A2A task store
Smallest, highest-value cross-pod win (`message:send` + poll works across pods). No bus needed for the poll path.

### LINKED ISSUE #329: Redis/Valkey backend — HITL approvals (store + wake bus)
Cross-pod conversational approvals via `approval:{id}` pub/sub. Fail-closed preserved.

### LINKED ISSUE #330: A2A streaming/cancel over the bus
Cross-pod `message:stream` / `subscribe` / `cancel` via the `a2a:task:{id}` (fan-out) and `a2a:cancel:{id}` (routed cancel) topics. Depends on the shared task store (phase 2) and the `EventBus` impl (phase 3).

justintime4tea:
# Redis/Valkey session store: cross-instance A2A and HITL

## What this adds

AURA behind a load balancer previously required session affinity: A2A tasks,
parked HITL approvals, and live streams existed only in the memory of the
instance that created them. This PR adds an optional Redis/Valkey session-store
backend that makes every cross-request flow work regardless of which instance
serves the request:

- **A2A tasks** — `message:send` → poll → `get` → `list` → history-by-context
  all read/write one shared, TTL'd task store. Terminal task states are
  immutable once recorded.
- **Conversational HITL approvals** — an approval parked on one instance can be
  resolved by a `POST /v1/approvals/{id}` landing on any other; the decision
  travels over pub/sub and wakes the parked call. Fail-closed behavior is
  unchanged: any lost wake ends in a denied timeout, never an approval.
- **A2A streaming and cancel** — `tasks/{id}:subscribe` attaches to a task
  executing on another instance (snapshot + live event relay), and
  `tasks/{id}:cancel` reaches whichever instance is executing. Relays detect
  dropped frames and surface an explicit resubscribe error rather than
  silently delivering partial content, and are lifetime-bounded so a dead
  executing instance cannot hang subscribers.

With no configuration, nothing changes: the in-memory backend remains the
default and single-instance behavior is identical.

## Configuration

Environment variables only — deliberately no TOML surface, because agent TOML
files are per-agent (a server loads many) while the session store is
deployment infrastructure with exactly one instance per server:

```bash
AURA_SESSION_STORE=redis                   # memory (default) | redis
AURA_SESSION_STORE_URL=redis://valkey:6379 # required for redis; Valkey compatible
AURA_SESSION_STORE_PREFIX=aura:prod        # optional key/topic namespace
AURA_SESSION_STORE_CONNECT_TIMEOUT_SECS=5  # optional
AURA_SESSION_STORE_TASK_TTL_SECS=86400     # optional; 0 = no expiry
```

The server fails fast at startup if the configured backend is unreachable, and
`/health` reports the backend with its ping latency. The Redis client is gated
behind the `session-store-redis` cargo feature; the in-memory backend always
compiles.

## Scope and known limitations

- Redis pub/sub is fire-and-forget by design; every delivery path is
  backstopped (approval timeout, relay store-poll + lifetime bound, store as
  source of truth). The design doc records each accepted window.
- Single-instance/Sentinel Redis topologies; Redis Cluster is out of scope.
- Helm/Valkey packaging is a planned follow-up (design doc §10–11).

## Reviewing

`docs/design/session-storage.md` is the companion design note (architecture,
key schema, trade-off decisions, phased rollout). Commit messages describe each
phase. Unit tests cover the components; the `integration-session-store` suite
(28 tests, `make test-integration-session-store-local`) exercises everything
against a live Valkey, including two-instance simulations of every cross-pod
flow.

Fixes: #328
Fixes: #329
Fixes: #330


crates/aura-web-server/src/session_store/redis.rs
```diff
@@ -0,0 +1,111 @@
+//! Redis/Valkey session-store backend.
+//!
+//! The A2A task store ([`task_store`]), the HITL approval store
+//! ([`approval_store`]), and the event bus ([`event_bus`]) are all
+//! Redis-backed, so A2A send/poll/list, conversational approvals, and —
+//! through `crate::a2a`'s bus bridge — A2A streaming/subscribe/cancel all
+//! work across instances. Each submodule documents its own key schema under
+//! the configured `key_prefix`. See `docs/design/session-storage.md`.
+
```
Shearerbeard, line 9: I haven't been the best at this out of habbit but we should probably start making sure new modules use the <module_name>.rs pattern instead of mod.rs

Shearerbeard (changes requested):
I don't think what i found testing this is actually in the diff but there is the possibility of racing the webhook approval before the record gets written to redis. 

```
approval_event_broker::publish(
                    &request_id,
                    ApprovalLifecycleEvent::Requested((&request).into()),
                )
                .await; <--- published

                let expires_at = chrono::Utc::now()
                    + chrono::Duration::from_std(*timeout)
                        .expect("approval timeout fits in chrono");
                let pending_event = events::pending(&request, &expires_at);
                let handle = registry.register(request, *timeout).await; <--- now in redis
```

This should be audited to make sure were always writing the record before any sort of external dispatch - likely something I missed in the prep pr. This is unlikely to block now unless something on the other end is programmatically approving but its easy enough to solve now

Shearerbeard (approved):

Issue #310, opened 2026-07-06
[EPIC] Blessed MCP Install Helper


charlesjohnson:
Stub, full body to follow.

A supported helper that connects a fresh AURA install to real data sources by installing and configuring MCP servers. Serves initiative I-1 (five minute install) and is the load-bearing dependency for I-2's documentation-led integration strategy.

Shearerbeard:
PR #381 (#373) brings forward some plumbing this epic can build on: cold MCP detection and a credential-free mcp_servers wire format on /aura/info. Live connection status and tool counts are not in that view. They would load from a second source via the runtime aura.info SSE.

PR #404, opened 2026-07-23
feat: guided mcp server install flow (/mcp)

### LINKED ISSUE #378: [FEATURE]: Empty worker mcp_filter grants access to all MCP tools
### Summary

Worker MCP access currently has two opposite meanings depending on whether `mcp_filter` contains any entries:

A non-empty `mcp_filter` acts as an allowlist. The worker receives only matching MCP tools.
An empty or omitted `mcp_filter` grants the worker access to every available MCP tool.

This creates a dangerous authorization boundary where removing the final allowlist entry (or failing to generate one) expands a worker’s permissions from limited access to unrestricted access.

This is especially risky for onboarding and MCP installation workflows. Connecting a new MCP server at the global level may unintentionally expose it to workers whose filters were omitted or left empty.

Example
```
# Limited to matching tools
[orchestration.worker.log-analyst]
mcp_filter = ["logs_*"]

# Receives every MCP tool
[orchestration.worker.writer]
mcp_filter = []

# Also receives every MCP tool
[orchestration.worker.researcher]
# mcp_filter omitted
```
The transition from ["logs_*"] to [] widens access rather than removing it.

###Security impact

A configuration omission or apparently restrictive edit can increase a worker’s privileges without an explicit allow-all declaration. This violates least-privilege expectations and makes accidental MCP tool exposure likely as configurations and available integrations change.

### Additional Context

_No response_

### Upload screenshots

_No response_

### Searched Issues

- [x] No similar issues found

### Code of Conduct

### LINKED ISSUE #402: Format-preserving writer for [mcp.servers] entries in agent config
### Summary

The MCP install helper (#310) needs to add and update `[mcp.servers.<name>]` entries in a user's `config.toml` "without destroying unrelated settings" (epic acceptance criteria). Today the agent config is read-only: `aura-config` only parses, and the format-preserving `toml_edit` machinery in `aura-cli/src/config.rs` is scoped to the CLI's own `cli.toml`.

This issue adds the reusable building block: a writer in `aura-config` that upserts a single MCP server entry into an existing config file.

### Scope

- `aura_config::writer::upsert_mcp_server(path, name, &McpServerConfig)` plus a string-level variant for testability.
- Format-preserving via `toml_edit`: unrelated tables, sibling servers, comments, and value decor survive the edit.
- Atomic write (temp file + rename in the same directory), matching the existing `cli.toml` save path.
- `{{ env.VAR }}` placeholder strings pass through verbatim, so callers can keep secrets out of the TOML.
- Creates `[mcp]` / `[mcp.servers]` as explicit tables when missing; replaces an existing server table wholesale on name collision.
- Errors (rather than clobbering) on malformed TOML or a missing config file.
- Written output must round-trip through the normal loader back into an equal `McpServerConfig` for all three transports (`stdio`, `http_streamable`, `sse`).

### Non-goals

- No user-facing command (the guided `/mcp` flow builds on this separately).
- No remove/rename/list operations.
- No change to worker `mcp_filter` semantics (#378).

### LINKED ISSUE #403: Guided /mcp command: add and verify MCP servers from the CLI
### Summary

With the config writer (#402) in place, this issue adds the user-facing half of the MCP install helper (#310): a guided `/mcp` command in the CLI's standalone mode that takes a user from "AURA is installed" to "AURA can successfully use tools against my data" without hand-authoring TOML.

### Scope

- `/mcp` — list the configured MCP servers for the active config with their transport and, when available, live connection status.
- `/mcp add` — a guided, deterministic wizard (no LLM in the loop for credentials):
  - Pick from a small curated catalog of blessed servers (each with prerequisites shown up front and the access AURA will require) or define a custom server (transport, url/cmd, headers).
  - Credentials are collected with masked input and stored as `{{ env.VAR }}` placeholders in the TOML; the secret value lands in the project `.env`, never in `config.toml`.
  - Config is updated in place via `aura_config::writer::upsert_mcp_server`, preserving unrelated settings.
  - An automatic connection and tool-discovery check runs against the newly added server and reports connected/failed status plus tool count.
  - On success, print a starter prompt that exercises the new integration.

### Non-goals

- HTTP mode (`--api-url`): the config lives server-side; a mutation endpoint is a separate discussion.
- Editing or removing existing servers (`/mcp remove` etc.).
- Changing worker `mcp_filter` semantics (#378).

teriyakichild:
## What

The first slice of the MCP install helper epic (#310): a guided `/mcp` flow in the CLI that takes a user from "AURA is installed" to "AURA can use tools against my data" without hand-authoring TOML — plus the config-writer foundation it stands on and the `mcp_filter` semantics fix that makes worker scoping expressible.

Closes #402
Closes #403
Closes #378

## The flow

`/mcp` lists the active agent's MCP servers (name, transport, target, description) from the credential-free `AgentInfo.mcp_servers` overview, so it behaves identically in standalone and HTTP modes.

`/mcp add` (standalone mode) is a deterministic wizard — no LLM ever sees a credential:

1. Pick from a curated catalog (Mezmo, PagerDuty, Datadog, Kubernetes with `--read-only`) or define a custom server (transport, URL/command, auth header with scheme prefix, e.g. `Authorization = "Bearer {{ env.MEZMO_API_KEY }}"`). Prerequisites and the access AURA will get are shown before any input.
2. Credentials come from an env var you already export (auto-offered when the conventional var is set, or name your own) or masked entry — only entered values are saved, to `.env` (0600 on create, values quoted for dotenvy), never to the TOML.
3. The server is verified **before** the config-write prompt: a bounded (30s) in-memory connection + tool-discovery check. Consent is informed by the live result; a failed check flips the prompt to write-anyway-defaulting-to-no.
4. Config updates go through a new format-preserving `toml_edit` writer (`aura_config::writer`): atomic temp+rename, file permissions carried over, comments and unrelated settings preserved, `{{ env.VAR }}` placeholders untouched.
5. For orchestrated configs, a per-worker access step uses the discovered tool names: filterless workers get a lockdown choice (scope to this server / no MCP tools / keep all), allowlisted workers get an append-style grant (namespace glob like `mezmo_*` when safe, exact names otherwise — never a generic-verb glob that would match other servers' tools).
6. On success, a starter prompt that exercises the new integration.

## Breaking change

`mcp_filter = []` now grants **no** MCP tools instead of every tool (an omitted filter still grants all). This applies to worker filters and the single-agent `[agent].mcp_filter`, makes "no tools" expressible (the wizard's lockdown option writes it), and closes the trap where deleting the last allowlist entry silently widened access to everything. No shipped config regresses; the `["__none__"]` workaround in quickstart is replaced. Migration notes: `docs/breaking-changes/20260723-empty-mcp-filter.md`. `client_tool_filter` is unchanged.

## Not in scope

- HTTP-mode `/mcp add` (server-side config mutation needs an auth + persistence story; the bootstrap agent in #391 is the likely path)
- Hot-reload — the wizard says "restart aura to activate"; `/mcp` list reflects the running roster until then
- Config-directory setups and `/mcp remove`

## Testing

- 39 workspace test suites green; clippy and fmt clean in default and `--no-default-features` (HTTP-only) builds
- New unit coverage: writer round-trips for all three transports (format/comment/permission preservation, inline-table conversion, atomicity), `.env` quoting round-tripped through dotenvy parsing, catalog templates render-and-reload with placeholders intact, secret inlining, worker filter parsing (`[]` vs omitted), glob-safety rules
- Manually exercised end to end against a local config (catalog + custom flows, worker scoping) and a local aura-web-server deployment (`/mcp` list via `/aura/info`, `add` correctly refusing in HTTP mode)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -308,6 +308,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.18",
  "toml 0.8.23",
+ "toml_edit 0.22.27",
  "tracing",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 serde_yaml = "0.9"
 toml = "0.8"
+toml_edit = { version = "0.22", features = ["serde"] }
 
 # Async runtime
 tokio = { version = "1.45", features = ["full", "test-util"] }
--- a/README.md
+++ b/README.md
@@ -480,7 +480,7 @@ For a fuller multi-worker example, see [configs/example-math-orchestration.toml]
 |-------|------|---------|-------------|
 | `description` | string | *required* | Short description shown to coordinator during planning |
 | `preamble` | string | *required* | System prompt for this worker |
-| `mcp_filter` | list | `[]` | Glob patterns selecting which MCP tools this worker can use |
+| `mcp_filter` | list | — | Glob patterns selecting which MCP tools this worker can use; omitted = all tools, `[]` = none |
 | `vector_stores` | list | `[]` | Named vector stores this worker has access to |
 | `turn_depth` | int | — | Per-worker tool-call depth limit (overrides `[agent].turn_depth`) |
 | `llm` | table | inherits `[agent.llm]` | Optional per-worker LLM override — different model (and other `[agent.llm]` fields) while reusing provider credentials |
--- a/crates/aura-cli/Cargo.toml
+++ b/crates/aura-cli/Cargo.toml
@@ -25,9 +25,7 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 toml = { workspace = true }
 # Format-preserving TOML editor for in-place [telemetry] cli.toml edits.
-# Pinned to the 0.22 line that `toml` 0.8 already builds on, so it adds
-# no new transitive tree.
-toml_edit = "0.22"
+toml_edit = { workspace = true }
 anyhow = { workspace = true }
 uuid = { workspace = true }
 regex = { workspace = true }
--- a/crates/aura-cli/README.md
+++ b/crates/aura-cli/README.md
@@ -274,6 +274,7 @@ Once inside the REPL, the slash commands below are available. All slash commands
 | `/rename <name>`   | Rename the current conversation                                     |
 | `/model <filter>`  | Browse and select a model (see [Model Selection](#model-selection)) |
 | `/style [name]`    | Switch visual style: `normal`, `high-contrast`, `no-colors`         |
+| `/mcp [add]`       | List the agent's MCP servers, or install one via a guided flow      |
 | `/quit` or `/exit` | Exit the REPL                                                       |
 
 ---
--- a/crates/aura-cli/src/backend/direct.rs
+++ b/crates/aura-cli/src/backend/direct.rs
@@ -119,11 +119,9 @@ impl DirectBackend {
         })
     }
 
-    /// Path the agent configs were loaded from, as given at startup.
-    ///
-    /// `load_config` accepts both a single TOML file and a directory of
-    /// them, so callers that want to *write* config (e.g. `/mcp add`) must
-    /// check `is_file()` before treating this as an editable target.
+    /// Path the agent configs were loaded from, as given at startup. May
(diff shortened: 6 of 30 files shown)


Shearerbeard (changes requested):
Lets get rid of the unnecessary comments and simplify

Shearerbeard (approved):

Issue #454, opened 2026-07-27
provider-unavailable errors are never retried; one 429 kills a worker task


Shearerbeard:
## Observed behavior

A worker hit a single 429 mid-task. The task failed immediately, the coordinator burned a
replan cycle reacting to it, and the replacement task re-ran every tool call from scratch.
When a whole wave hits the same rate limit, the run either aborts outright (all failures
provider-flavored, nothing completed) or grinds through replan cycles that face the same
throttled provider with no delay.

## What the code does today

- The worker attempt loop treats every `Err` as hard: `last_error = Some(e); break`
  (`orchestrator.rs:3173-3177`). The second of `MAX_WORKER_ATTEMPTS = 2` only ever runs
  for the missing-`submit_result` soft failure. A 429, 503, or dropped connection is
  final on the first occurrence.
- The planning loop does retry transient errors (`ProviderOverloaded`, `AgentTimeout`) but
  with zero delay between attempts, so it hammers a throttled provider and burns its
  `max_plan_parse_retries` budget in milliseconds (`orchestrator.rs:1480-1499`). Each
  failed attempt also leaves an unpaired user message in the coordinator conversation.
- Classification is string matching over rig's flattened `ProviderError(String)`
  (`orchestrator.rs:3339-3387`). It misses "overloaded" (Anthropic 529), plain 500, and
  every connection-failure shape ("connection refused", "reset by peer", dns errors), all
  of which land in `AgentError` and get treated as replannable agent mistakes. The
  `"timed out"` branch is checked first, so a provider error whose text mentions a timeout
  is misfiled as `AgentTimeout` and evades the provider short-circuit
  (`orchestrator.rs:3391-3408`).
- A related classifier already exists and is unused: `RetryHint`
  (`tool_call_observer.rs:153-219`) is computed for MCP tool errors and then dropped when
  events convert (`orchestrator.rs:237-254`). It is not a drop-in for provider errors.

## Why this needs more than a bugfix

The natural fix is what other agent frameworks do: retry the failed provider call with
exponential backoff and jitter, at the level of a single turn. AURA cannot do that from
outside rig. In the fork, the per-turn stream open propagates failure with `?` and kills
the whole multi-turn stream (`rig streaming.rs:275-282`), and the chat history lives in an
`Arc<RwLock<...>>` local to the generator that is only surfaced on `MaxDepthError` and
`PromptCancelled`, not on completion errors. The only retry unit reachable from AURA is
[...]

Issue #91, opened 2026-05-08
Support OpenAI Responses API


andrewelizondo:
We should support both but prefer responses when available for performance. 

https://developers.openai.com/api/docs/guides/migrate-to-responses

Issue #62, opened 2026-04-22
first_chunk_timeout fires during silent reasoning on OpenAI Chat Completions


teriyakichild:
## Summary

The `FIRST_CHUNK_TIMEOUT_SECS` safety net (default 30s) can fire during healthy reasoning-model turns because OpenAI's Chat Completions API streams **nothing** on the wire while the model is thinking. This is not a hung provider; it's the expected shape of a reasoning turn on that endpoint. Today aura cancels the request anyway.

## Context

This surfaced while testing a legitimate agent pattern: a system prompt that instructs the agent to **relay its final output via a tool call rather than as natural-language output tokens**. This pattern is common when wiring agents to downstream systems (e.g. `submit_result(text)`, `final_answer(...)`, or any tool whose argument carries the analysis payload). Under this shape, the model:

1. Does all of its work during the reasoning phase.
2. Emits no natural-language text at all.
3. Concludes the turn with a single tool_call whose arguments contain the entire answer.

On OpenAI Chat Completions this means the SSE stream is genuinely silent from request start until the tool_call deltas begin arriving. Even with `reasoning_effort = "medium"` on `gpt-5.2`, the reasoning-before-tool-call gap easily exceeds the default 30s for non-trivial prompts.

## Reproduction

1. Agent config with `model = "gpt-5.2"` and (optionally) `reasoning_effort = "medium"` or higher.
2. System prompt forcing the tool-only output shape (no natural-language text), e.g. "emit your final answer ONLY by calling tool X with your analysis in an argument."
3. Prompt the agent with something that requires substantive reasoning, e.g. "tell me a story."
4. Observe in `docker logs aura-test`:
   ```
   Current conversation depth: 1/20
   # ... ~30s later ...
   First chunk timeout (30s) - no response from LLM provider
   Cancelling request '...': timeout
   ```
   The model was not hung — aura just never saw a stream item.

## Root Cause

- `crates/aura/src/builder.rs:183` explicitly forces OpenAI's Chat Completions endpoint via `client.completions_api()` (over rig's default Responses API).
- Chat Completions **does not stream reasoning tokens or summaries** — during a reasoning phase the SSE stream is silent until the first content token or tool_call delta.
- `crates/aura-web-server/src/streaming/handlers.rs:157-160` clears `first_chunk_received` on any stream item, which works fine for non-reasoning models but never fires during a reasoning-only warmup.
[...]

PR #458, opened 2026-07-28
Add inactivity timeouts for silent provider streams

### LINKED ISSUE #394: no inactivity timeout for provider hangs; per_call_timeout_secs bounds a worker's whole task
## Observed behavior

An orchestration worker ran several tool calls, then the provider hung while opening the stream for a later turn. Nothing detected the hang. The task sat silent until the cumulative `per_call_timeout_secs` budget (300s in this config) ran out, at which point the whole task failed.

### 1. `per_call_timeout_secs` is a whole-task timeout for workers, not a per-call one

In `stream_and_forward` (`crates/aura/src/orchestration/orchestrator.rs`), the `tokio::time::timeout` wraps the entire `stream_future`, which contains the full `agent.stream_chat()` ReAct loop: every LLM turn and every tool execution up to the worker's `max_depth`. A worker that does ten productive turns draws down the same budget as one that hangs on its first turn.

For coordinator phases (planning, continuation routing) the name is roughly accurate, since `stream_and_collect` uses `max_depth=1` and early-exits after a one-shot tool decision. For workers, "per call" means "per invocation of `stream_and_forward`", which is the worker's entire task.

This creates a bad tradeoff: the timeout cannot distinguish a hung provider from a busy worker. Raising the value to accommodate long tasks makes hang detection slower; lowering it kills legitimate long tasks while they are actively making progress. The doc comment on `TimeoutsConfig` ("Prevents a single hung LLM call from blocking...") describes coordinator behavior, not worker behavior.

### 2. No inactivity detection anywhere after the first stream item

The web server's `FIRST_CHUNK_TIMEOUT_SECS` (default 90s) only fires before the first item on the SSE stream. `first_chunk_received` is set once in the streaming select loop (`crates/aura-web-server/src/streaming/handlers.rs`) and never reset, so the timer is permanently disarmed after the first event. Since session-info, tool, and reasoning events all count as items, a request that has run any tools no longer has stall protection.

After that point the only backstops are:

- `STREAMING_TIMEOUT_SECS` (default 900s), which bounds the whole request regardless of activity
- the heartbeat, which detects client disconnects but not a stalled provider

So a provider hang on turn N of a single-agent request, or inside a worker's ReAct loop, goes undetected until a coarse whole-request or whole-task timeout fires.

## Proposal

Add an inactivity timeout: a deadline that re-arms on every stream item (text delta, tool call, tool result, reasoning delta all count as liveness) and fails the stream only if it goes silent for N seconds. Apply it in both places:

1. Inside `stream_and_forward`'s item loop, so a hung worker turn fails fast without penalizing long multi-turn tasks. Config could live next to `per_call_timeout_secs` as something like `inactivity_timeout_secs`.
2. In the web server's streaming select loop, re-arming the existing `first_chunk_sleep` (or a new sleep) on each `stream.next()` item, closing the same gap for single-agent mode.

Tool executions that legitimately take longer than the inactivity window need consideration. MCP progress notifications could count as liveness, or the window could be suspended while a tool is executing, since tool hangs are a separate concern from provider hangs.

Separately, the `TimeoutsConfig` doc comment should say what the timeout actually bounds for workers.

Shearerbeard:
Add an inactivity deadline that fires when the LLM provider goes silent between stream chunks, even while the connection stays open. Today nothing watches for that: the first-chunk timer disarms after the first stream item, and the per-call budget is whole-task. A stall ran until the 900s `STREAMING_TIMEOUT_SECS` kill.

The new `InactivityDeadline` (crates/aura/src/inactivity.rs) re-arms on each stream item and suspends during tool execution, so slow MCP tool calls and paused HITL approvals won't trip it. The orchestrator loops (coordinator and worker) and the web server's SSE select loop share it.

Config:
- `[orchestration.timeouts].inactivity_timeout_secs` for orchestration
- `--inactivity-timeout-secs` / `INACTIVITY_TIMEOUT_SECS` for the server

Both default to 0 (disabled).

When the orchestrator hits the deadline, that task fails with a "no stream progress" error and the run completes with partial results instead of dying mid-wave.

Tested against a live provider that stalls mid-stream. Disabled, the request hangs to the 900s kill. Enabled, single-agent requests fail at the window; orchestrated workers lose the stalled task and the run returns partial results.

Orchestrated worker activity reaches the server layer as tool and progress events rather than stream items, which re-arm the deadline but can't suspend it. A worker tool call that goes quiet past the window can still trip the server-layer deadline. The TOML knob suspends during tool execution and is the reliable one for orchestrated deployments.

First-party OpenAI reasoning models have a blind spot of their own. Aura pins OpenAI to Chat Completions, which streams nothing while the model thinks, so a long think reads as inactivity (the same failure mode as #62 on the first-chunk timer). Anthropic, Gemini, and OpenAI-compatible `reasoning_content` streams all re-arm while thinking, so this is specific to first-party OpenAI. Size the window above worst-case think time for those models, or leave the knob off. The real fix is reasoning summaries over the Responses API: #91. Retry and backoff for provider errors is #454.

Fixes: #394


crates/aura-config/src/config.rs
```diff
@@ -468,6 +480,35 @@ impl Config {
...
+}
+
+/// Warn when the inactivity window cannot fire before the per-call budget.
+fn inactivity_vs_per_call_warning(
+    per_call_timeout_secs: u64,
+    inactivity_timeout_secs: u64,
+) -> Option<String> {
+    if per_call_timeout_secs > 0 && inactivity_timeout_secs >= per_call_timeout_secs {
+        Some(format!(
+            "inactivity_timeout_secs ({inactivity_timeout_secs}s) is greater than or equal to per_call_timeout_secs ({per_call_timeout_secs}s); the per-call budget always fires first and the inactivity window is dead config"
```
jakedipity: Are these typos? Not sure what the block in the parenthesis represents.

```
inactivity_timeout_secs ({inactivity_timeout_secs}s)
per_call_timeout_secs ({per_call_timeout_secs}s)
```

Also `is dead config` doesn't seem grammatically correct. I'm also clueless what this warning is actually telling me

crates/aura-config/src/config.rs
```diff
@@ -1007,6 +1048,21 @@ mod tests {
         assert!(hitl_timeout_conflict_warning(&hitl, 0).is_none());
     }
 
+    #[test]
+    fn test_no_inner_bound_warning() {
+        assert!(no_inner_bound_warning(0, 0).is_some());
+        assert!(no_inner_bound_warning(60, 0).is_none());
+        assert!(no_inner_bound_warning(0, 60).is_none());
+    }
```
jakedipity, line 1057: nit: missing both set case `(60, 60)`

crates/aura-config/src/config.rs
```diff
@@ -1007,6 +1048,21 @@ mod tests {
...
+
+    #[test]
+    fn test_inactivity_vs_per_call_warning() {
+        assert!(inactivity_vs_per_call_warning(60, 60).is_some());
+        assert!(inactivity_vs_per_call_warning(60, 120).is_some());
+        assert!(inactivity_vs_per_call_warning(60, 30).is_none());
+        assert!(inactivity_vs_per_call_warning(0, 120).is_none());
+    }
```
jakedipity, line 1066: nit: what about the `(0, 0)` case? Maybe I'm wrong but I feel like it should be suppressed in that case.

configs/example-math-orchestration.toml
```diff
@@ -70,6 +70,10 @@ allow_clarification = true
 # Show tool names to coordinator during planning
 tools_in_planning = "summary"
 
+[orchestration.timeouts]
+per_call_timeout_secs = 120
+inactivity_timeout_secs = 45
```
jakedipity: nit: I think it would be worthwhile to document what each of these config options do in all our of example tomls.

crates/aura-config/src/orchestration.rs
```diff
@@ -125,24 +125,28 @@ pub struct WorkerConfig {
...
+    /// Wall-clock budget (seconds) for one coordinator phase or one worker
+    /// task. 0 disables.
     #[serde(default = "default_per_call_timeout_secs")]
     pub per_call_timeout_secs: u64,
+
+    /// Inactivity window (seconds) for coordinator and worker streams; governs
+    /// both the initial silence before the first token and gaps between tokens.
+    /// 0 disables.
+    #[serde(default = "default_inactivity_timeout_secs")]
+    pub inactivity_timeout_secs: u64,
```
jakedipity: Do you think it's worthwhile tagging streaming to this config name? From the outset I was confused what was the difference between `per_call_timeout`

crates/aura-config/src/orchestration.rs
```diff
@@ -944,11 +959,28 @@ mod tests {
         "#;
         let config: OrchestrationConfig = toml::from_str(toml).unwrap();
         assert_eq!(config.per_call_timeout_secs(), 45);
+        assert_eq!(config.inactivity_timeout_secs(), 120);
         assert_eq!(config.memory_dir(), Some("/tmp/new-style"));
         assert_eq!(config.result_artifact_threshold(), 8000);
         assert_eq!(config.result_summary_length(), 1500);
     }
 
+    #[test]
```
jakedipity, line 986: Is there a test for the default values of these timeouts?

crates/aura-config/src/config.rs
Shearerbeard: I would agree that dead config should be something like invalid config, {inactivity_timeout_secs} is just interpolation so this would print "inactivity_timeout_secs (120s)" for example. https://doc.rust-lang.org/std/macro.format.html

crates/aura-config/src/config.rs
Shearerbeard: will add

crates/aura-config/src/config.rs
Shearerbeard: will add

configs/example-math-orchestration.toml
Shearerbeard: I can do that - most of our example stuff is being extracted out to docs but this one is used in integration tests so I could add a note

crates/aura-config/src/orchestration.rs
Shearerbeard: the streaming naming is a good idea - i'll run a few ideas by the channel

jakedipity (approved):

Issue #417, opened 2026-07-24
[FEATURE]:  Use models.dev to suggest models and parameters


Shearerbeard:
### Summary

Relevant to onboarding and customization - I found that opencode uses the models.dev catalogue to generate suggestions and completions for model names, defaults, and parameters. We can use this to keep our suggestions up to date for configuring aura in the aura CLI without maintaining our own provider -> model map.

<img width="1730" height="1279" alt="Image" src="https://github.com/user-attachments/assets/ce700365-ea52-4539-b1d9-af7478566fd7" />

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #464, opened 2026-07-28
feat(cli): recommend latest chatgpt and opus models in init

### LINKED ISSUE #463: [FEATURE]: Aura init doesn't recommend the latest frontier models
### Summary

The AURA INIT command is recommending older models (gpt 5.5, opus 4.x). It should recommend the latest frontier models available.

### Additional Context

_No response_

### Upload screenshots

_No response_

### Searched Issues

- [x] No similar issues found

### Code of Conduct

charlesjohnson:
Updates the `aura init` model recommendations so the suggested defaults are the latest flagships. OpenAI is bumped from `gpt-5.5` to `gpt-5.6`, and the Anthropic list is reordered so `claude-opus-4-8` is the suggested default ahead of Sonnet. Anthropic Fable is intentionally left out of the recommendation list. Recommendations live in `family_roots`, and the affected `ranking`/`spec` tests are updated to match.

Resolves #463.

### DIFF
--- a/crates/aura-cli/src/init/provider.rs
+++ b/crates/aura-cli/src/init/provider.rs
@@ -78,8 +78,10 @@ pub(crate) fn default_key_env(provider: Provider) -> Option<&'static str> {
 /// editorial choice — keep them current as providers ship new flagships.
 pub(crate) fn family_roots(provider: Provider) -> &'static [&'static str] {
     match provider {
-        Provider::OpenAI => &["gpt-5.5"],
-        Provider::Anthropic => &["claude-sonnet-4-6", "claude-opus-4-8", "claude-haiku-4-5"],
+        // GPT-5.6 ships as three flavors; `terra` (balanced) is the default —
+        // `sol` is the pricier frontier tier, `luna` the cheapest.
+        Provider::OpenAI => &["gpt-5.6-terra", "gpt-5.6-sol", "gpt-5.6-luna"],
+        Provider::Anthropic => &["claude-opus-5", "claude-sonnet-4-6", "claude-haiku-4-5"],
         Provider::Gemini => &["gemini-3.5-flash", "gemini-3.1-pro"],
         // Uncurated providers (and bedrock, which has no list endpoint).
         Provider::OpenRouter | Provider::Ollama | Provider::Bedrock => &[],
--- a/crates/aura-cli/src/init/ranking.rs
+++ b/crates/aura-cli/src/init/ranking.rs
@@ -230,20 +230,27 @@ mod tests {
             "o3",
             "o4-mini",
             "gpt-4.1",
-            "gpt-5.4",
             "gpt-5.5",
-            "gpt-5.5-2025-11-01", // dated snapshot of a recommended id
+            "gpt-5.6-sol",
+            "gpt-5.6-terra",
+            "gpt-5.6-terra-2026-07-09", // dated snapshot of a recommended id
+            "gpt-5.6-luna",
             "text-embedding-3-small",
         ]
         .iter()
         .map(|s| s.to_string())
         .collect();
 
-        // Only the recommended gpt-5.5 is shown, and the clean id wins over its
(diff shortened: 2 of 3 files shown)


crates/aura-cli/src/init/provider.rs
```diff
@@ -78,8 +78,8 @@ pub(crate) fn default_key_env(provider: Provider) -> Option<&'static str> {
 /// editorial choice — keep them current as providers ship new flagships.
 pub(crate) fn family_roots(provider: Provider) -> &'static [&'static str] {
     match provider {
-        Provider::OpenAI => &["gpt-5.5"],
-        Provider::Anthropic => &["claude-sonnet-4-6", "claude-opus-4-8", "claude-haiku-4-5"],
+        Provider::OpenAI => &["gpt-5.6"],
```
Shearerbeard: Image - gpt 5.6 is 3 models now and there can be some price consequences on which class you pick (more specifically we have to be careful about defaulting to 5.6-sol). This model string is invalid

crates/aura-cli/src/init/ranking.rs
```diff
@@ -230,20 +230,20 @@ mod tests {
             "o3",
             "o4-mini",
             "gpt-4.1",
-            "gpt-5.4",
             "gpt-5.5",
-            "gpt-5.5-2025-11-01", // dated snapshot of a recommended id
+            "gpt-5.6",
+            "gpt-5.6-2025-11-01", // dated snapshot of a recommended id
```
Shearerbeard: These are a hallucinations (5.6 is after the training date on most claude code models) - gpt 5.6 series are 2026 models and come in 3 flavors. If your coding via harness only I would recommend auto and manual verification steps.

crates/aura-cli/src/init/spec.rs
```diff
@@ -493,7 +493,7 @@ mod tests {
         a.offline = false;
         let only_openai = |var: &str| var == "OPENAI_API_KEY";
         let detected = |var: &str| (var == "OPENAI_API_KEY").then(|| "sk-detected".to_string());
-        let lister = FixedLister(vec!["gpt-5.4", "gpt-5.5", "text-embedding-3-small"]);
+        let lister = FixedLister(vec!["gpt-5.5", "gpt-5.6", "text-embedding-3-small"]);
```
Shearerbeard: This is curious and I missed it in the earlier pr but text-embedding 3 should be lumped in with vector store adapter not an inference moddel

Shearerbeard (changes requested):
This needs fixes and verification - models.dev is a good source of truth for model strings. IMO this is the last edit we should make like this in aura-init before moving to having a process built in based on models.dev syncs https://github.com/mezmo/aura/issues/417.  Noted https://github.com/mezmo/aura/issues/466 is in process and related - this adds the extra reasoning options needed to support these models and all their options.

charlesjohnson:
Thanks @Shearerbeard — you were right, the original strings were hallucinated. Corrected and verified against models.dev:

- **OpenAI:** `gpt-5.6` was invalid. Replaced with the real three-flavor family, defaulting to the balanced **`gpt-5.6-terra`** ($2.50/$15) rather than the frontier **`gpt-5.6-sol`** ($30/$180) you flagged; `gpt-5.6-luna` is listed as the cheapest option.
- **Anthropic:** bumped the suggested default to **`claude-opus-5`** (2026-07-24), which is newer than the `claude-opus-4-8` I had. Fable still excluded.

Tests in `ranking.rs`/`spec.rs` updated to the real strings and all pass. Force-pushed the corrected commit.

Understood on #417 being the durable fix (sync from models.dev) and this being the last manual edit of this kind, and that #466 adds the reasoning/effort options these models need — happy to hold this behind #466 if you'd prefer they land together.

Left the `text-embedding-3-small` fixture in `spec.rs` as-is since it's a pre-existing negative case (verifies embedding models get filtered out of the chat shortlist), but glad to move it if you want it folded into the vector-store adapter categorization here.

Shearerbeard (approved):

PR #476, opened 2026-07-29
fix(ci): fail the coverage lane when the compile fails

### LINKED ISSUE #140: make test-integration docker build has errors that are ignored...
Not sure whether that is relevant, but I see:

aura-integration-test-1  | error: failed to load source for dependency `rig-core`
aura-integration-test-1  |
aura-integration-test-1  | Caused by:
aura-integration-test-1  |   Unable to update https://github.com/mezmo/rig.git?rev=d7e9d92#d7e9d92e
aura-integration-test-1  |
aura-integration-test-1  | Caused by:
aura-integration-test-1  |   can't checkout from 'https://github.com/mezmo/rig.git': you are in the offline mode (--frozen)

charlesjohnson:
The `coverage` recipe joins the compile and the test run into one shell command with `;` separators, and prefixed the whole recipe line with `-`, so make ignored the line's status and took it from `$(MAKE) nextest || touch ...`, which is always 0; a failed `cargo build --frozen` scrolled past while the lane stayed green, since nextest runs without `--frozen` and quietly rebuilt over the network. The `-` was written to keep a *test* failure from aborting the recipe before grcov ran, and silently widened to cover the compile when that step was grafted onto the same line in cb233a55. This drops the `-` and terminates the shell explicitly on a failed compile; test failures stay on the existing marker-check path, so grcov still runs and reports. Verified all four paths (build-fail exits non-zero with grcov skipped, tests-fail runs grcov then fails, all-pass exits 0) and confirmed `cargo build --all-targets --workspace --frozen` now succeeds offline under `--network none` in the `cook-debug` image, so this does not red the lane.

Fixes #140

### DIFF
--- a/.makefiles/rust.mk
+++ b/.makefiles/rust.mk
@@ -23,11 +23,12 @@ build-rust: $(DOCKER_ENV) ## Build all rust targets
 .PHONY:coverage
 coverage: $(DOCKER_ENV) $(REPORT_DIR) $(GRCOV_BIN) ## Run the local test suite with code coverage
 	-$(MAKE) debug-PROJECT_ROOT
-	-export RUSTFLAGS="--allow=warnings -Cinstrument-coverage"; \
+	@# Terminate the shell on a failed compile: make only sees the last status.
+	export RUSTFLAGS="--allow=warnings -Cinstrument-coverage"; \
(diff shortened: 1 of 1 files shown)


Shearerbeard (approved):
Good catch

PR #484, opened 2026-07-29
Retry transient provider failures in orchestration

### LINKED ISSUE #454: provider-unavailable errors are never retried; one 429 kills a worker task

Shearerbeard:
## Why

Orchestrated workers and coordinator planning calls die on the first transient provider failure. One 429 at SSE stream-open kills a worker task (#454). This PR is staged in three commits; the first two are AURA-only and integration-green on the current rig pin.

## Commits

1. **`fix(orchestration): classify provider 5xx and connection failures`** — failure classification fix so transient provider errors are recognized as retryable.
2. **`feat(orchestration): back off transient planning retries`** — `[orchestration.retry]` config (`base_delay_ms`/`max_delay_ms`/`max_retries`); the coordinator resends a transiently failed planning call with exponential backoff, budget adjudicated per coordinator call.
3. **(not yet in this PR) rig pin bump** — picks up mezmo/rig#9, which retries retryable status codes at SSE stream-open below the agent loop. Commit 3 lands here after the rig PR merges to `mezmo`; reverting it ships commits 1-2 AURA-only on the old pin. That is why this PR is a draft.

## Interaction note (measured)

Once commit 3 lands, the rig's bounded stream-open retry (~31s worst case: 1+2+4+8+16s) runs BEFORE aura's planning backoff ever sees the error. Hard-down provider worst case per coordinator call: 4 aura attempts x ~31s rig + 3.5s aura = ~127s, vs ~3.5s today. Measured with a mock provider (`m-new-planning-nest`): 6 rig POSTs on a clean exponential, then aura retries take over.

## Verification

- Commits 1-2: `cargo test --workspace` green; Docker orchestration integration suite 95/0 on the old pin; rebased onto 0.1.8 with `cargo test -p aura --lib` (890) and `cargo test -p aura-config --lib` (195) green.
- Stage-3 behavior (rig side, not in this diff): full mocked-failure matrix (429/500/502/503/504/529 x hint/no-hint, HTTP-date hint treated as absent, exhaustion bound at exactly max_retries+1 POSTs, 400 fast-fail parity with the old pin), real-OpenAI smoke, and the nesting measurement above.

Fixes: #454

### DIFF
--- a/crates/aura-config/src/orchestration.rs
+++ b/crates/aura-config/src/orchestration.rs
@@ -157,9 +157,7 @@ impl Default for TimeoutsConfig {
 /// Retry configuration for the orchestration planning loop.
 ///
 /// The schedule is local to aura-config: it does not reuse rig's
-/// `ExponentialBackoff` (aura-config stays free of rig-core) and is
-/// independent of `RetryHint`, which governs MCP tool-error retries for a
-/// different consumer.
+/// `ExponentialBackoff` (aura-config stays free of rig-core).
 ///
 /// # Example
 ///
@@ -199,10 +197,7 @@ impl RetryConfig {
     ///
     /// Retry 1 uses the base delay; each subsequent retry doubles the
     /// previous delay, capped at `max_delay_ms`. Saturating arithmetic — a
-    /// huge `retry_num` returns the cap without panicking. Each retry is a
-    /// fresh `per_call_timeout_secs` budget, so one coordinator planning call
-    /// is bounded by `(max_retries + 1) × per_call_timeout_secs` plus the
-    /// cumulative backoff.
+    /// huge `retry_num` returns the cap without panicking.
     pub fn delay_for_retry(&self, retry_num: usize) -> std::time::Duration {
         // Multiplier is 2^(retry_num - 1): retry 1 -> base, each subsequent doubles.
         let exponent = retry_num.saturating_sub(1).min(128);
--- a/crates/aura-config/src/orchestration.rs
+++ b/crates/aura-config/src/orchestration.rs
@@ -151,6 +151,70 @@ impl Default for TimeoutsConfig {
 }
 
 // ============================================================================
+// Retry Sub-Config
+// ============================================================================
+
+/// Retry configuration for the orchestration planning loop.
+///
+/// The schedule is local to aura-config: it does not reuse rig's
+/// `ExponentialBackoff` (aura-config stays free of rig-core) and is
+/// independent of `RetryHint`, which governs MCP tool-error retries for a
+/// different consumer.
+///
+/// # Example
+///
+/// ```toml
+/// [orchestration.retry]
+/// base_delay_ms = 500
+/// max_delay_ms = 5000
+/// max_retries = 3
+/// ```
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RetryConfig {
+    /// Base delay (milliseconds). Default: 500.
+    #[serde(default = "default_base_delay_ms")]
+    pub base_delay_ms: u64,
+
+    /// Maximum delay (milliseconds). Default: 5000.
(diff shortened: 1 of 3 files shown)


justintime4tea (approved):
LGTM!

Issue #462, opened 2026-07-28
[BUG]: Docker quickstart config doesn't match the config generated by aura init command


charlesjohnson:
### Summary

The config.toml used by the Docker-based QuickStart is outdated, and doesn't match the config.toml generated by aura init.

### Reproduction

Inspect the config.toml files.

### Relevant log output

```shell

```

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #386, opened 2026-07-17
Benchmark continuation-prompt clarity for a batteries-included default config


Shearerbeard:
## Placeholder

The transitive dep bug (#221) came down to continuation-prompt clarity:
workers don't need every ancestor's output, just the right context passed down
the direct line. The #249 fix passed the
full ancestor chain under a token budget - scores improved, but the
continuation prompt and history stayed just as noisy - so we closed it and are
superseding it with the restructuring prototyped in #374. Notes and evidence
live in the terminalbench-aura repo.

This ticket tracks benchmarking that approach on terminalbench against a
neutral default config for #212 (batteries-included): a generic set of worker
prompts (explore, verify, ...) instead of application-specific roles, so the
default config ships with worker prompts that earn their place on benchmark
evidence rather than SRE-specific tuning.

Details land as the daily runs produce them - this is a stub so the work is
visible on the board.

teriyakichild:
Benchmark verdict on the neutral-workers question, from the init-template round-2 campaign (full writeup in the ai-experiments repo, `HANDOFF-init-template-round2-results.md`; runs cataloged in `aura-e2e/results-index.json`, ids `sonnet46-r2-*`).

The generic analyst/operator/debugger/verifier split from terminalbench-aura, ported onto the RCA mock-mcp suite with the same fan-out planning and nudges as the domain-worker arms (Sonnet 4.6/Bedrock, two runs each):

| Arm | Score (adj) | Tokens | ~s/scenario |
|---|---|---|---|
| Domain workers, inline guidance | 143 / 141 of 144 | 5.68M / 5.58M | 332 / 303 |
| Generic workers | 144 / 143 of 144 | 12.30M / 13.03M | 370 / 385 |

("adj" restores 12 points the harness's `worker:log-analyst present` assertion takes from any non-domain worker naming. That's a harness artifact; the assertion needs parameterizing before another arm like this.)

Score parity, 2.2x token cost, 15-25% more wall time. The cost is structural: every generic worker carries the full 23-tool surface, and the verifier adds a fixed ~87s serial phase to every scenario, including the ones that were already right. The debugger never dispatched in 24 scenarios.

So #505 keeps domain workers as the init default. Two pieces of the generic shape survived into the template: the self-contained-task-description prose in the planning section, and the verifier as an opt-in commented-out worker. Proposing the generic set ships as a reference config instead; noted on #241.

Issue #212, opened 2026-06-10
[EPIC] Batteries-Included Default Config


gjanco:
Ship a default configuration that works on first run without hand editing. Out of scope: prompt and context assembly, which belongs to #202. Serves initiative #370.

PR #505, opened 2026-08-03
feat(init): apply benchmark-validated orchestration defaults

teriyakichild:
## What

Applies the July 2026 benchmark campaign results to the config `aura init` generates. One file: `crates/aura-cli/src/init/config_template.toml`.

## Why these defaults

Three A/B rounds on the aura-e2e RCA mock-mcp suite (12 scenarios, Sonnet 4.6/Bedrock) plus an SREGym-Lite replication, cataloged in the ai-experiments repo (`aura-e2e/results-index.json`, ids `sonnet46-r2-*`):

| Change | Evidence |
|---|---|
| Discover-then-fan-out planning, null-results and synthesis guidance, delivered inline in the system prompt | 143 and 141 of 144 vs 141 for the skill-delivered reference, at roughly a third fewer tokens |
| `tools_in_planning = "full"` | 144/144 perfect RCA run at ~20% fewer tokens; SREGym-Lite diagnosis 45% -> 70% |
| `per_call_timeout_secs` 120 -> 600 | the budget bounds a worker's whole multi-turn task; the 300s reference arm lost 2 diagnostic points to it, and 120s fails healthy tasks on reasoning-model latency alone |
| `nudge_last_turn` on, `nudge_turns_remaining = 3` | fired twice in a perfect 12-scenario run; converts MaxDepthError retry spirals into partial submissions |
| `stream_inactivity_timeout_secs` 0 -> 180 | zero false fires in a 5-hour campaign; comment carries the silent-reasoning caveat |
| `max_planning_cycles` 2 -> 3 | discovery + fan-out + one targeted follow-up cycle |
| Worker slice/stopping rules, exact-name `mcp_filter` warning, opt-in commented-out verifier worker | recurring failure classes from the campaign forensics |

Domain-split workers stay. The generic analyst/operator/debugger/verifier shape from the terminalbench experiments scored parity but at 2.2x tokens, since every worker carries the full tool surface and the verifier adds a fixed serial phase to every request. Details on #386.

## Testing

`cargo test -p aura-cli init::` - 54 tests, including rendering the template through the real config parser.

Follow-up (separate change): re-sync the Docker quickstart config with the new template, as in #462.

Refs: #212, #386

### DIFF
--- a/crates/aura-cli/src/init/config_template.toml
+++ b/crates/aura-cli/src/init/config_template.toml
@@ -10,6 +10,10 @@ memory_dir = "/tmp/aura"
 
 [agent]
 name = "{{name}}"
+# Warn workers when few turns remain and require a final-turn submission —
+# converts turn-limit failures into partial results instead of lost work.
+nudge_last_turn = true
+nudge_turns_remaining = 3
 system_prompt = """
 You are an SRE Orchestrator that decomposes operational tasks into \
 specialized sub-tasks and delegates to the appropriate worker.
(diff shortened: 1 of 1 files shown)


Shearerbeard (approved):
Initial testing looked good - i'll follow up with some bechmarks based on these defaults

charlesjohnson (approved):

Issue #384, opened 2026-07-17
[EPIC] Agent Workflows


Shearerbeard:
## Overview

HITL today approves one tool call at a time, mid-execution, while an SSE stream
holds the request open. That works attended, but it's the wrong shape for
unattended remediation: an SRE wants the whole proposed sequence in front of
them - the exact calls in order, each with its intent and read/write class -
so they can approve it the way they'd approve runbook steps, before
anything runs.

The goal is for the agent to assemble that sequence itself: a reviewable
workflow of exact tool calls with per-step intent (we already force
`_aura_reasoning` on orchestration tool calls), destructive steps gated. AURA's
half of the equation is inputs and outputs: emit the proposal over the HITL
webhook with enough context and intent that the receiving side can make a real
approval decision. The decision comes back, and only the approved version
executes.

This builds directly on shipped HITL: the approval exchange plus header
forwarding (#276) and intent forwarding (#277) are the same wires this needs.

## User outcome

An SRE (or a governance layer acting for one) can:

- See the proposed workflow before execution: ordered calls, arguments,
  per-step intent, which steps are destructive
- Approve or deny the exact sequence, knowing the approval is bound to that
  version and any drift from it fails closed

## Scope notes

- This epic is the AURA side only: proposal format, webhook payloads, decision
  handling, execution binding. The review UX and routing live downstream.
- Related state work is tracked in #325 / #209 / #210 - a proposal that waits
  hours on approval needs somewhere durable to live, which is the same parking
  problem HITL V2 already has.

Issue #271, opened 2026-06-24
HITL Park/Reify


Shearerbeard:
HITL with no in band single request blocking on approval (unattended). First webhook flow is just verification. Target later sends results back through approval. Entire orchestration and agent execution flow needs to park either entire execution or the specifically blocked section until other work blocks then reify the whole thing. This needs durable state storage (ties into #210).

- Blocked on durable serialization of orchestration flow
- Blocked by state storage (cross pod) so that session can be reified - auth is optional but needs to be solved at a later date
- Blocked on some sort of target cryptographic identity verification to prevent forgery.

PR #507, opened 2026-08-05
fix(a2a): deliver a terminal frame on cross-instance cancel

### LINKED ISSUE #475: Cross-instance cancel leaves the executing instance's subscribers without a terminal frame
In a two-instance deployment, cancel an A2A task through the instance that is not executing it and the executing instance's own subscribers never learn the outcome: their stream closes with no terminal frame, while subscribers relaying through the other instance receive `TASK_STATE_CANCELED` normally. The task store itself ends correct and terminal - only local stream delivery is affected.

The asymmetry is in the subscription paths. A subscriber on the executing instance is served by the local execution stream (`crates/aura-web-server/src/a2a/request_handler.rs:101-102` at `18a37458`); a subscriber on any other instance is served by the bus relay (`request_handler.rs:106-107`). When a cancel arrives at a non-executing instance, that instance drives the cancel and writes the terminal status to the shared store, and the bus carries it to relay subscribers. The executing instance's routed-cancel listener only stops the execution and discards the cancel's events (`bus_bridge.rs:294-310`, including the discard loop), so the local execution stream ends without a terminal frame.

A client that keys off the terminal frame (rather than polling the task store) never receives one, and cannot tell a cancelled task from a dropped connection.

Reproduced deterministically (12/12) on a 2-instance compose rig: send on instance A, subscribe on both, cancel through B; A's subscriber receives only non-terminal frames and a bare stream close.

justintime4tea:
Cancelling an A2A task through an instance that is not executing it left the executing instance's own subscribers without a terminal frame: their stream closed after the last non-terminal event, while subscribers relaying through another instance received TASK_STATE_CANCELED normally. The task store ended correct and terminal — only local stream delivery was affected — so a client that keys off the terminal frame rather than polling the store could not tell a cancelled task from a dropped connection.

The routed-cancel listener ran the inner executor's cancel purely for its side effect and drained the resulting events into the void. Stopping an execution ends its stream without a terminal event by design, since cancel() is what emits Canceled, so discarding those events left the local execution stream with no outcome to report and the handler closed the subscription when that stream ended.

Forward the cancel's events into the execution stream instead. The merge is biased toward them — a stopped execution has nothing further to say about its own outcome — and the cancel's terminal status ends the stream. Fixing this at the executor covers every local consumer at once: tasks/{id}:subscribe, message:stream, and the stored record. The frames are deliberately not republished, because the instance that received the cancel already puts the terminal frame on the fan-out topic for relays.

That makes both ends of a routed cancel record the same terminal status, which the Redis update script rejected outright: it treated any write to a terminal record as a conflict. Left as it was, the losing side of the race turned a successful cancel into a failure — the cancelling instance's own cancelTask call returning "task ... is terminal and cannot be updated". Refuse a terminal write only when it would change the recorded state; re-recording the state already stored leaves the record alone and reports its unchanged version. Completed over Canceled is still rejected, so an execution that misses a routed cancel still cannot overwrite the outcome.

Separately, release the execution's cancel-map entry and its cancellation-registry registration from a Drop guard, the way the active-request tracker already does. A consumer stops polling after a terminal event, which drops the execution generator where it stands, so cleanup written at the end of the generator body never ran when an execution failed mid-stream — leaking the entry and the agent handle it owns, along with that agent's MCP connections. The guard runs on every exit path, including the two other drop points reachable while an execution is registered. This moves the cancel map to a std::sync::Mutex, since Drop cannot await; nothing holds the lock across an await, which the Send bound on the executor's streams enforces at compile time.

Fixes: #475

### DIFF
--- a/crates/aura-web-server/src/a2a/agent_executor.rs
+++ b/crates/aura-web-server/src/a2a/agent_executor.rs
@@ -1,6 +1,6 @@
 use std::cmp::Reverse;
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex, MutexGuard};
 
 use a2a::{
     A2AError, AgentCapabilities, AgentCard, AgentInterface, AgentSkill, Artifact, ListTasksRequest,
@@ -14,7 +14,6 @@ use aura::{RequestCancellation, StreamItem, StreamedAssistantContent, StreamingA
 use futures_util::StreamExt;
 use futures_util::stream::BoxStream;
 use serde_json::Value;
-use tokio::sync::Mutex;
 use tokio_util::sync::CancellationToken;
 use tracing::{Level, event};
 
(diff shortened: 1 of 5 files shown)


Shearerbeard (approved):
Awesome fix and unblocker for #271 and final implementation of #384 - similar to my last review we will probably generalize some of the park and a2a machinery to a more generic location with or after #271 so it can be reused for other invocations

PR #508, opened 2026-08-05
fix(a2a): include assistant turns in conversation history

### LINKED ISSUE #368: [BUG]: A2A conversation history omits assistant turns — follow-up messages don't see prior answers
Ref: #325, #328

## Summary

A follow-up A2A message on an existing `contextId` is given the prior **user** turns but never the agent's own **responses**. Every follow-up therefore behaves as if it has never spoken.

Observed against `aura-web-server` with A2A enabled:

```
task 1 (user): "In exactly three sentences, explain what a log pipeline is."
               -> completes, answers correctly

task 2 (user, same contextId): "How many sentences did you just use?"
               -> "I haven't responded to that prompt yet, so I used zero sentences."
```

Expected: the follow-up sees the previous assistant answer and replies "three".

## Root cause

`get_history_for_context` (`crates/aura-web-server/src/a2a/agent_executor.rs:549`) rebuilds history by listing tasks for the context and flat-mapping each task's `history` field:

```rust
let chat_history: Vec<aura::Message> = task_history
    .into_iter()
    .flat_map(|t| t.history.unwrap())
    .filter_map(|m| convert_a2a_msg_to_aura(&m))
    .collect();
```

The context lookup itself works — `list(context_id)` returns every task in the conversation. The problem is what lives in `task.history`: the agent's response is written to **artifacts**, not to `history`. A completed task looks like this:

```
status: TASK_STATE_COMPLETED
history messages: 1
   role=ROLE_USER   text="In exactly three sentences, explain what a log pipeline is."
artifacts: [("response", 84 parts), ("final", 1 part)]

ROLE_AGENT present in history? False
```

So `flat_map(|t| t.history)` can only ever yield user turns. `convert_a2a_msg_to_aura` has a `Role::Agent => Message::assistant(...)` arm that is unreachable in practice.

The same call also passes `include_artifacts: Some(false)`, so the assistant text is not even fetched from the store.

## Reproduction

**Pre-requisites**
- `aura-web-server` running with `--enable-a2a` and any single-agent config (no HITL gating).

**Steps**

1. Send a first message and note `task.contextId`:

```bash
curl -s -X POST http://localhost:8080/a2a/v1/message:send \
  -H "Content-Type: application/json" -H "x-aura-model: <agent-alias>" \
  -d '{"message":{"messageId":"m1","role":"ROLE_USER","parts":[{"text":"In exactly three sentences, explain what a log pipeline is."}]}}'
```

2. Poll `GET /a2a/v1/tasks/{id}` until `TASK_STATE_COMPLETED`. Note that `history` holds only the `ROLE_USER` message while the answer sits in the `response` artifact.

3. Send a follow-up on the same `contextId`:

```bash
curl -s -X POST http://localhost:8080/a2a/v1/message:send \
  -H "Content-Type: application/json" -H "x-aura-model: <agent-alias>" \
  -d '{"message":{"messageId":"m2","contextId":"<CTX>","role":"ROLE_USER","parts":[{"text":"How many sentences did you just use?"}]}}'
```

4. Poll it. The answer denies having responded.

## Why the existing test does not catch this

`test_sequential_tasks_share_context` (`crates/aura-web-server/tests/a2a_test.rs:713`) passes, because the only thing it asserts is that task 2 can recall **what was asked** in task 1:

```
task 1 (user): "Call mock_tool with message='context-test-{uuid}'. Do not say anything else."
task 2 (user): "Tell me what was asked in the first call for this context."
assert: response contains "context-test-{uuid}"
```

That UUID originates in task 1's **user** message — the half of history that does carry. The test validates the user-turn path and is blind to the missing assistant turns.

The test's own comment states the expectation that is currently false:

```rust
// Wait for task 1 to complete so its full history (tool call, tool result,
// assistant response) is persisted in the task store before task 2 starts.
```

Tool calls, tool results, and assistant responses are not in `task.history`.

## Suggested fix

Reconstruct the assistant turn from the task's `response` artifact (requiring `include_artifacts: Some(true)` in the `ListTasksRequest`), or write the agent's final message into `task.history` as `ROLE_AGENT` when the task completes.

Whichever way, extend `test_sequential_tasks_share_context` (or add a sibling) to assert on something only the **assistant** turn could supply — the current phrasing passes without it.

## Scope note

Pre-existing, not a regression. `git diff main..feature/GH-327-session-storage-trait-and-in-memory` touches only `shared_task_store.rs` under `crates/aura-web-server/src/a2a/`; `get_history_for_context` dates to `349014dc feat(a2a): initial implementation`.

Relevant to the session-storage roadmap: #328 (Redis/Valkey A2A task store) is motivated partly by making "conversation history by `context_id` survive load-balancing". Making that history durable and cross-pod will not help while the assistant half of every exchange never reaches the store. Worth fixing before or alongside #328.

justintime4tea:
A follow-up message on an existing contextId was given the prior user turns but never the agent's own replies, so it answered as if it had never spoken.

get_history_for_context rebuilt history from Task::history alone. The upstream request handler writes that field once, with the prompt, and never appends to it; the agent's reply is streamed out as artifacts instead, leaving the Role::Agent arm of convert_a2a_msg_to_aura unreachable. The same call also passed include_artifacts: Some(false), so the reply was never fetched, and sorted the context's tasks newest-first, which would have replayed a multi-task exchange backwards.

History now rebuilds the assistant turn from the artifacts the client saw: the streamed "response" chunks concatenated with no separator, falling back to "final" for a reply that only arrived as a final response. Tool-call, reasoning, and scratchpad artifacts are skipped. Tasks replay oldest-first, and a task already carrying a ROLE_AGENT message keeps it rather than gaining a second one. The two artifact ids are now module consts shared with the sites that emit them, so reader and writer cannot drift apart.

A2A clients still see only the prompt in task.history. The upstream handler owns every task write and carries a stale in-memory task across events, so writing the agent turn back would either race it or have to go through SharedTaskStore, which conflicts with the Redis store's terminal-immutability rule. Reconstructing on read works with both stores and touches no write path.

test_sequential_tasks_share_the_assistant_turn covers the half that test_sequential_tasks_share_context cannot. It routes a per-run secret in through a forwarded header that the agent reports, then sends the follow-up without that header so the tool cannot surface the value again, and asserts the follow-up made no tool call at all, leaving the prior assistant turn as the only possible source.

Fixes: #368

### DIFF
--- a/crates/aura-web-server/src/a2a/agent_executor.rs
+++ b/crates/aura-web-server/src/a2a/agent_executor.rs
@@ -1,4 +1,3 @@
-use std::cmp::Reverse;
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex, MutexGuard};
 
@@ -23,6 +22,12 @@ use crate::{
 };
 
 const PLAIN_TEXT: &str = "text/plain";
+
+/// Artifact id of the assistant's reply text, one part per streamed chunk.
+const RESPONSE_ARTIFACT_ID: &str = "response";
+
+/// Artifact id of the assistant's complete reply and the turn's token usage.
+const FINAL_ARTIFACT_ID: &str = "final";
 
 pub struct AuraAgentExecutor {
     app_state: Arc<AppState>,
@@ -286,15 +291,14 @@ impl AgentExecutor for AuraAgentExecutor {
                     Ok(StreamItem::StreamAssistantItem(StreamedAssistantContent::Text(t))) => {
                         event!(Level::DEBUG, request_id, t, "stream content received");
 
-                        const ARTIFACT_ID: &str = "response";
-                        let append = append_tracker.entry((task_id.clone(), context_id.clone(), ARTIFACT_ID.to_owned()))
+                        let append = append_tracker.entry((task_id.clone(), context_id.clone(), RESPONSE_ARTIFACT_ID.to_owned()))
                             .and_modify(|e| *e = true)
                             .or_insert(false);
 
                         event!(Level::DEBUG, request_id, "response returned and should be appended: {}", *append);
 
                         let artifact = Artifact {
-                            artifact_id: ARTIFACT_ID.to_owned(),
+                            artifact_id: RESPONSE_ARTIFACT_ID.to_owned(),
                             name: Some("Response".to_owned()),
                             description: None,
                             parts: vec![Part::text(t)],
@@ -413,13 +417,12 @@ impl AgentExecutor for AuraAgentExecutor {
                         event!(Level::DEBUG, request_id, "reasoning delta");
                     }
                     Ok(StreamItem::Final(final_info)) => {
-                        const ARTIFACT_ID: &str = "final";
-                        let append = append_tracker.entry((task_id.clone(), context_id.clone(), ARTIFACT_ID.to_owned()))
+                        let append = append_tracker.entry((task_id.clone(), context_id.clone(), FINAL_ARTIFACT_ID.to_owned()))
                             .and_modify(|e| *e = true)
                             .or_insert(false);
 
                         let artifact = Artifact {
-                            artifact_id: ARTIFACT_ID.to_owned(),
+                            artifact_id: FINAL_ARTIFACT_ID.to_owned(),
                             name: Some("Final Info".into()),
                             description: None,
                             parts: vec![Part::text(final_info.content)],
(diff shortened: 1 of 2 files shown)


Shearerbeard (dismissed):
Approving for now but i think after we flesh out park/reify #271  and get closer to task attach/detach this will have the same generic home + backend for all agent initializations - not just specific A2A

Shearerbeard (approved):

PR #513, opened 2026-08-06
feat(cli): enable scratchpad for servers added via /mcp add

### LINKED ISSUE #512: /mcp add: enable scratchpad interception for newly added servers
## Problem

Servers written by the `/mcp add` wizard get an empty scratchpad table. Since interception only applies to tools matching an entry under `[mcp.servers.<name>.scratchpad]`, wizard-added servers never have large tool outputs diverted to the scratchpad — even when the config already has `[agent.scratchpad] enabled = true` (as configs generated by `aura init` do). Users have to know about the per-server table and add it by hand.

## Proposal

Have the wizard write a wildcard entry for every server it adds:

```toml
[mcp.servers.<name>.scratchpad]
"*" = { min_tokens = 5120 }
```

- The entry appears in the wizard's config preview before the write is confirmed, so it can still be declined with the rest of the server.
- When `[agent.scratchpad]` is not enabled, print the prerequisites (`enabled = true`, top-level `memory_dir`, `[agent.llm].context_window`) instead of editing those sections — flipping `enabled = true` on a config missing the other two would fail validation on restart.
- The entry is inert while agent-level scratchpad is off, so writing it unconditionally is safe.

teriyakichild:
Fixes #512

## What

Servers written by the `/mcp add` wizard now get a wildcard scratchpad entry:

```toml
[mcp.servers.<name>.scratchpad]
"*" = { min_tokens = 5120 }
```

Large tool outputs are diverted to scratchpad storage instead of filling the context window. The entry shows in the wizard's config preview before the write is confirmed, so declining the write still declines everything.

## Why not prompt, and why not touch [agent.scratchpad]

- The entry is inert while agent-level scratchpad is off, so writing it unconditionally is safe and saves a wizard question.
- When the config lacks `[agent.scratchpad] enabled = true`, the wizard prints the prerequisites (`enabled = true`, top-level `memory_dir`, `[agent.llm].context_window`) instead of editing them — flipping the flag on a config missing the other two would fail validation on restart.

## Changes

- `aura-cli`: new `enable_scratchpad` step in the wizard; consolidated the three read-and-parse config readers (`orchestrated_workers`, `config_has_server`, and the new `agent_scratchpad_enabled`) onto a shared `read_config_doc` helper
- `aura-config`: added `McpServerConfig::scratchpad_mut` alongside the existing accessor
- Tests: writer round-trip proving the wildcard entry survives serialization and reloads through the loader; rendered-config assertion; `[agent.scratchpad].enabled` reader coverage (true/false/absent/missing file)

## Verification

- `cargo test -p aura-cli -p aura-config` — all green (438 + 196)
- `cargo clippy --all-targets` clean on both crates; `cargo check -p aura-cli --no-default-features` confirms the HTTP-only build is unaffected

Docs note: the `/mcp add` page in `mezmo/documentation` needs a line about the new scratchpad step.

### DIFF
--- a/crates/aura-cli/src/repl/mcp/wizard.rs
+++ b/crates/aura-cli/src/repl/mcp/wizard.rs
@@ -1,5 +1,6 @@
 //! Guided `/mcp add` flow: pick a catalog or custom server, collect
-//! credentials with masked input, write the config + `.env`.
+//! credentials with masked input, enable scratchpad interception for the
+//! server's tool outputs, write the config + `.env`.
 //!
 //! Standalone mode only, and deliberately deterministic: no LLM ever
 //! sees a credential.
@@ -9,6 +10,7 @@ use std::fs;
 use std::path::Path;
 use std::time::Duration;
 
+use aura_config::ScratchpadToolEntry;
 use aura_config::config::McpServerConfig;
 
 use super::catalog::{CATALOG, CatalogEntry, Template};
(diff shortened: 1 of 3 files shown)


Shearerbeard (approved):
LGTM

PR #525, opened 2026-08-11
fix(a2a): durably record approval decisions on the redis backend

### LINKED ISSUE #474: Approval decisions on the redis backend leave no durable record
When AURA runs with the Redis session store, resolving a HITL approval persists nothing. The Redis store's `resolve` ignores the decision it is given (the `_decision` parameter at `crates/aura-web-server/src/session_store/redis/approval_store.rs:101-112`, sha `18a37458`) and deletes the parked request record with an atomic `GETDEL` (`approval_store.rs:52-70`) as its at-most-once claim. The decision's only carrier is the wake publish, and that publish is fire-and-forget (`crates/aura/src/hitl/registry.rs:170-188`): if it is lost, the decision is gone, and the parked call waits out its timeout and fails closed even though an operator already approved it. A resolver crash after the `GETDEL` and before the publish loses the decision the same way.

This matters for any deployment that treats approvals as an audit surface: nothing records that a decision happened or what it was. An approval can be granted and never take effect. Cross-instance resolve itself works, and the fail-closed timeout backstop holds - the gap is durability, not routing.

Reproduced deterministically (10/10) on a 2-instance compose rig by suppressing the wake channel and observing the approval fail closed with no recoverable decision.

Durable approval state is in scope for #271; this issue tracks the defect on main until that work lands.

justintime4tea:
Resolving a HITL approval deleted the parked record and carried the decision only on the fire-and-forget wake publish: a lost publish (or a resolver crash between the claim and the publish) lost an operator's decision, and the parked call failed closed at its timeout despite the grant, leaving no record that a decision happened.

The store is now the durable carrier and the bus only the low-latency wake:

- ApprovalStore::resolve records the decision atomically with the at-most-once claim (one Lua script on redis; a decided map in memory), readable back through the new ApprovalStore::decision.
- The parked approval's wake task polls the store as a fallback, so a lost publish costs at most one poll interval, and a failed wake subscription no longer means an unwakeable park.
- The await's timeout backstop reads the store once more before failing closed, covering a decision recorded inside the final poll gap. Cancellation still fails closed without a store read.

Fixes: #474

### DIFF
--- a/crates/aura-web-server/src/session_store/redis/approval_store.rs
+++ b/crates/aura-web-server/src/session_store/redis/approval_store.rs
@@ -1,22 +1,28 @@
 //! Redis-backed HITL approval store: a parked approval is readable and
-//! resolvable from any instance.
+//! resolvable from any instance, and a resolution leaves a durable decision
+//! record the parking instance can recover if the bus wake is lost.
 //!
 //! Key schema (all under the configured `key_prefix`, default `aura`):
 //!
-//! | Key                            | Type                 | Purpose                  |
-//! | ------------------------------ | -------------------- | ------------------------ |
-//! | `{p}:approval:{decision_id}`   | string (record JSON) | parked approval record   |
-//! | `{p}:approval:req:{request_id}`| set of decision ids  | `cancel_request` fan-out |
+//! | Key                                    | Type                 | Purpose                  |
+//! | -------------------------------------- | -------------------- | ------------------------ |
+//! | `{p}:approval:{decision_id}`           | string (record JSON) | parked approval record   |
+//! | `{p}:approval:decision:{decision_id}`  | string (record JSON) | recorded decision        |
+//! | `{p}:approval:req:{request_id}`        | set of decision ids  | `cancel_request` fan-out |
 //!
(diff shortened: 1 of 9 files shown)


crates/aura-web-server/src/session_store/redis/approval_store.rs
```diff
@@ -27,6 +33,33 @@ use super::request_err;
 const MIN_TTL_SECS: u64 = 1;
 /// Margin the request index's TTL keeps over its newest record's TTL.
 const REQ_INDEX_TTL_MARGIN_SECS: u64 = 60;
+/// Margin the decision record's TTL keeps over the parked record's remaining
+/// TTL.
+const DECISION_TTL_MARGIN_MS: u64 = 60_000;
```
justintime4tea, line 38: @Shearerbeard this should be a long enough margin of time, yes? "True" durability will come with the park/reify stuff so I figured a TTL here would be fine since this isn't meant to be "truly" durable, just durable enough to resolve things.

Shearerbeard: I think TTL to cover any timing overlap is good for now - this works

Shearerbeard (approved):

PR #543, opened 2026-08-14
fix(telemetry): add system prompt attribute to spans

### LINKED ISSUE #272: OTel traces don't include the system prompt
Design side decision from early days of OSS decision is that OTEL doesn't capture system messages. This is simply to add them.

- [ ] Single Agent
- [ ] Coordinator
- [ ] Worker

dhable:
Records the assembled preamble, which includes the skills catalog and orchestration instructions, on a span each time these system prompt instructions are sent to the LLM. This instrumentation is controlled by the existing OTEL_RECORD_CONTENT and OTEL_CONTENT_MAX_LENGTH config options.

Fixes: #272


dhable:

jakedipity (approved):

PR #544, opened 2026-08-14
Approver identity header forwarding to downstream MCP calls

### LINKED ISSUE #496: [FEATURE]:  HITL approver identity forwarding to downstream
### Summary

Optional config to take identity headers from the approver takes president for downstream MCP actions

### Additional Context

_No response_

### Upload screenshots

_No response_

### Searched Issues

- [x] No similar issues found

### Code of Conduct

Shearerbeard:
Forwards the approving human's identity headers from an approved webhook decision onto the single gated MCP call that decision unblocks.

- The approval decision id is stamped at both entry points, the webhook arm and the `request_approval` tool, so every gated call carries one.
- An approved webhook response can supply the configured `tool_headers_from_response` values. They are captured as validated overrides on the decision, only on approval, and the capture fails closed naming any missing header, never a value.
- The overrides ride the rmcp request extension onto exactly one outbound request on the http-streamable and SSE transports, replacing the frozen client identity for that call only. A gated stdio call carrying overrides fails closed.
- The gated call's `mcp.tool_call` span records the applied override names, never values. A route that would capture over cleartext warns at boot instead of being rejected, since TLS can terminate outside the process.
- A hitl integration suite lands behind its own feature flag with a make target against the mock fixture. The transport and route-scope decisions are recorded in an ADR.

This change was reviewed as five parts (#571, #572, #573, #574, then this PR). GitHub's stack merge would not land the stack against this repository's review rules, so the parts are closed unmerged and the whole change lands here as one rebase merge. The per-part review threads stay on those PRs; the commits here are the same patches rebased onto current `main`.

Fixes: #496

### DIFF
--- a/.makefiles/aura.mk
+++ b/.makefiles/aura.mk
@@ -262,3 +262,29 @@ test-integration-session-store-local: $(REPORT_DIR) ## Start an ephemeral Valkey
 	test_exit=$$?; \
 	docker stop $(VALKEY_TEST_NAME) >/dev/null 2>&1 || true; \
 	exit $$test_exit
+
+# --- HITL header-forwarding integration tests (needs only mock-mcp; each ---
+# --- test case spawns and tears down its own aura-web-server) ---
+
+.PHONY:test-integration-hitl-local
+test-integration-hitl-local: $(REPORT_DIR) ## Start mock-mcp, run HITL header-forwarding integration tests, then cleanup
+	@echo "Starting mock-mcp for HITL header-forwarding testing..."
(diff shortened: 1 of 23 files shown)


Shearerbeard:
@adrianfurlong - tagging you as a functionality reviewer. The ticket is on GH if you have any notes when you get back tomorrow

justintime4tea (dismissed):

justintime4tea (dismissed):

dhable (dismissed):

dhable (approved):

PR #555, opened 2026-08-18
[4/4] feat(web): add detail query parameter to /aura/info

### LINKED ISSUE #550: [FEATURE]:  /aura/info surface tool details
### Summary

/aura/info endpoint shows MCP data from hitting the live servers but does not show tool schemas. This feature should be gated by some sort of argument as the tool schemas get verbose.

- Optionally return a list of tools with their name per server
- Optionally go a step further and return the JSONSchema for input/output schemas on each tool (insanely verbose)

This allows external systems to query AURAs current capabilities to make suggestions or write rules around governance when implementing HITL interrupts. 

### Additional Context

_No response_

### Upload screenshots

_No response_

### Searched Issues

- [x] No similar issues found

### Code of Conduct

dhable:
External systems need AURA's live tool inventory to write governance
rules and decide where HITL interrupts belong.

Add `?detail=tools` for spec-shaped MCP `Tool` objects per server,
and `?detail=tools:summary` for names and descriptions. Omitting the
parameter leaves the response and its cost unchanged; an unknown
token is rejected with 400.

Discovery runs per request with the caller's forwarded headers, so
the result reflects that caller's authorization. A 30s timeout
bounds it.

Fixes: #550

### DIFF
--- a/crates/aura-web-server/src/handlers.rs
+++ b/crates/aura-web-server/src/handlers.rs
@@ -8,7 +8,7 @@ use aura::{
 use aura_events::{AgentInfo, ServerInfo};
 use axum::Json;
 use axum::body::Body;
-use axum::extract::State;
+use axum::extract::{Query, State};
 use axum::http::{HeaderMap, StatusCode};
 use axum::response::{IntoResponse, Response};
 use bytes::Bytes;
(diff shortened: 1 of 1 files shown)


Shearerbeard (dismissed):

dhable:
> 16:17:33  aura-integration-test-1  |     Request 1: Tool results should be in content. Got: The upstream model provider (openai/gpt-5.1) returned an error and the request could not be completed. Please try again or contact support if the issue persists.

Not sure if this is a temporary issue. Seems like it should be since the other PRs all ran the same test successfully in isolation.

Shearerbeard (approved):

PR #557, opened 2026-08-18
feat(cli): configurable status line under the input frame

ericlake:
## Summary

- Replace the fixed `prompt | completion | context` row under the input frame with a segment-based status line: model, cwd, git branch, context meter, `in / out` tokens, scratchpad `in / out`, and MCP tally. Rendered locally from state the CLI already holds — no extra requests or tokens. The line is fitted to the terminal width (cwd shortened first, then trailing segments dropped); the right-hand `AURA, by Mezmo!` / `esc to stop` text is only added once every segment fits.
- `[status_line] segments = [...]` in the layered `cli.toml` picks and orders segments (project overrides global). Unknown names warn on stderr and are skipped; omitting the block keeps the default set. Git branch is read from `.git/HEAD` (worktree pointer files followed) — no subprocess.
- Consume `aura.session_info` (`model`, `model_context_limit`) and `aura.tool_usage` so the context meter measures against the model's real window and updates mid-turn during tool loops. The meter renders only when a window is reported — single agents with `[agent.llm].context_window` set — and is absent otherwise, including orchestration, where `aura.usage` is an aggregate across many agent contexts. This retires the old `Context left: N%` figure, which was measured against a hard-coded 8M compaction ceiling.
- Per-segment theme roles: hues lifted from the `.welcome` logo gradient in `normal`, bold ANSI in `high-contrast`, dim in `no-color`. Context and MCP switch to warning/error colours when degraded (≥70%/≥90% fill; failed servers).
- Adjacent fixes surfaced while wiring: scratchpad counts were displayed ÷4 (they are tokens on the wire); the optional response-summary round-trip overwrote the context figure with its own tiny request; switching models left the previous window in place for a turn.
- Docs: `crates/aura-cli/README.md` (`cli.toml` example + segment reference) and the CLI section of `CLAUDE.md`.

Deliberately unchanged, for reviewers: single-agent `aura.usage.prompt_tokens` is captured from the first LLM call of a turn only (`UsageState::store_usage`), so `in` under-reports billed input on tool-heavy turns; changing that alters wire semantics for every client and is out of scope here.

## Screenshots
image


Shearerbeard (dismissed):

crates/aura-cli/src/theme.rs
```diff
@@ -322,6 +340,37 @@ pub static NORMAL: Theme = Theme {
     key_label: Style::fg(Color::Reset),
     stream_aura: Style::fg(Color::Yellow),
     stream_message: Style::fg(Color::Cyan),
+    // Status line hues are lifted from the `.welcome` logo gradient.
+    status_model: Style::fg(Color::Rgb {
```
justintime4tea: It looks like the other styles use `Color::SomeColor`

It'd be awesome if these colors were instead derived from `Style::fg(Color::SomeColor)` as well because then they use whatever colors the user has configured for their terminal. Not everyones red is red.

Using hard-coded RGB's, a user could have some crazy color scheme and then either not have visible status elements or elements are severely unreadable due to contrast between the status elements and the background color... could we test using `Color::SomeColor` for these instead of hard-coded RGB?

ericlake: Yeah it probably hard coded them because I told it to based them off of the color of the Aura logo. Let me see what we can do.

justintime4tea (approved):

PR #566, opened 2026-08-20
feat(otel): fix empty assistant content 400, enrich phoenix span

justintime4tea:
Reasoning models that call a tool without emitting any text produced an assistant message with an empty content array, which OpenAI rejects with a 400 `empty_array`. Retries re-sent the same history, so the run failed deterministically and reached the user as a tooling error instead of an answer.

Pin rig-core to the merged `mezmo` head, which omits assistant content when it is empty. The workspace dependency and the `[patch.crates-io]` entry have to move together or the build resolves two rig versions. The same bump carries the http_client stream-open retries and the turn-span attributes below.

Phoenix prices a span only when it is LLM-kind and carries both `llm.model_name` and token counts. Aura recorded aggregate token counts on its own spans, which forced a choice between pricing the root span or the per-phase spans, and risked double-counting once Phoenix rolled the descendants up.

Take token counts off Aura's spans entirely. The rig fork now records provider, model, and per-call usage on each `agent.turn` span, making the turns the single cost anchor: every aggregate — per worker, per phase, per trace — is Phoenix's rollup of them.

- pass provider and model to every `AgentBuilder` — the six single-agent provider arms, the coordinator, and each worker — so turn spans identify the model that actually ran them, and per-worker LLM overrides price at their own rate.
- drop `set_token_usage` and its call sites: the planning span, the worker spans, the streaming root, and the non-streaming completion path. `is_orchestration` and `StreamOtelContext::usage_state` existed only to arbitrate which span held the aggregate, so they go with it.

Aura's agent-level spans (`agent.stream`, `agent.prompt`, `agent.chat`, `orchestration.planning`, `orchestration.worker`) still export as LLM so Phoenix renders the tools and invocation parameters they advertise. Carrying no token counts, they price at zero and leave the turn rollup intact. `infer_span_kind` no longer inspects attributes, only the span name.

Phoenix also reads a set of OpenInference attributes Aura never emitted, leaving its UI showing "Unnamed Agent" and empty panes. Fill them in:

- `llm.provider` beside `llm.system`, since the pricing table matches on provider and model together.
- agent names on every span that creates an agent.
- `user.id` from a new OpenAI-compatible `user` request field, and `metadata` from the request's metadata map.
- `llm.invocation_parameters` from the effective LLM config, and `llm.tools.{i}.tool.json_schema` from the MCP schemas each agent advertises, filtered per worker.
- `llm.prompt_template.*` on worker spans, showing the task template beside the values filled into it. The variables carry request content, so they follow the `OTEL_RECORD_CONTENT` gate and the usual truncation; the template itself is static and always recorded.
- `retrieval.documents.{i}.*` on vector searches, which now export as RETRIEVER spans.
- `aura.version` and `aura.mode` on the streaming root, so traces filter by release and by serving path, plus `service.version` on the tracer resource.
- `openinference.project.name`, from `PHOENIX_PROJECT_NAME`. Phoenix routes traces to a project by that resource attribute alone and ignores `service.name`, so without it everything lands in the default project.
- an `exception` span event carrying `exception.message` on every error, which is the shape Phoenix's exception view reads.

Dashboards and alerts should be checked before this ships: `llm.token_count.total` and `llm.token_count.tool_completion` are no longer emitted on Aura spans. Per-turn prompt and completion counts on `agent.turn` replace them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3757,7 +3757,7 @@ dependencies = [
 [[package]]
 name = "rig-core"
 version = "0.28.0"
-source = "git+https://github.com/mezmo/rig.git?rev=e6f008eaf555733d93368c9dd174d1d61d7d096d#e6f008eaf555733d93368c9dd174d1d61d7d096d"
+source = "git+https://github.com/mezmo/rig.git?rev=5b0238a1d74e1d7c68034384b69e17f62890d517#5b0238a1d74e1d7c68034384b69e17f62890d517"
 dependencies = [
  "as-any",
  "async-stream",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ license = "Apache-2.0"
 repository = "https://github.com/mezmo/aura"
 
 [workspace.dependencies]
-rig-core = { git = "https://github.com/mezmo/rig.git", rev = "e6f008eaf555733d93368c9dd174d1d61d7d096d", default-features = false, features = ["rmcp", "reqwest-rustls"] }
+rig-core = { git = "https://github.com/mezmo/rig.git", rev = "5b0238a1d74e1d7c68034384b69e17f62890d517", default-features = false, features = ["rmcp", "reqwest-rustls"] }
 # Pinned to exact versions compatible with our rig-core 0.28.x fork.
 # Newer versions pull in rig-core 0.31+ which is semver-incompatible with the fork.
 rig-bedrock = "=0.3.10"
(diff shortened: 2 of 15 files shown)


Shearerbeard (dismissed):

Shearerbeard:
@justintime4tea this looks solid but I do have some concerns that I think will show up in testing - the tool schema can be really large and from the diff I'm not sure how often were sending that. @teriyakichild and I will do some testing today to see what it looks like - thank you for the improvements!

justintime4tea:
> The PR appears safe to merge, with only non-blocking cleanup needed to relocate behavioral field documentation to the code that implements it.
> 
> The changed runtime and telemetry paths have no accepted blocking defect; the retained issue is documentation placement that may drift as telemetry behavior evolves.

Ya, maybe we need to add "AURA_INCLUDE_TOOLS_IN_OTEL" env - it's just really helpful in diagnosing why a huge token usage is occuring when it's the result of 31 tools from no filtering pattern 🤣 Nice to see in the traces. It can be big if you have a lot of tools... but then that's the problem this helps identify?

Shearerbeard (approved):

Shearerbeard (approved):

PR #571, opened 2026-08-22
Stamp the approval decision id at both entry points

Shearerbeard:
#531 left two approval surfaces without a decision id on their spans: the webhook arm resolves through the client seam without calling `decide`, and the `request_approval` tool skips `decide_for_gate`. This stamps both public entries and routes the gate's conversational arm through an unstamped inner fn so no span is stamped twice.

Part of #496.

### DIFF
--- a/crates/aura/src/hitl/gate.rs
+++ b/crates/aura/src/hitl/gate.rs
@@ -283,12 +283,26 @@ mod tests {
                 self.spans().iter().any(|span| span.name == name)
             }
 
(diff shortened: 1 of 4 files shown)


dhable (approved):

justintime4tea (approved):

dhable (approved):

justintime4tea (approved):

Shearerbeard:

PR #572, opened 2026-08-22
Capture approver headers on approved webhook decisions

Shearerbeard:
The capture half of approver header forwarding: an approved webhook response yields the configured `tool_headers_from_response` values as validated overrides on the gate decision. Capture happens only on approval with a non-empty mapping and fails closed naming missing headers, never values. Both webhook arms race the round trip against the request cancel token, so a client disconnect resolves as cancelled with no override. A capture-configured route must use https at this point in the stack; part 4 relaxes that to a boot-time warning.

Part of #496.

### DIFF
--- a/crates/aura-config/src/config.rs
+++ b/crates/aura-config/src/config.rs
@@ -1258,6 +1258,59 @@ tool_headers_from_response = { "Authorization" = "x-approver-token" }
         }
     }
 
+    /// An explicitly empty map is the same feature-off state as an absent
+    /// one, and survives a serialize round trip as an absent key.
+    #[test]
+    fn hitl_webhook_tool_headers_empty_map_is_feature_off() {
+        let toml = r#"
+require_approval = ["kubectl_*"]
+
+[route]
+mode = "webhook"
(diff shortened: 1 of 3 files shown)


Shearerbeard:
Addressed in the full stack. The two `todo!()` panic seams are filled one PR up, in #573; at stack top (#544) zero `todo!()` remain, verified by grep across all five branches. The cleartext capture over http without HMAC warns once per route at startup (`warn_on_cleartext_capture`), adjacent to the existing boot-time HMAC check. Rejecting was ruled out because TLS termination ahead of the process (trusted gateway) is a legitimate topology; cleartext capture with mappings must stay usable. The one hard rule stays unchanged: HMAC over http fails closed at boot. This accepted negative consequence is recorded in docs/adr/2026-08-13-approver-identity-forwarding.md. Type-level comment placements are addressed at the stack top.

justintime4tea (dismissed):

dhable (dismissed):

dhable (dismissed):

justintime4tea (dismissed):

Shearerbeard:

PR #573, opened 2026-08-22
Apply approver header overrides to the gated MCP call

Shearerbeard:
The delivery half: captured overrides ride the rmcp request extension onto exactly one outbound request on the http-streamable and SSE transports. The override replaces the frozen client identity for that call only. A gated stdio call carrying overrides fails closed. Wire-level tests drive a recording loopback server through the real request-construction site.


justintime4tea (dismissed):

dhable (dismissed):

justintime4tea (dismissed):

dhable (dismissed):

Shearerbeard:

PR #574, opened 2026-08-22
Header forwarding audit span, integration suite, and docs

Shearerbeard:
Three closers for the feature: the cleartext-capture rule softens from reject to a boot-time warning since TLS can terminate outside the process; the gated call's `mcp.tool_call` span gets the applied override names, never values; and the first hitl integration suite lands behind its own feature flag with a make target against the mock fixture. The transport and route-scope decisions are recorded as an ADR.

Part of #496.

### DIFF
--- a/.makefiles/aura.mk
+++ b/.makefiles/aura.mk
@@ -262,3 +262,29 @@ test-integration-session-store-local: $(REPORT_DIR) ## Start an ephemeral Valkey
 	test_exit=$$?; \
 	docker stop $(VALKEY_TEST_NAME) >/dev/null 2>&1 || true; \
 	exit $$test_exit
+
+# --- HITL header-forwarding integration tests (needs only mock-mcp; each ---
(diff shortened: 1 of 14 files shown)


Shearerbeard:
Addressed at the stack top (#544). Behavior narration has been moved off the data definitions named in this review: type-level comments now carry identity statements only, and the runtime detail lives at the implementing functions and branches. This follows the comment convention in CLAUDE.md (document behavior where it lives; apply the drift test to every type-level comment), which is now enforced across the stack. The enum-variant comments on the test harness types flagged here were reduced to reply shape, with the serve-loop mechanics documented at serve_one_decision and the cleanup and log-draining behavior at the harness methods that implement it.

justintime4tea (dismissed):

justintime4tea (dismissed):

dhable (dismissed):

Shearerbeard:

PR #605, opened 2026-08-26
feat(governance): add core config and catalog envelope types 

### LINKED ISSUE #599: [FEATURE]: Send the MCP tool catalog to a webhook on demand
# [FEATURE]: Send the MCP tool catalog to a webhook on demand

## Summary

AURA can discover tools from its configured MCP servers and expose their schemas through `/aura/info?detail=tools`. A governance system may need this catalog during setup, before it has observed any governed tool requests.

Add an explicit command that discovers the current MCP tool catalog and sends one snapshot to a configured webhook. Possible command names discussed were `/governance init` and `/mcp sync`; the final name remains open.

The command must run only when requested. AURA should not send the catalog automatically when it starts or restarts. Users who need periodic synchronization can schedule the command outside AURA.

The catalog is observed metadata. Sending it must not create a policy, grant permission, or change how tool calls are governed.

## Requested behavior

- Provide an explicit command that discovers and sends the current MCP tool catalog.
- Send a versioned catalog event that is separate from an approval request.
- Treat a successful `2xx` response as acknowledgment. The receiver does not return an approval decision.
- Include:
  - AURA agent identifier and name
  - Configured MCP server name or namespace
  - Tool name, title, and description
  - Input schema, including available argument names and types
  - Output schema and tool annotations when available
- Use the tool names and schemas AURA exposes to the model, including any sanitization AURA applies.
- Exclude credentials, authorization headers, environment variables, command arguments, and other secrets.
- Preserve the difference between a server that returned no tools and a server whose catalog could not be collected.
- Report delivery or discovery failures without changing HITL enforcement for later tool calls.
- Allow the command to be run again when the operator wants to refresh the catalog.

## Acceptance criteria

- A user can explicitly trigger a catalog sync.
- The receiver gets the tools and argument schemas available at the time of the request.
- Tools with the same name from different MCP servers remain distinguishable.
- Multiple configured MCP servers can appear in one snapshot.
- An unavailable MCP server does not prevent available catalog entries from being sent.
- No configured secrets appear in the payload.
- The receiver can distinguish the catalog event from an approval request.
- A failed delivery does not bypass or disable later HITL checks.
- Starting or restarting AURA does not automatically send the catalog.
- Automated tests cover serialization, partial discovery, secret exclusion, delivery, and repeated explicit synchronization.

## Open implementation questions

- Should the catalog event use the existing HITL webhook configuration or a separate webhook?
- Should the command be named `/governance init`, `/mcp sync`, or something else?
- Should the payload be a standalone catalog event or part of a broader agent registration/card definition?

A separate event avoids adding tool schemas to normal approval requests. Reusing the existing endpoint may avoid duplicate webhook configuration. This issue does not choose between those approaches yet.

## Non-goals

- Automatic synchronization on startup or restart
- Built-in periodic synchronization
- Continuous catalog-change notifications
- Creating or changing governance policies
- Treating MCP annotations as authoritative permission data
- Reporting AURA health or providing a heartbeat
- Replacing `/aura/info?detail=tools`
- Building the lower-fidelity catalog that a receiver may infer from ordinary approval requests

## Related work

- #550 — `/aura/info` surface tool details
- #186 — Explore tool namespacing across MCP servers
- #494 — Send the requesting agent name on webhook approvals
- #495 — Send instance GUID on webhook structure

dhable:
Add governance configuration schema and catalog envelope builder for
MCP tool discovery sync:

- Config is done through the [governance.catalog] section in
aura-config
- Wire format types, webhook client and conversion is handled in
aura::goverence

Fixes https://github.com/mezmo/aura/issues/599

### DIFF
--- a/crates/aura/src/governance/client.rs
+++ b/crates/aura/src/governance/client.rs
@@ -1,0 +1,117 @@
+//! HTTP client for sending catalog snapshots to the governance webhook.
+
+use std::collections::HashMap;
+use std::time::Duration;
+
+use aura_config::CatalogWebhookConfig;
+use reqwest::header::{CONTENT_TYPE, HeaderMap};
+use tracing::info;
+
+use super::envelope::CatalogEnvelope;
+use crate::hitl::{SigningContext, WebhookHmac};
+
+/// Maximum time to wait for a TCP connection before failing.
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Errors that can occur while sending a catalog to the governance webhook.
+#[derive(Debug, thiserror::Error)]
+pub enum CatalogError {
+    #[error("catalog webhook transport error: {0}")]
+    Transport(String),
+    #[error("catalog webhook returned status {status}: {body}")]
+    BadStatus { status: u16, body: String },
+    #[error("catalog webhook HMAC signing failed: {0}")]
+    Signing(String),
+    #[error("catalog serialization failed: {0}")]
+    Serialization(String),
+}
+
+/// HTTP client for the governance catalog webhook.
+pub struct CatalogClient {
+    client: reqwest::Client,
+    url: String,
+    timeout: Duration,
+    headers: HeaderMap,
+    hmac: Option<WebhookHmac>,
+}
+
+impl CatalogClient {
+    /// Build a catalog client from config.
+    ///
+    /// `req_headers` is the inbound request headers for `headers_from_request`
+    /// resolution. Pass `None` in CLI standalone mode.
+    pub fn from_config(
+        config: &CatalogWebhookConfig,
+        req_headers: Option<&HashMap<String, String>>,
+    ) -> Result<Self, CatalogError> {
+        let client = reqwest::Client::builder()
+            .connect_timeout(CONNECT_TIMEOUT)
+            .build()
+            .expect("reqwest client builder only fails on TLS backend init");
+
+        let headers = crate::webhook_utils::resolve_headers(
+            &config.headers,
+            &config.headers_from_request,
+            req_headers,
+        );
+        let hmac = crate::webhook_utils::build_hmac_from_secret(
+            config.hmac.as_ref().map(|c| c.secret.as_str()),
+        )
+        .map_err(|e| CatalogError::Signing(e.to_string()))?;
+        let timeout = Duration::from_secs(config.timeout_secs);
+
+        Ok(Self {
+            client,
+            url: config.url.as_str().to_string(),
+            timeout,
+            headers,
+            hmac,
+        })
+    }
+
+    /// Send a catalog envelope to the governance webhook.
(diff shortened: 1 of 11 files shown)


crates/aura/src/governance/client.rs
```diff
@@ -0,0 +1,170 @@
...
+            });
+        }
+
+        info!(url = %self.url, event_id = %envelope.event_id, "Catalog sync successful");
+        Ok(())
+    }
+}
+
+/// Resolve webhook headers from static config and request header mappings.
+fn resolve_headers(
```
justintime4tea: **`resolve_headers` duplicates `hitl::route::resolve_webhook_headers`**

`crates/aura/src/governance/client.rs:113` is a near-verbatim copy of `resolve_webhook_headers` at `crates/aura/src/hitl/route.rs:384`. A line-by-line diff shows the bodies
differ only in the function name and two log strings:

| | `hitl/route.rs:384` | `governance/client.rs:113` |
|---|---|---|
| name | `resolve_webhook_headers` | `resolve_headers` |
| info log | `"Webhook route: resolved …"` | `"Catalog webhook: resolved …"` |
| warn log | `"Skipping invalid webhook header …"` | `"Skipping invalid catalog webhook header …"` |

Both do the same three steps: lowercase the static header keys, overlay `headers_from_request` via `rig_builder::apply_request_header_mappings` (already shared), then convert
to a `HeaderMap`, skipping invalid names/values with a warning. The config shapes line up too — `CatalogWebhookConfig` and the HITL route config both carry `headers` +
`headers_from_request` as `HashMap<String, String>`.

Beyond the ~35 duplicated lines, the copy doesn't inherit the nine unit tests `resolve_webhook_headers` already has (`hitl/route.rs:1612`–`1830`: static fallback, request
override, case handling, invalid-header skipping).

Suggested fix — one shared function parameterized by a log label:

```rust
pub(crate) fn resolve_outbound_headers(
    context: &str,                       // "Webhook route" / "Catalog webhook"
    static_headers: &HashMap<String, String>,
    headers_from_request: &HashMap<String, String>,
    req_headers: Option<&HashMap<String, String>>,
) -> HeaderMap
```

Smallest change is to put it beside `apply_request_header_mappings` in `rig_builder.rs`, since both call sites already reach in there; a small `outbound_headers` module would
sit better semantically than growing `rig_builder`.

Worth noting one layer down: the `HashMap` → `HeaderMap` conversion loop by itself is repeated twice more, in `mcp_sse.rs:112` and `mcp_streamable_http.rs:147` (no
`headers_from_request` resolution in those). A lower-level `to_header_map(&HashMap, context)` would cover all four sites, with `resolve_outbound_headers` layered on top

dhable: Resolved by extracting some common webhook code into a new module. There's probably room for further refactoring on the webhook code but that can wait.

Shearerbeard (approved):
Approved - @dhable I'll leave it up to your discretion if being explicit about the AgentEntry id new type is worth anything going forward to the webserver/runtime split work

PR #617, opened 2026-08-31
fix: surface failed tasks' artifacts and preserve worker confidence

adrianfurlong:
Two defects that combined to make an orchestration run report the opposite of what it had found: a worker retrieved the evidence answering the user's question, failed before submitting, and the final answer asserted a later empty result as established fact.

## Failed tasks' artifacts were invisible to the coordinator

A worker that times out or exhausts its depth budget can still have completed its tool calls successfully. That output is written to artifacts, and `load_tool_traces_for_plan` loads trace records for every task regardless of state — but `build_continuation_prompt` rendered artifact lines only in the Complete branch. The coordinator saw `Task N → failed [timeout]` with no indication that the results existed on disk, so the next plan repeated the work from scratch.

Adds the `render_artifact_lines` call to the Failed branch, mirroring the Complete branch. The existing `RESULT_FORWARDING` guidance already tells the coordinator to pass artifact filenames on to the next iteration, so no prompt change is needed for this half.

## Low-confidence results were reported as fact

Worker confidence reaches the coordinator's context (`submit_result` → persistence → `(confidence: low)` in COMPLETED TASKS), but the `respond_directly` synthesis rules only required findings be inlined concretely — nothing required uncertainty to survive. A worker's hedged, low-confidence disjunction was delivered to the user as an unqualified fact and turned into a recommended action.

Adds two rules to `continuation_prompt.md`: carry the worker's confidence qualifier across, and treat an empty result as "the query returned no matches" rather than as evidence that the underlying thing does not exist.

Prompt rules only bind the path that runs the coordinator. `build_raw_task_results` ships task results straight to the user when the post-execute coordinator call errors or the replan/time budget is spent, bypassing synthesis entirely — so it now tags each completed task with its confidence at render time.

The changes need to land together — forwarding more findings into a synthesis step that strips uncertainty would raise the confidence of whatever gets through.

## Verification

`cargo test -p aura --lib` — 1088 passed, 0 failed. `cargo fmt --check` clean, no new clippy warnings.

Both regression tests were confirmed to catch their bug: each fails with its fix reverted and passes with it restored.

### DIFF
--- a/crates/aura/src/orchestration/orchestrator.rs
+++ b/crates/aura/src/orchestration/orchestrator.rs
@@ -4554,9 +4554,18 @@ Assign tasks to the worker whose tools best match the required operations."#,
         for t in &plan.tasks {
             match &t.state {
                 TaskState::Complete { result } => {
+                    // This text reaches the user unmediated — the coordinator
+                    // never ran, so the synthesis rules that preserve a
+                    // worker's confidence do not apply. Tag it here instead,
+                    // or an uncertain finding is shipped as an unqualified one.
+                    let confidence = t
+                        .structured_output
+                        .as_ref()
+                        .map(|so| format!(" (confidence: {})", so.confidence))
+                        .unwrap_or_default();
                     out.push_str(&format!(
-                        "## Task {}: {}\n\n{}\n\n",
-                        t.id, t.description, result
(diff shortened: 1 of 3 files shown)


Shearerbeard:
@adrianfurlong if you still want to get this one in we can swap the target to nightly

Issue #618, opened 2026-08-31
[FEATURE]:  Agent event schema and broker adapter


jakedipity:
### Summary

Agents publish straight into four process-global brokers keyed by request id, so observing an agent means being an HTTP request. Give agents an event vocabulary of their own, plus an adapter that republishes it onto those brokers, so producers can move over without any consumer changing.

### Goals

- An event type covering what an agent surfaces today: tool lifecycle, MCP progress, reasoning and response content, the usage reports, HITL approvals, MCP status.
- Independent of the wire format's serialization constraints, and free of request-scoped correlation, which belongs to whatever observes the agent rather than to the agent.
- SSE, A2A, and CLI output unchanged while the new schema exists alongside the current path.
- Demonstrable parity — a regression test that fails when translation drops or reshapes a field, exercising the real broker registry rather than stubbed channels.
- Adoption incremental and reversible behind a flag, with no producer obliged to move.

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

justintime4tea:
Landed in #619, merged to `nightly` on 2026-09-17 as 6ea85b44.

`aura_events::agent` defines the event vocabulary and `aura::agent_events` republishes it onto the request-scoped brokers, so consumers are unaffected. Parity is covered by `crates/aura-web-server/tests/agent_event_differential.rs`, which drives both paths through the real broker registry and asserts byte-identical SSE.

Two goals landed with caveats worth recording:

- *SSE/A2A/CLI output unchanged* — two frames changed shape deliberately: `aura.tool_usage` carries the emitting agent's `AgentContext`, and `aura.progress` builds its fallback message from `Progress`'s `Display`. `AgentContext::agent_id` deserializes with a default so payloads predating the field still parse.
- *Adoption reversible behind a flag* — `ENV_AGENT_EVENTS`/`agent_events_enabled()` reading `AURA_AGENT_EVENTS` shipped in 6ea85b44 and was removed in f3bdbd61 (#620), where the single-agent producers moved onto `agent_events::emit` unconditionally. There is no longer a runtime seam to fall back to the broker-only path.

Closing manually: the commits use `Ref:` footers, which link but do not close.

PR #619, opened 2026-08-31
feat(events): add agent event schema and broker adapter

jakedipity:
Adds aura_events::agent, the event vocabulary a running agent emits, and aura::agent_events, which republishes it onto the request-scoped brokers so consumers are unaffected. A differential test asserts both paths produce identical SSE.

The schema is typed with domain value types rather than bare primitives: ToolCallId, ToolName, TokenCount, TokenUsage, and Progress. The tool_event_broker's ToolCallId/ToolName aliases resolve to the newtypes, so publish_tool_requested and publish_tool_start no longer take interchangeable strings positionally, publish_tool_usage takes one TokenUsage instead of three bare u64s, and ProgressNotification carries the current/total pair as a single value that owns percent(). Scratchpad token counts move from usize to u64 with every other count.

No producer emits the schema yet; agent_events_enabled reads AURA_AGENT_EVENTS and is the seam producers will gate on.

Ref: https://github.com/mezmo/aura/issues/618

### DIFF
--- a/crates/aura-events/src/agent.rs
+++ b/crates/aura-events/src/agent.rs
@@ -1,0 +1,269 @@
+//! The agent-produced event vocabulary.
+//!
+//! [`AgentEvent`] is what a running agent emits. Observers — an SSE producer, an
+//! A2A status bridge, an OTel exporter — consume this stream and project it into
+//! whatever shape they serve. Contrast [`crate::AuraStreamEvent`], which is the
+//! HTTP *wire* form of one such projection.
+//!
+//! Two properties distinguish this schema from the wire schema:
+//!
+//! - **Internally tagged.** [`crate::AuraStreamEvent`] is `#[serde(untagged)]`,
+//!   so its variant order is load-bearing during deserialization. This enum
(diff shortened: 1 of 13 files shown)


crates/aura-events/src/agent.rs
```diff
@@ -0,0 +1,233 @@
...
+        content: String,
+    },
+
+    Reasoning {
+        content: String,
+    },
+
+    /// The model's decision to call a tool, ahead of any execution.
+    ToolRequested {
+        tool_id: String,
```
justintime4tea, line 77: In some of the other PRs we have fields on some enums like `tool_call_id`, what are your thoughts on changing this to `tool_call_id` to keep it all canonical? There are a few more occurrences in other places that have this same dissonance.

jakedipity: Renamed to `tool_call_id` to be consistent

justintime4tea (dismissed):

Shearerbeard (approved):

Issue #623, opened 2026-08-31
[FEATURE]:  Orchestration emits the agent event schema


jakedipity:
### Summary

Orchestration already sends its events through a single stream, in band with response content, but in an event type of its own. Separately, the frames those events become are declared twice — once where the server emits them, once where the CLI parses them — with nothing keeping the two in agreement. Put orchestration on the agent event schema and reduce the frame declaration to one.

### Goals

- Orchestration events expressed in the same vocabulary as single-agent events, so a consumer no longer branches on which mode produced a run.
- One declaration of the orchestration SSE frames, shared by the side that emits and the side that parses, so a field cannot disagree between them.
- The frames themselves unchanged, since clients depend on their names and shapes.
- Variants that nothing emits removed rather than carried.
- Both consumers moved together — SSE and A2A read the same stream item, so neither can be left behind.

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

justintime4tea:
Landed in #632, merged to `nightly` as two commits:

- f227357e `refactor(events): define orchestration SSE frames once` — collapses the duplicate `OrchestrationStreamEvent` declaration. It had already diverged: the CLI-parsed copy in `aura-events` carried `PhaseStarted`/`PhaseCompleted` variants that nothing emitted.
- 30c5f840 `feat(events): orchestration emits the agent event schema` — `OrchestratorEvent`'s variants move onto `AgentEventPayload`, so orchestrated and single-agent runs share one vocabulary and `StreamItem` carries that payload directly.

Emitted frames are unchanged; the projection reads the same field sets under a different type, so SSE and A2A moved together.

Closing manually: the commits use `Ref:` footers, which link but do not close.

Issue #627, opened 2026-08-31
[FEATURE]:  Remove the request-scoped broker path


jakedipity:
### Summary

With every producer and consumer on the agent event schema, the machinery that carried events before it becomes dead weight: the request-scoped brokers, the old streaming trait, the orchestration translation, the compatibility adapter, and the flag that chose between paths. Remove them.

The SSE frame types are not part of that. They are what one observer projects the schema into, and the agent layer only needs to stop owning them.

### Goals

- The process-global brokers keyed by request id gone, along with the subscribe and unsubscribe lifecycle every request currently pays.
- Orchestration events reaching an observer the way every other event does, rather than riding the item stream and being translated separately at the far end.
- The agent layer holding no SSE frame type, so the frames belong to the producer that formats them and the consumers that parse them.
- The adapter and the flag gone, leaving one path rather than two.
- No behavior change, and no test weakened to allow the removal — anything that breaks here was either untested or genuinely still in use.

### Additional Context

The orchestration and single-agent SSE frame enums stay separate. Merging them would change no bytes, but `AuraStreamEvent` is `#[serde(untagged)]` and the CLI deserializes it, so adding variants changes which variant an existing payload matches. A shadowed variant fails the consumer's pattern match and drops the event with no error. That is a live risk to a consumer in exchange for one type instead of two, both of which already format through the same helper and live in the same crate.

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #624, opened 2026-08-31
[FEATURE]:  HITL approvals move onto the agent event schema


jakedipity:
### Summary

The decision route announces an approval's lifecycle by publishing straight to the approval broker, the last producer still reaching a broker on its own. Move those sites onto the agent event schema so approvals travel the same way as tool lifecycle, MCP progress, and usage.

### Goals

- Approval lifecycle carried in the same vocabulary as the rest of an agent's events, rather than on a private channel.
- The publish sites going through the same entry point every other broker-backed producer uses.
- Attended and unattended routes reaching the same outcomes they do today, including timeout and cancellation.
- Emitted events unchanged, demonstrably: the existing approval tests subscribe to the broker and assert on what arrives, so a lost or reshaped field fails them.

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

justintime4tea:
This has landed in nightly [per this PR](https://github.com/mezmo/aura/pull/634)

Issue #620, opened 2026-08-31
[FEATURE]:  Single-agent producers emit the agent event schema


jakedipity:
### Summary

The single-agent emission sites each call a broker directly, so an agent's output has no shared vocabulary and no single place it passes through. Move them onto the agent event schema, publishing through one entry point that translates to the brokers, so consumers stay unchanged and later work has a seam to attach to.

### Goals

- Every single-agent emission — tool lifecycle, MCP progress, tool usage — expressed as an agent event rather than a broker-specific call.
- One entry point every emission passes through, so a real event stream can be introduced later without revisiting the emission sites.
- SSE output unchanged, demonstrably: a test drives the same sequence through the broker calls and through the new path and compares the frames, exercising the real broker registry rather than stubbed channels.
- Integration suites run against the emission sites, which have no unit coverage of their own.

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

justintime4tea:
Landed in #621, merged to `nightly` as f3bdbd61.

All four single-agent emission sites — the hook's `tool_requested` and `tool_usage`, MCP `tool_start`, and MCP progress — build `AgentEvent`s and publish through `agent_events::emit`, which routes each payload to the broker it belongs to. An MCP client now holds the request it serves and the agent it serves it for as one `CallContext`, so a producer reads both under a single lock and attributes an event only to the request that named it.

Differential coverage extends `agent_event_differential.rs` to drive the same sequence through the broker calls and through `emit`, asserting the SSE frames match.

Note for #626/#627: this commit also removed the `AURA_AGENT_EVENTS` flag added in #618, so the two paths are no longer selectable at runtime.

Closing manually: the commits use `Ref:` footers, which link but do not close.

PR #632, opened 2026-09-01
refactor(events): define orchestration SSE frames once

jakedipity:
refactor(events): define orchestration SSE frames once
The producer and consumer each declared their own copy of OrchestrationStreamEvent, and they disagreed: plan_created did not parse, and iteration_complete bound to synthesizing. aura-events now holds the single definition, matching what the server emits, and the aura copy re-exports it. RoutingMode and IterationTimings move with it.

Emitted frames are unchanged; only the parsing side is corrected.

Ref: #623

---

feat(events): orchestration emits the agent event schema
Orchestration carried its own OrchestratorEvent enum between the orchestrator and the SSE projection. Its variants move onto AgentEventPayload, so orchestrated and single-agent runs share one vocabulary, and StreamItem carries that payload directly.

Emitted frames are unchanged; the projection reads the same field sets under a different type.

Ref: #623


crates/aura-events/src/agent.rs
```diff
@@ -142,6 +139,84 @@ pub enum AgentEventPayload {
...
+        will_replan: bool,
+        reasoning: String,
+        gaps: Vec<String>,
+        timings: IterationTimings,
+    },
+
+    ReplanStarted {
+        iteration: usize,
+        /// `"coordinator"` or `"failure"`.
+        trigger: String,
```
justintime4tea, line 216: Went down a bit of a rabbit hole on this one, so sharing what I found and a couple of ideas — curious what you think.

Where my head is at: the *reason* for a replan currently only makes it into the coordinator's continuation prompt. With #562 pushing toward agent events being the thing observers (SSE, A2A, OTEL) project from, it feels like a reason the agent already knows ought to live in the event too, not just in prompt text. This variant seemed like the natural place to start, and once I dug in, the current field turned out to be a bit of a fossil.

**The doc is describing values that don't exist anymore.** Neither `"coordinator"` nor `"failure"` is ever emitted — there's one emitter (`orchestrator.rs:4109`) with one caller (`:4792`), and the value is always `"post_execute_create_plan"`. Looks like the failure-driven branch went away in 6a745109 when it got folded into the unified post-execute call, and the doc just never caught up (`trigger_replan`'s own doc still says "coordinator-routed, failure-driven" too). And as far as I can tell nothing reads it — the CLI has no `replan_started` handler, and the server `debug!`s it and passes it along. So it's a constant string with a comment about two states that aren't there. Given #623's "variants nothing emits get removed rather than carried," it felt worth flagging even though it's a docstring rather than a variant.

**What I'd love to see in this PR, if you're up for it:** `trigger` as a small typed enum — one variant for now, correctly documented. I *think* this stays inside the "frames unchanged" line: `AgentEventPayload` isn't the wire type, the projection at `handlers.rs:1343` is already per-variant, and `replan_started` takes `impl Into<String>`, so an `as_str()` there should leave the bytes identical. Totally open to just dropping the field instead if you'd rather — it's constant and unread. Mostly I'd just like to avoid a `String` that reads like it varies when it doesn't.

Something like this is what I had in mind — named after the wire string so `as_str()` and `rename_all` agree for free, and it lines up with what the span already records on that branch (`orchestration.post_execute_decision = "create_plan"`, `:4710`):

```rust
#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
#[serde(rename_all = "snake_case")]
pub enum ReplanTrigger {
    /// The post-execute coordinator call routed to `create_plan`.
    PostExecuteCreatePlan,
}

impl ReplanTrigger {
    pub fn as_str(self) -> &'static str {
        match self {
            Self::PostExecuteCreatePlan => "post_execute_create_plan",
        }
    }
}
```

If the failure-driven path ever comes back it gets its own variant and its own string, and the projection stays `trigger.as_str()`. Does that match your read, or am I missing a reason to keep it stringly?

**Something for later - I can create a new issue just not sure where:** the failure reason *is* actually reachable at the emission site — `FailureSummary` (`:4553`) goes into `post_execute_ctx` (`:4590`), which is only borrowed after that, so it's alive at the call site; `trigger_replan` just gets `None` (`:4797`). But I'm not sure `replan_started` is even the right home for it. `IterationComplete` already has `reasoning` + `gaps` — the exact two fields of `FailureSummary` — on both the payload and the frame, and they're hardcoded empty on the replan path (`:4774`), only populated on the two terminal paths (`:4739`, `:4842`) that never replan. **Filling them from `post_execute_ctx.failure_summary` would be the same information with no new field and no shape change. I was thinking of filing that against #576 the way #612 did for the tool lifecycle — probably easiest once the SSE producer is in so it's only wired once.** Not blocking anything here; just wanted to check whether that sequencing makes sense to you before I open it.

(And to answer my own original question — the coordinator's stated reason isn't actually invisible today; `PlanCreated.routing_rationale` fires right before this. What's missing is the *failure* side, which is the part I'd expect an observer to actually act on.)

What are your thoughts, @jakedipity, @Shearerbeard, @dhable?

dhable (dismissed):

justintime4tea (dismissed):

Shearerbeard (changes requested):
I don't know if this belongs in this exact PR but as part of this work what i would love to see is a collapse between "coordinator did this" and "single agent did this". For sure we have some orchestration specific events around plan an execute however I think it was a mistake on my part to have so much divergence between the single agent concerns and top level coordinator concerns - sure there are extra tools if workers are configured but we should be building our schema around unification to make that work easier. Maintaining single agent and orchestration as separate braches of execution has caused a lot of extra work. If we consolidate now on the schema lines we can rip the single agent execution path out completely when we cut over to agent driver under the hood which actually doesn't have any distinction. Orchestration truly is a top level agent with tools for managing workers.

jakedipity:
> I don't know if this belongs in this exact PR but as part of this work what i would love to see is a collapse between "coordinator did this" and "single agent did this". For sure we have some orchestration specific events around plan an execute however I think it was a mistake on my part to have so much divergence between the single agent concerns and top level coordinator concerns - sure there are extra tools if workers are configured but we should be building our schema around unification to make that work easier. Maintaining single agent and orchestration as separate braches of execution has caused a lot of extra work. If we consolidate now on the schema lines we can rip the single agent execution path out completely when we cut over to agent driver under the hood which actually doesn't have any distinction. Orchestration truly is a top level agent with tools for managing workers.

Yeah, agreed, and it's the plan. #575's non-goals already say workers and the plan DAG are just tools. This PR is #618 + #623 under it; single-agent producers move over in #620, approvals in #624, and #627 is where the single-agent execution path actually gets deleted.

Shearerbeard (approved):

Issue #626, opened 2026-08-31
[FEATURE]:  Make the agent event schema the default path


jakedipity:
### Summary

Once every producer emits the schema, make it the only path a run's events take. Producers currently emit it alongside the brokers it duplicates, so nothing yet depends on it.

The coupling this removes is visible in the standalone CLI: with no HTTP involved it still builds server state in process, calls the web-server handlers, and re-parses the resulting SSE bytes, because there is no way to consume an agent without serializing to SSE first. Making a run's events readable from its handle is what lets a consumer stop doing that.

Consumers reading the stream directly is also what makes the request-scoped brokers redundant, so retiring them and the identity they are keyed by belongs here rather than earlier.

The SSE handler keeps projecting onto its existing wire schema so behavior holds. Replacing that projection with a producer built on the event stream is #576, the A2A equivalent is #577, and the CLI moves off its in-process SSE round trip with them.

### Goals

- A run's events reaching its observer directly, rather than through process-global brokers.
- The run's stream carrying response content as well, so an observer sees what an agent is saying and not only what it is doing. Clients read `StreamItem` for text and tool calls today, which is what keeps them on the old path.
- The per-request registries retired, since a run's handle is what holds its channels once consumers read from it.
- Run identity independent of HTTP, so a request id is no longer what a run is keyed by. The registries are what key on it, so they go together.
- Cancellation reaching MCP through the run's handle rather than a separate process-global registry.

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #634, opened 2026-09-01
feat(events): approvals emit the agent event schema

jakedipity:
Every approval publish site builds an AgentEvent and goes through agent_events::emit. That covers the decision route's five, park mode's pair in the gate, and the two sweeps that complete a parked ticket on cancellation. hitl::events hands each producer a finished event, so an approval's payload and its envelope come from one scope.

Naming a worker's approval needs the same parent id the orchestrator stamps on that worker's other events, so that id is a constant in aura-events rather than a literal in each module.

A task with no assigned worker has no name in that scope, so its approval is attributed to the coordinator running it, while the orchestrator names the same agent by its own id. The broker payload carries no agent field, so neither value reaches a client.

Ref: #624

### DIFF
--- a/crates/aura-web-server/tests/agent_event_differential.rs
+++ b/crates/aura-web-server/tests/agent_event_differential.rs
@@ -1,0 +1,390 @@
+//! Proves the agent event schema reaches consumers unchanged.
+//!
+//! Each case drives `process_sse_stream_full` twice over the same logical
+//! sequence — once publishing to the request-scoped brokers directly, once
+//! publishing [`AgentEvent`]s through [`aura::agent_events`] — and asserts the
+//! SSE frames are byte-identical. Only the producer side differs; both runs
+//! subscribe the way `handlers::stream_chat_completion` does, so the broker
+//! registry and its request-id routing are under test rather than stubbed.
+//!
+//! A variant that loses a field in translation fails here.
+
+use std::sync::Arc;
+use std::time::Duration;
+
+use aura::agent_events::{Routed, publish_to_brokers};
+use aura::tool_event_broker::publish_tool_requested;
+use aura::{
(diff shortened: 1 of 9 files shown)


justintime4tea (dismissed):

crates/aura-web-server/tests/agent_event_differential.rs
Shearerbeard, line 82: This was my only big feedback on this - I would probably just move to a single event subscription and have the client be in charge of matching or ignoring events as needed. To much juggling of micro subscriptions and we can homogenize a lot of the event handling and forwarding if its the same general event shape based on the producer. Is your event stream meant to be multi tenant or per agent?

crates/aura/src/agent_events.rs
Shearerbeard, line 41: Is this the "ACK" piece?

Shearerbeard (changes requested):
A few minor pieces of feedback to consider

crates/aura-web-server/tests/agent_event_differential.rs
jakedipity: It's per run, not per agent. The brokers key on `request_id` so a coordinator and its workers share the channels, and `AgentContext` on the envelope is what tells them apart. Agreed on collapsing to one subscription, that's #627. The four in that helper are only there so the test can run both paths and compare bytes. They go when the brokers do.

crates/aura/src/agent_events.rs
jakedipity: Not an ack, it's just the enqueue result. `Delivered` means a subscriber was registered and the send succeeded. `emit` already logs `NoSubscriber` and most producers just drop the value.

A real ack would be worth it for HITL though. If `approval_pending` gets dropped the approver never sees the prompt and the call fails closed at the deadline. That's #626, when the broker hop becomes a real stream.

justintime4tea (approved):

Shearerbeard (approved):

PR #635, opened 2026-09-02
feat(session-store): file-backed approval store

Shearerbeard:
## What

A file-backed `ApprovalStore` backend: one JSON file per decision id (`{root}/tickets/{id}.json`, `{root}/decisions/{id}.json`), reusing `ParkedApprovalRecord`/`DecisionRecord` as the on-disk shape. Selected by `AURA_SESSION_STORE=file` with a required `AURA_SESSION_STORE_PATH`; the A2A task store and event bus stay in memory (single-pod, single-writer model).

## Contract (what park mode keys on)

- `resolve` refuses past the ticket's `expires_at`, uniformly with an unknown id - enforced on BOTH the file and memory backends, so the contract holds on whichever is configured.
- `get` returns the ticket before and after the decision, and `decision` returns the recorded decision, both until `remove` (retention until removed, not a TTL margin).
- `resolve` moves the ticket into the decision file rather than deleting it; at-most-once via `File::create_new` on the decision file (`AlreadyExists` reads as `NotFound`).
- `cancel_request` removes undecided tickets by owner id; decision ids are validated as canonical UUIDs at every path-building site, so nothing escapes the store root.
- Ticket publication is temp-file + same-directory rename; the create_new-to-sync crash window is fail-closed and documented (module doc) with a one-file operator recovery.

## Testing

- The backend-agnostic approval battery is factored out of the redis integration test into `tests/common` and run against BOTH the file and memory backends without Docker, plus retention, expiry, resolve-moves-ticket, cancel-by-owner, and restart-durability cases (register, drop the store, reopen, resolve, read the decision).
- `cargo test -p aura -p aura-web-server` and `cargo clippy --workspace --all-targets -- -D warnings` clean; live restart check passed (register, kill -9, restart, resolve 204, decision file on disk).

## Review trail

Two fresh-context reviews returned explicit PASS: a granular Rust review (at-most-once, expiry on both backends, path containment) and a feature-scope frontier review (poll-shape coherence with the planned run-resource resume UX, durability boundary). The frontier round's three doc-coherence findings were fixed in `e3195f45` and re-reviewed clean. Full findings ledgers live on the board cards.


Shearerbeard:
Closing in favor of #643: nightly is now the primary PR target, and the retarget path causes odd GitHub behavior, so the store ships as a fresh PR against nightly carrying the v2 fix round (blocking-pool restructure, resolve commit-point semantics, comment pass).

Issue #630, opened 2026-09-01
support llm prompt caching with cache-usage visibility


teriyakichild:
Track support for provider prompt caching and end-to-end cache-usage visibility:

- `prompt_caching` config flag for Anthropic (`cache_control` breakpoints via the rig fork) and Bedrock (`cachePoint` blocks via the fork's vendored rig-bedrock); OpenAI caching is automatic server-side.
- Cache read/creation token counts surfaced through `UsageState`, the `aura.usage` SSE event, the CLI status line, and OpenInference span attributes (`llm.token_count.prompt_details.cache_read`/`cache_write`) for Phoenix.
- Fork side landed in mezmo/rig#12.

Verified live against Anthropic-format mocks, an OpenAI-format mock, and real Bedrock (`us.anthropic.claude-sonnet-5`): cold request writes the prefix to cache, warm request reads it back at ~10% input price, with the split visible in Phoenix's token breakdown.

PR #639, opened 2026-09-02
test(web-server): fix lib test compile after cache_usage addition

teriyakichild:
The aura-web-server lib test target has not compiled on nightly since `FinalResponseInfo` gained the `cache_usage` field: the streaming handler tests' `final_item` helper still built the struct without it. Adds `cache_usage: None`, which matches what those tests simulated before the field existed.

`cargo test -p aura-web-server --lib`: 167 passed.

Refs: #630


teriyakichild:
Folding this into #638 instead.

PR #658, opened 2026-09-03
feat(orchestration): reify and continue parked runs

Shearerbeard:
## Summary

Stacks on #656. Four slices plus review fixes, park mode stays default-off throughout:

- **86cd0d1f** retires the ticket vocabulary in the storage layer: `tickets/` becomes `approvals/`, `ResolvedEntry.ticket` becomes `approval` (no consumers of the old names yet - park mode is default-off and the park line consumes the store only through the trait API).
- **b2bcf88c** adds the recorded-decisions consult seam: `RecordedDecisions`/`CallKey` in the park module, consumed at the gate immediately after glob matching when armed via `with_recorded_decisions` (`None` on the live path, zero behavior change). A hit maps through the existing approval mapping so denial feedback is identical to the live path; a miss while the task's continuation is in flight fails closed; otherwise the call re-parks. A drop guard clears the strict entry on every exit path.
- **8ede9494** adds the stubbed-model test rig (scripted turns, per-turn request capture, a stall hook for deterministic race windows) and an ungated smoke test pinning model fidelity end to end.
- **c9f85791** lands reify: `load_recorded_decisions` builds the consult set from the store's approvals and recorded decisions (never the checkpoint copy), `ResumingDocumentHandle` tombstones each call before its invocation, `execute_task` takes the optional continuation and resumes the normal loop after sentinel replacement, and consumed decisions leave the store with the task. `RehydrateError` maps to the resume endpoint's condition rows, including the parked row carrying every outstanding id.
- **4f3dc809 / fa6ccf89** carry the Gate A fix round (parked row, scope-borrow refusals with negative tests) .

A `#[cfg(test)]` seam injects scripted models through a `ProviderAgent::Scripted` variant so `execute_task` is drivable end to end; with no override installed the build is unchanged.

## Proof coverage

Orphan dual-trigger (depth exhaustion and the mid-turn stream break - on the pinned rig, turn-start exhaustion ends blocked, not orphaned), the full park -> drop -> rehydrate -> resume loop with zero human input and replan state round-tripping, denial parity, concurrent tombstone appends, exactly-once ordered consumption, strict isolation between sibling tasks, and the store correcting a stale document.

## Test plan

`cargo test -p aura` (1156 passed), `cargo test -p aura-web-server` (167 passed), `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo fmt --all --check` clean.

Closes part of #271

### DIFF
--- a/crates/aura-web-server/tests/file_session_store_test.rs
+++ b/crates/aura-web-server/tests/file_session_store_test.rs
@@ -1,8 +1,8 @@
 //! Conformance and contract tests for the file-backed approval store
 //! (`AURA_SESSION_STORE=file`): the shared backend-agnostic battery plus the
 //! §2.5 contract points specific to durable, retention-until-remove storage —
-//! expiry refusal, retention past the decision, the ticket's move into the
-//! decision file, owner-scoped cancel of undecided tickets, and survival of a
+//! expiry refusal, retention past the decision, the approval's move into the
+//! decision file, owner-scoped cancel of undecided approvals, and survival of a
 //! process restart. The same battery runs against the memory backend to pin
 //! the uniform contract. No Docker: every test gets its own tempdir.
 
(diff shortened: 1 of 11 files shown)


Shearerbeard:
Closing in favor of a two-PR split (same commits, no rewrite): #660 carries the vocabulary homogenization + the recorded-decisions consult seam (~600 lines, zero behavior change), and the follow-up carries the rig + reify + proofs. Review the smaller slices in stack order.

Issue #495, opened 2026-07-30
[FEATURE]:  Send instance guid on webhook structure


Shearerbeard:
### Summary

Make instance guid an argument so we know pod origin on HITL

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Shearerbeard:
Again this is more governance transparency @dhable - it might be worth generating an identifying guid for what instance of aura a HITL interrupt came from

dhable:
The easiest method for generating a guid would just be to generate a hash based off of the current timestamp at boot - e.g. UUID v4. We might also want something unique but stable across restarts to make it easier to correlate HITL requests across a longer time frame. We could build a guid using the following:

```
instance_guid = uuidv5(aura_ns, instance_fingerprint)
  where
    aura_ns is something like "mezmo.com/aura:v1" and used to identify the 
    instance_fingerprint is both environment and config seed to identify where and what is running
```
Since UUIDv5 is a stable hash, we should be able to track the same instance across restarts if we pick sane fingerprint values. These values are a fixed pattern concatenated bytes, prefixed by the number of bytes that follow to avoid ambiguity with multiple byte fields. Thus  `name="ab", alias="c"` → `2:ab1:c` vs `name="a", alias="bc"` → `1:a2:b`

To compute a unique environment seed (i.e. where the aura process is running), aura would use the following list in order:
* `AURA_INSTANCE_ID` environment variable value. This allows for overriding the remaining values if desired.
* k8s downward API (`"{POD_NAMESPACE}:{POD_NAME}"`)
* `machine_uid` crate to determine a HW identifier (`/etc/machine-id` for Linux, `gethostuuid()` for Mac, `MachineGuid` registry value for Windows)
* If no value can be determined, then generate a random UUIDv4 value and emit a warning log with the suggestion to set `AURA_INSTANCE_ID`.

To compute a unique config seed (i.e. what the process is running), aura would use the following list in order:
* `[agent].instance_seed` (new property) from the config file
* Agent name and alias fields (`"{[agent].name}{[agent].alias}"`). Since the agent name is required in the config, we should always have a value available. Does it make sense to guard against an empty string at this point?

This identify value would then be exposed through the the web API, the session info, etc. for discovery by a consumer.

[...]

PR #659, opened 2026-09-03
doc(adr): propose polling for parked webhook decisions

teriyakichild:
## Summary

Proposed ADR amending the park/reify ADR (decisions 5, 8, 10, 12, 13) and Route A in the HITL architecture ADR. Targets `nightly`; the reify implementation is in #658.

**Problem.** A hosted governance receiver cannot reach an aura agent inside a customer's network. Once Route A can park, the decision ingress (`POST /v1/approvals/{decision_id}`) is not a usable wake path for that deployment: it requires the receiver to dial in.

**Decision.** Route A becomes parkable via a `202 {"status":"pending"}` response on the existing signed response leg, and aura fetches the decision itself:

- **Batched long-poll from aura** to a configured `poll_url`, carrying the outstanding decision ids and an instance id. The receiver holds the request up to a wait budget and returns early when any decision resolves. Every leg is outbound; the receiver never needs a route to the agent.
- **Poller as the reaper's sibling** and the unattended wake actor: same lease and fencing rules, idempotent through first-write-wins resolve, and the pod that claims and reifies since no inbound request will ever arrive. This is the one place decision 5 is extended.
- **Signing is mandatory** for parked webhook approvals. New contexts bind request and response to a per-poll id. `poll_url` without a webhook secret is a config error.
- **No fetch failure denies.** Transport, status, signature, and missing-id cases leave the run parked. Only reaper expiry denies.

Rejected: ingress callback, longer synchronous hold, per-decision polling, outbound persistent stream (deferred; same body shapes), receiver-supplied poll URL (SSRF surface).

## Things to check

- The 2026-07-21 park/reify ADR is not on this branch (it lives on the closed #611 stack), so the relative links to it resolve only once it lands. Happy to rebase onto wherever that ADR ends up.
- The instance id on the webhook payload (#495) is cited as a prerequisite, not assumed present.
- Unattended outcome delivery (where a reified unattended run's result goes with no SSE subscriber) is explicitly deferred to a separate decision. Flagging in case that conflicts with the A2A plans.

Refs: #271

### DIFF
--- a/docs/adr/2026-09-03-hitl-webhook-decision-polling.md
+++ b/docs/adr/2026-09-03-hitl-webhook-decision-polling.md
@@ -1,0 +1,372 @@
+<!-- markdownlint-disable MD033 -->
+# Poll the approval webhook for parked decisions instead of receiving a callback
+
+- Status: **proposed**
+- Deciders: Tony Rogers
+- Date: 2026-09-03
+- Amends: [2026-07-21-hitl-park-reify.md](2026-07-21-hitl-park-reify.md)
+  (decisions 5, 8, 10, 12, 13) and the Route A definition in
+  [2026-06-16-hitl-approval-architecture.md](2026-06-16-hitl-approval-architecture.md)
+
+Technical Story: [#271](https://github.com/mezmo/aura/issues/271) (HITL
+Park/Reify); polling tracked by: unfiled
+
+## Context and Problem Statement
+
+The HITL architecture ADR defines two decision routes. Route A (webhook) is one
+synchronous HTTP round trip: aura POSTs the approval request to a receiver and
+blocks up to `timeout_secs` for the decision in the response body. Route B
+(conversational) parks the call and receives the decision through the
+**decision ingress**, `POST /v1/approvals/{decision_id}` on the aura web
+server.
+
+The park/reify ADR makes parking durable. It does so only for Route B: the
+webhook route "resolves in one round trip and keeps no parked records", and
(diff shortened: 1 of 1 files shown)


docs/adr/2026-09-03-hitl-webhook-decision-polling.md
```diff
@@ -0,0 +1,372 @@
...
+  holds; the receiver holds the request until one resolves or a wait budget
+  expires.
+- **E. Outbound persistent stream**: aura opens an SSE or WebSocket connection
+  to the receiver and decisions are pushed down it.
+- **F. Receiver-supplied poll URL**: the pending response carries the URL aura
+  should poll next.
+
+## Decision Outcome
+
+Chosen option: **D, batched long-poll**, with the pending response as the
```
esatterwhite, line 87: Is the idea here that it always polls in the case it gets a pending? or its just one of the options?

If its an option - Something about the webhook request should indicate the preferred decisions response method.
If you have the callback url configured we need to know that aura is waiting to be told a answer
if you don't have the callback, we have to hold a request open an try to capture a decision so we can respond to aura.

if you have polling setup, we don't need to do either of those things. 
But mezmo has no understanding of agent configuration to make this decision

It _could_ be configured on the platform side and that could inform the decisions. but it means you can't just change aura configuration and have things work. 

it creates _more_ configuration management

Shearerbeard:
@teriyakichild I'll probably commit here when I'm done - a few decisions have changed since this proposal so lets leave it open for now

PR #662, opened 2026-09-04
Derive approval cancellation events from the store's cleared set

Shearerbeard:
## What

`ApprovalStore::cancel_request` now returns the approvals it actually cleared, and the park sweep publishes one `approval_completed(cancelled)` event per returned record - replacing the old check-then-act sweep whose pre-check windows could publish `cancelled` for a ticket a human had already decided (204 stands, stream disagreed). The sweep runs as its own spawned task, so an abandoned caller await cannot stop publication mid-flight.

Backend notes:

- File: two-phase classify-then-remove under the lock; a decision file present marks resolve residue (removed, never returned); per-file remove failures stay retryable.
- Redis: the sweep take is a type-guarded script inside one MULTI pipe (string -> GETDEL + in-script SREM; wrong-typed key -> warn, left in place with its index entry kept for retry; stale id -> SREM). Resolve's atomic Lua claim is untouched.

## Conceded boundary (ruled on the board)

A lost EXEC reply, a process crash, or runtime shutdown can still lose those cancellation events (registry warn-and-empty; the task dies). Recorded as the V1 residual after five review rounds converged; the registration-data fallback was evaluated and rejected as unsafe.

## Review history

Five rounds on the board ledger: two in-harness Gate A passes, two external adversarial rounds (findings fixed in-range), and a final full-range Gate A pass over this exact head. Tests: the four sweep interleaving pins, wrong-type/invalid-UTF-8 redis legs against a live server, and the file-store residue legs.

Card: P48 (park/reify V1 line)

### DIFF
--- a/crates/aura-web-server/tests/redis_session_store_test.rs
+++ b/crates/aura-web-server/tests/redis_session_store_test.rs
@@ -505,8 +505,9 @@ async fn approval_cancel_request_returns_cleared_set() {
 
 /// One corrupt record must not fail `cancel_request` for its whole request:
 /// with a list planted at one approval key the sweep still returns the valid
-/// sibling and sweeps both ids from the index, and the same holds when the
-/// planted record is non-UTF-8 bytes.
+/// sibling and sweeps its id from the index, while the wrong-typed key and
+/// its index entry stay in place for a later sweep to retry; the same holds
+/// when the planted record is non-UTF-8 bytes, which the sweep consumes.
 #[tokio::test]
 async fn cancel_request_skips_a_wrong_type_value_and_returns_valid_records() {
     let config = test_config(60);
@@ -545,13 +546,60 @@ async fn cancel_request_skips_a_wrong_type_value_and_returns_valid_records() {
     );
     assert!(approvals.get(&valid_id).await.unwrap().is_none());
     assert!(
-        redis::cmd("SMEMBERS")
+        !redis::cmd("SISMEMBER")
             .arg(&req_index_key)
+            .arg(valid_id.to_string())
+            .query_async::<bool>(&mut raw)
+            .await
+            .unwrap(),
+        "the swept record's index entry went with it"
+    );
+    assert!(
+        redis::cmd("SISMEMBER")
+            .arg(&req_index_key)
+            .arg(corrupt_id.to_string())
+            .query_async::<bool>(&mut raw)
+            .await
+            .unwrap(),
+        "the wrong-typed id keeps its index entry"
+    );
+    assert_eq!(
+        redis::cmd("TYPE")
+            .arg(&corrupt_key)
+            .query_async::<String>(&mut raw)
+            .await
+            .unwrap(),
+        "list",
+        "the wrong-typed key is left in place"
+    );
+    assert_eq!(
+        redis::cmd("LRANGE")
+            .arg(&corrupt_key)
+            .arg(0)
+            .arg(-1)
             .query_async::<Vec<String>>(&mut raw)
             .await
-            .unwrap()
-            .is_empty(),
-        "SREM swept both ids from the request index"
+            .unwrap(),
+        ["planted list, not a record"],
+        "the planted list survives the sweep"
+    );
+
+    let cleared_again = approvals
+        .cancel_request("req-wrong-type")
+        .await
+        .expect("a retry over a wrong-typed key must not fail the sweep");
+    assert!(
+        cleared_again.is_empty(),
+        "a wrong-typed key alone clears nothing"
+    );
+    assert!(
+        redis::cmd("SISMEMBER")
+            .arg(&req_index_key)
+            .arg(corrupt_id.to_string())
+            .query_async::<bool>(&mut raw)
+            .await
+            .unwrap(),
+        "the retry keeps the wrong-typed id indexed"
     );
 
     let second_valid = make_parked("req-wrong-type", Duration::from_secs(60));
@@ -579,13 +627,17 @@ async fn cancel_request_skips_a_wrong_type_value_and_returns_valid_records() {
         "the valid record is returned unchanged"
     );
     assert!(
+        approvals.get(&non_utf8_id).await.unwrap().is_none(),
+        "the non-UTF-8 record was consumed"
+    );
+    assert_eq!(
         redis::cmd("SMEMBERS")
             .arg(&req_index_key)
             .query_async::<Vec<String>>(&mut raw)
             .await
-            .unwrap()
-            .is_empty(),
-        "SREM swept both ids from the request index"
+            .unwrap(),
+        vec![corrupt_id.to_string()],
+        "the non-UTF-8 record's index entry was swept; the wrong-typed one remains"
     );
 }
 
--- a/crates/aura-web-server/tests/redis_session_store_test.rs
+++ b/crates/aura-web-server/tests/redis_session_store_test.rs
@@ -457,6 +457,34 @@ async fn approval_cancel_request_returns_cleared_set() {
         Err(ResolveError::NotFound),
         "a cleared ticket resolves NotFound"
     );
+
+    // A registration racing the sweep — same request id, added after the
+    // cancel — keeps its index entry: a second cancel still discovers and
+    // takes it.
+    let late = make_parked("req-cancel-return", Duration::from_secs(60));
+    let late_id = late.request.decision_id;
+    let late_record = ParkedApprovalRecord::from(&late);
+    approvals.register(late).await.unwrap();
+
+    let cleared_late = approvals.cancel_request("req-cancel-return").await.unwrap();
+
+    assert_eq!(
+        cleared_late.len(),
+        1,
+        "the post-cancel registration is discoverable"
+    );
+    assert_eq!(
+        ParkedApprovalRecord::from(&cleared_late[0]),
+        late_record,
+        "the late record is returned unchanged"
+    );
+    assert_eq!(
+        approvals
+            .resolve(&late_id, ApprovalDecision::Approved)
+            .await,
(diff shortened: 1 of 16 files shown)


jakedipity (commented):
Review of the five items I'd act on. Verified green locally at `f91dd617` — `cargo clippy --workspace --all-targets` (0 warnings), `cargo +nightly fmt --check`, `cargo test --workspace`, and the 36 Redis integration tests against a live valkey 8 including both new sweep tests.

The design change is right. Deriving the cancelled events from what the store actually cleared removes a real check-then-act window — the old path did `recorded_decision()`, then `try_parked()`, then published, so a decision landing mid-sequence produced a `cancelled` event contradicting a recorded approval. The new shape can't disagree.

Two inline comments below. Three more findings sit on lines this diff doesn't touch, so they're here instead.

---

### The owner-id convention is spelled out twice

https://github.com/mezmo/aura/blob/f91dd6171e1b195b8bd0a5f9bb65102ed3b86221/crates/aura/src/hitl/gate.rs#L166

`park/commit.rs:52` already defines `run_owner_id` for exactly this string, and the whole sweep depends on the two spellings staying byte-identical. The two sides also reach the run id differently — the gate formats a `RunId` through its `Display`, while `ParkGuard` and the orchestrator pass `persistence.run_id()` as a raw `&str`. Both are canonical UUIDs today (`Uuid::now_v7().to_string()`), and `guard_learns_the_decision_at_registration` does cover the pairing end to end, so nothing is broken. One import and one call site makes it structural instead of conventional.

(I'd skip retyping `run_owner_id` to take `&RunId` — that means threading the type through `ParkGuard` and the orchestrator, which costs more than it buys.)

---

### `unrecorded_guard_drop_is_inert` passes whether or not the guard arms

https://github.com/mezmo/aura/blob/f91dd6171e1b195b8bd0a5f9bb65102ed3b86221/crates/aura/src/orchestration/park/guard.rs#L216-L238

The fixture registers under owner `run:someone-else` while the guard carries run id `0191e8c0-4444-…-042`. Delete the `if *self.parked_calls.lock()... == 0 { return; }` early return and `Drop` sweeps owner `run:0191e8c0-4444-…-042`, which matches nothing — `store.get(&other).is_some()` still holds and the test stays green.

This is the only test of the arming condition the diff rewired. Now that the sweep is owner-scoped rather than id-list-scoped, the fixture has to register under `run:{run_id}` to discriminate at all.

---

### The design doc still declares the old contract

https://github.com/mezmo/aura/blob/f91dd6171e1b195b8bd0a5f9bb65102ed3b86221/docs/design/session-storage.md#L226

Line 226 still reads `-> Result<(), SessionStoreError>`, line 332 still says `cancel_request()` "deletes matching store entries", and the key-schema row at line 408 still describes the index-key delete. Both integration test files cite this doc as the contract source — `redis_session_store_test.rs` names §12 in its header.

The return value is now load-bearing, since it's the sole source of `approval_completed(cancelled)` events. Someone writing a new backend from this doc ships one that clears entries correctly and emits no cancellation events at all.

---

Smaller things I didn't file separately: `ParkGuard::parked_calls` is a `Mutex<usize>` whose value is only ever compared to zero (an `AtomicBool armed` drops the lock and both poisoning `expect`s); `RedisApprovalStore::take` now decodes a record its only caller discards via `.map(|_| ())`, which adds a `Decode` error path to `remove` after the key is already GETDEL'd; the comment at `orchestrator.rs:572` says the guard "stays inert until the park arm records a decision id", naming a mechanism this diff deleted; and the wrong-typed-key test plants a key with `RPUSH` and no TTL that the code deliberately leaves in place, so each run leaks one immortal key into the shared Redis.

crates/aura/src/session_store/mod.rs
```diff
@@ -84,8 +84,14 @@ pub trait ApprovalStore: Send + Sync {
     /// Remove a parked entry.
     async fn remove(&self, id: &DecisionId) -> Result<(), SessionStoreError>;
 
-    /// Remove every approval parked under a request id.
-    async fn cancel_request(&self, request_id: &str) -> Result<(), SessionStoreError>;
+    /// Remove every approval parked under a request id, returning the
+    /// approvals actually cleared. A ticket decided before the sweep is
+    /// absent — resolve removed it — so the return is the authoritative
+    /// record of what the cancellation applies to.
```
jakedipity: The invariant holds, but "resolve removed it" is only the mechanism for the memory and Redis backends.

The file backend's `resolve` unlinks the approval best-effort (`file.rs:208-218`) and explicitly tolerates the file surviving. The property holds there because `cancel_request_sync` checks `decision_path().try_exists()` and routes the survivor into `stale_decided` — it sweeps the residue without returning it.

Worth rewording so the trait doc doesn't assert a mechanism one of its implementors doesn't use. The guarantee is the same either way; only the reason differs.

crates/aura-web-server/tests/redis_session_store_test.rs
```diff
@@ -419,6 +420,227 @@ async fn approval_cancel_request_removes_only_matching() {
...
+        "a cleared ticket resolves NotFound"
+    );
+
+    // A registration racing the sweep — same request id, added after the
+    // cancel — keeps its index entry: a second cancel still discovers and
+    // takes it.
+    let late = make_parked("req-cancel-return", Duration::from_secs(60));
+    let late_id = late.request.decision_id;
+    let late_record = ParkedApprovalRecord::from(&late);
+    approvals.register(late).await.unwrap();
```
jakedipity: This doesn't exercise the race the comment describes, and the comment half concedes it — "added after the cancel". `approvals.register(late)` runs strictly after `cancel_request` returned, so nothing is concurrent with the sweep.

Under the old `pipe.del(&req_key)` the sweep would drop the whole index key and this `register` would `sadd` it straight back. `SISMEMBER` is true either way, and `cleared_late.len() == 1` holds either way. Revert the index handling to the old `DEL` and everything from here down still passes.

Since replacing the index-key `DEL` with per-id `SREM` is the central change in this PR, this is the assertion that should be defending it. Catching the real window means landing a registration between the `SMEMBERS` and the `EXEC` — an interleave, not a sequence.

Shearerbeard:
> Review of the five items I'd act on. Verified green locally at `f91dd617` — `cargo clippy --workspace --all-targets` (0 warnings), `cargo +nightly fmt --check`, `cargo test --workspace`, and the 36 Redis integration tests against a live valkey 8 including both new sweep tests.
> 
> The design change is right. Deriving the cancelled events from what the store actually cleared removes a real check-then-act window — the old path did `recorded_decision()`, then `try_parked()`, then published, so a decision landing mid-sequence produced a `cancelled` event contradicting a recorded approval. The new shape can't disagree.
> 
> Two inline comments below. Three more findings sit on lines this diff doesn't touch, so they're here instead.
> 
> ### The owner-id convention is spelled out twice
> https://github.com/mezmo/aura/blob/f91dd6171e1b195b8bd0a5f9bb65102ed3b86221/crates/aura/src/hitl/gate.rs#L166
> 
> `park/commit.rs:52` already defines `run_owner_id` for exactly this string, and the whole sweep depends on the two spellings staying byte-identical. The two sides also reach the run id differently — the gate formats a `RunId` through its `Display`, while `ParkGuard` and the orchestrator pass `persistence.run_id()` as a raw `&str`. Both are canonical UUIDs today (`Uuid::now_v7().to_string()`), and `guard_learns_the_decision_at_registration` does cover the pairing end to end, so nothing is broken. One import and one call site makes it structural instead of conventional.
> 
> (I'd skip retyping `run_owner_id` to take `&RunId` — that means threading the type through `ParkGuard` and the orchestrator, which costs more than it buys.)
> 
> ### `unrecorded_guard_drop_is_inert` passes whether or not the guard arms
> https://github.com/mezmo/aura/blob/f91dd6171e1b195b8bd0a5f9bb65102ed3b86221/crates/aura/src/orchestration/park/guard.rs#L216-L238
> 
> The fixture registers under owner `run:someone-else` while the guard carries run id `0191e8c0-4444-…-042`. Delete the `if *self.parked_calls.lock()... == 0 { return; }` early return and `Drop` sweeps owner `run:0191e8c0-4444-…-042`, which matches nothing — `store.get(&other).is_some()` still holds and the test stays green.
> 
> This is the only test of the arming condition the diff rewired. Now that the sweep is owner-scoped rather than id-list-scoped, the fixture has to register under `run:{run_id}` to discriminate at all.
> 
> ### The design doc still declares the old contract
> https://github.com/mezmo/aura/blob/f91dd6171e1b195b8bd0a5f9bb65102ed3b86221/docs/design/session-storage.md#L226
> 
> Line 226 still reads `-> Result<(), SessionStoreError>`, line 332 still says `cancel_request()` "deletes matching store entries", and the key-schema row at line 408 still describes the index-key delete. Both integration test files cite this doc as the contract source — `redis_session_store_test.rs` names §12 in its header.
> 
> The return value is now load-bearing, since it's the sole source of `approval_completed(cancelled)` events. Someone writing a new backend from this doc ships one that clears entries correctly and emits no cancellation events at all.
> 
> Smaller things I didn't file separately: `ParkGuard::parked_calls` is a `Mutex<usize>` whose value is only ever compared to zero (an `AtomicBool armed` drops the lock and both poisoning `expect`s); `RedisApprovalStore::take` now decodes a record its only caller discards via `.map(|_| ())`, which adds a `Decode` error path to `remove` after the key is already GETDEL'd; the comment at `orchestrator.rs:572` says the guard "stays inert until the park arm records a decision id", naming a mechanism this diff deleted; and the wrong-typed-key test plants a key with `RPUSH` and no TTL that the code deliberately leaves in place, so each run leaks one immortal key into the shared Redis.

The above should be resolved

jakedipity (approved):

justintime4tea (approved):

Issue #633, opened 2026-09-01
[BUG]: aura HITL tool pass through headers not being found


jmoses:
### Summary

With a webhook hitl endoint configure, with auth headers included, aura fails hitl checks with an error that the expected headers aren't present.  This error presents for http and https webhooks.

### Reproduction

```
[mcp.servers.local]
transport = "http_streamable"
url = "http://xxx/mcp"

[hitl]
require_approval = ["restart_server"]

[hitl.route]
mode = "webhook"
url = "https://xxx/"
timeout_secs = 5

[hitl.route.tool_headers_from_response]
x-source-header = "x-dest-header"
```

```
Question: please restart the server named foobarbaz
Answer: I couldn’t restart the server `foobarbaz` because the restart operation failed with a tool approval/channel error: `tool call blocked: approval channel error: approver identity capture failed: response missing mapped headers ["x-source-header"]`.

Even after human approval, the tool call remained blocked with the same error, so the server restart did not take place.
```

```
aura-1  | 2026-09-01T21:11:21.010Z WARN  rig::agent::prompt_request::streaming: Error while calling tool: Toolset error: ToolCallError: ToolCallError: ToolCallError: tool call blocked: approval channel error: approver identity capture failed: response missing mapped headers ["x-source-header"]
```

### Relevant log output

```shell

```

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #663, opened 2026-09-06
Name the missing response header when HITL approver capture fails

Shearerbeard:
## What

`tool_headers_from_response` maps a webhook response header onto the MCP tool call. When the approved response was missing that header, the error named only the TOML key side, so with `x-source-header = "x-dest-header"` aura complained about `x-source-header` - the header the webhook had actually returned. This PR names both sides, logs the direction at boot, and makes the failure visible under the default console filter. The example config, field rustdoc, and design doc state the key/value direction; the hosted docs page follows in a PR against mezmo/documentation.

## Operator-visible changes

- Capture failures now read: `approver identity capture failed: webhook response missing header "x-dest-header" (mapped to tool header "x-source-header")`. Names only; the same text reaches the chat error and the `aura.approval_completed` event.
- Each binary logs one boot line per mapping: `HITL approver identity forwarding on http://127.0.0.1:8477: tool header "x-source-header" <- webhook response header "x-dest-header"` (origin redacted to scheme://host).
- A capture failure logs `WARN aura::hitl: HITL approver header forwarding blocked: ... decision_id=...`, visible with no RUST_LOG set.
- The cleartext-capture warning now says what an observer can read (the captured header values, not the configured names) and that the route remains enabled.
- Webhook origin redaction goes through the URL parser, so a token hidden behind a backslash in the URL cannot reach a log line.

Fixes: #633

### DIFF
--- a/crates/aura-web-server/src/server.rs
+++ b/crates/aura-web-server/src/server.rs
@@ -425,6 +425,11 @@ async fn run(args: ServerArgs) -> std::io::Result<()> {
                 },
             )?;
             aura::hitl::warn_on_cleartext_capture(hitl);
+            if let Some(summary) = aura::hitl::approver_forwarding_summary(hitl) {
+                for line in summary.lines() {
+                    info!("{line}");
+                }
+            }
         }
     }
 
--- a/crates/aura/src/hitl/route.rs
+++ b/crates/aura/src/hitl/route.rs
@@ -121,10 +121,11 @@ pub enum ApprovalError {
     /// ingress). The decision inside is untrusted and discarded.
     #[error("approval webhook response failed signature verification: {0}")]
     ResponseUnverified(String),
-    /// Approver identity capture failed closed: the approved response was
-    /// missing mapped headers. Names only, never values; the message is
-    /// the event-level audit signal. Wraps only the capture kind, so
-    /// application-time failures cannot be mislabeled as capture failures.
+    /// Approver identity capture failed closed: the approved webhook
+    /// response was missing a mapped response header. Names only, never
+    /// values; the message is the event-level audit signal. Wraps only
+    /// the capture kind, so application-time failures cannot be
+    /// mislabeled as capture failures.
     #[error("{0}")]
     CaptureFailed(#[from] crate::approver_headers::CaptureError),
 }
@@ -518,10 +519,22 @@ impl WebhookClient {
                 let overrides = if self.tool_header_mappings.is_empty() {
                     None
                 } else {
-                    Some(crate::approver_headers::ApproverHeaders::from_captured(
-                        &self.tool_header_mappings,
-                        &response_headers,
-                    )?)
+                    Some(
+                        crate::approver_headers::ApproverHeaders::from_captured(
+                            &self.tool_header_mappings,
+                            &response_headers,
+                        )
+                        .map_err(|err| {
+                            let decision_id = request.decision_id;
+                            tracing::warn!(
+                                target: "aura::hitl",
+                                %decision_id,
+                                "HITL approver header forwarding blocked: {err}; the gated \
+                                 call was not executed"
+                            );
+                            ApprovalError::CaptureFailed(err)
+                        })?,
+                    )
                 };
                 Ok(GateDecision::Approved { overrides })
             }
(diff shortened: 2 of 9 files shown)


jmoses:
OIC I had the direction wrong in the header map.  I'm not surprised at that, I struggled with that (well not struggled, but hit it) on some other header map

crates/aura/src/hitl/route.rs
```diff
@@ -738,23 +751,54 @@ pub fn cleartext_capture_warning(config: &HitlConfig) -> Option<String> {
...
-        host_port.to_string()
-    } else {
-        format!("{scheme}://{host_port}")
+    match url::Url::parse(url) {
+        Ok(parsed) => {
+            let host = parsed.host_str().unwrap_or_default();
+            match parsed.port() {
+                Some(port) => format!("{}://{host}:{port}", parsed.scheme()),
+                None => format!("{}://{host}", parsed.scheme()),
+            }
```
jakedipity, line 799: `Url::port()` returns `None` when the port matches the scheme default, so an explicitly configured `:443` or `:80` is dropped from the output. `https://host:443/hook` renders as `https://host`. Two configs that differ only in whether the default port is written out now produce identical log lines.

`port_or_known_default()` keeps the old behaviour.

Separately, `Url::parse` lowercases and punycodes the host, so `https://approvals.exämple.com` logs as `https://xn--exmple-cua.com`. Neither is covered by the new tests, which only use `:8443`.

One more thing worth a look: `orchestration/overview.rs:224` already has `sanitize_url` doing this job with `Origin::ascii_serialization()`. Two sanitizers for secret-bearing URLs means a fix to one can miss the other.

jakedipity (commented):
Six inline comments below. Three more findings are on lines this diff does not touch, so they are here.

I verified the first two by reading the code. The rest of the review came from an automated pass and I did not re-check each one.

---

### The webhook URL is logged and sent to the client in full

https://github.com/mezmo/aura/blob/04170254927faa354664268f60f2492b2feb3d3c/crates/aura/src/hitl/route.rs#L480-L494

The `Misconfigured` reason is built with `url.as_str()`, so it carries userinfo, path and query. It then goes two places:

1. `route.rs:491` logs it with `tracing::error!(reason, ...)`.
2. `route.rs:587` returns it as `ApprovalError::Misconfigured`, `route.rs:257` turns that into `events::completed_error(...)`, and `route.rs:261` publishes it as an `ApprovalCompleted` event. The CLI renders that in chat.

With `url = "http://svc:s3cr3t@approvals.internal/hook?key=abc"` and an HMAC secret set, `s3cr3t` and `key=abc` reach both the operator log and the end user. `redact_to_origin` is right there in the same file and is not used on this path.

The trigger is narrow, since it needs `http://` plus a configured secret. But that is the case this code exists to catch. Also worth checking: `docs/design/hitl.md:516` now says every log line carries scheme and host only, which `route.rs:491` contradicts.

---

### `DecisionRouteConfig` has no `deny_unknown_fields`

https://github.com/mezmo/aura/blob/04170254927faa354664268f60f2492b2feb3d3c/crates/aura-config/src/config.rs#L1463-L1470

The attribute is on `config.rs:66` and `config.rs:806` but not here. Misspell `tool_headers_from_response`, or put it under `mode = "conversational"`, and serde drops the key. Boot succeeds. The mapping is empty, so `approver_forwarding_summary` and `cleartext_capture_warning` both return `None`, which looks exactly like a correct route with no forwarding configured. Every gated call then runs under the requester's identity.

Adding the attribute turns both mistakes into boot failures.

---

### The cancel recheck can drop an error the warn already logged

https://github.com/mezmo/aura/blob/04170254927faa354664268f60f2492b2feb3d3c/crates/aura/src/hitl/route.rs#L243-L252

`Some(decision) if !cancel.is_cancelled()` is rechecked after the select. If a client disconnects between the webhook reply landing and that recheck, a completed `Err(CaptureFailed)` falls to the `_` arm and is replaced with `Ok(Cancelled(ClientDisconnected))`. The warn at `route.rs:529` has already fired.

The result is a WARN saying forwarding was blocked next to an event saying the call was cancelled, both for the same `decision_id`. Logging after the recheck instead would avoid this.

crates/aura/src/hitl/route.rs
```diff
@@ -518,10 +519,22 @@ impl WebhookClient {
...
+                            &response_headers,
+                        )
+                        .map_err(|err| {
+                            let decision_id = request.decision_id;
+                            tracing::warn!(
+                                target: "aura::hitl",
+                                %decision_id,
+                                "HITL approver header forwarding blocked: {err}; the gated \
+                                 call was not executed"
+                            );
```
jakedipity, line 534: Three things about this warn.

It fires once per gated tool call, and `ensure_operational_warnings` (`logging.rs:282`) force-adds `aura::hitl=warn`, so no `RUST_LOG` value suppresses it. A webhook that stops setting the header produces one WARN per call for as long as the misconfiguration lasts. `route.rs:732` states the rule this breaks: "Do not call from `WebhookClient` construction: that would turn one misconfiguration into a warning per chat request." Once per call is worse than once per request.

In CLI standalone mode it produces nothing at all. `aura-cli/src/logging.rs:64` installs no fmt layer without `--log-file`, so the event is dropped. `docs/design/hitl.md:550` says this warn makes the failure visible under the default console filter, which holds only for the web server.

It also only covers `CaptureFailed`. `Transport`, `BadStatus`, `Parse`, `ResponseUnverified` and `Misconfigured` stay silent. The `Err(err)` arm at `route.rs:257` already has `decision_id` and `err` in scope and covers all six.

crates/aura/src/hitl/route.rs
```diff
@@ -1864,6 +1912,148 @@ mod tests {
...
+                log.contains(
+                    "webhook response missing header \"x-approver-id\" \
+                     (mapped to tool header \"x-forwarded-user\")"
+                ),
+                "the warn must quote the both-sides error text, got: {log}"
+            );
+            assert!(
+                !log.contains("alice"),
+                "the warn must never carry a header value, got: {log}"
+            );
```
jakedipity, line 2054: This assertion cannot fail. The response is seeded with `x-forwarded-user: alice`, but `user_mapping()` maps `x-forwarded-user <- x-approver-id`, so `from_captured` looks up `x-approver-id`, misses, and never reaches the `Some(value)` arm. `alice` is never captured, so nothing could have leaked it.

Add a `HeaderValue` to the `warn!` or to `MissingMapping`'s `Display` and this test stays green.

`missing_pair_error_names_both_sides_and_never_the_value` at `approver_headers.rs:437` has the same shape. `missing_names_are_all_reported_and_sorted` gets it right by seeding the mapped name, so it is worth copying that setup here.

crates/aura/src/hitl/route.rs
```diff
@@ -1779,7 +1823,7 @@ mod tests {
         fn warn_on_cleartext_capture_warns_once_and_redacts_the_url() {
             let log = captured_warn_log(|| {
                 super::super::warn_on_cleartext_capture(&webhook_config(
-                    "http://token:secret@approvals.example.com:8443/aura/hook?key=shh",
+                    "http://token:secret@approvals.example.com:8443\\private-token?key=shh",
```
jakedipity, line 1826: This fixture was swapped to the backslash form rather than added alongside the old one. Same in `cleartext_capture_warning_redacts_the_url` (1862) and `approver_forwarding_summary_redacts_the_url` (1985). The remaining summary test uses a bare host with no path.

So nothing now asserts that an ordinary `scheme://host/path?query` redacts to origin, which is what every real webhook URL looks like. `examples/complete/hitl-webhook.toml` ships `https://approvals.example.com/aura`.

Keeping one plain-path fixture alongside the backslash one would cover both.

crates/aura-web-server/src/server.rs
```diff
@@ -425,6 +425,11 @@ async fn run(args: ServerArgs) -> std::io::Result<()> {
                 },
             )?;
             aura::hitl::warn_on_cleartext_capture(hitl);
+            if let Some(summary) = aura::hitl::approver_forwarding_summary(hitl) {
+                for line in summary.lines() {
+                    info!("{line}");
+                }
```
jakedipity, line 431: This is `info!` on target `aura_web_server::server`, and the default filter drops it in the deployments that need it.

`logging.rs:534` builds the filter as `{binary_name}=info`, then `ensure_operational_warnings` pins only `aura_config=warn` and `aura::hitl=warn`. Setting `RUST_LOG` replaces the default string and leaves those two pins. `compose/base.yml:72` sets `RUST_LOG: warn`, and the integration harness spawns the server the same way (`hitl_header_forwarding_tests.rs:98`).

So under `RUST_LOG=warn` the cleartext WARN on the line above survives and this summary does not. `docs/design/hitl.md:510` presents the two as equally guaranteed. Switching to `target: "aura::hitl"` does not help, since that pin is at warn as well.

examples/complete/hitl-webhook.toml
```diff
@@ -47,6 +47,11 @@ headers = { "x-aura-tenant" = "sre-prod" }
 # service can see who asked. Keys are webhook header names, values are inbound
 # request header names (matched case-insensitively).
 headers_from_request = { "authorization" = "authorization" }
+# Forward the approver's identity onto the gated tool call. Keys are outbound
+# MCP tool header names, values are webhook response header names (matched
+# case-insensitively). Every value-side header must be present on an approved
+# response or the gated call fails closed.
+tool_headers_from_response = { "authorization" = "x-approver-token" }
```
jakedipity, line 54: This makes the shipped example fail closed unless the webhook returns `x-approver-token`, and Prerequisites (lines 10-19) still asks only for an endpoint that replies allow/deny.

Someone follows the example, approves a `helm_uninstall`, and gets `approval channel error: webhook response missing header "x-approver-token"`. The action they just approved does not run. Nothing catches it, because `test_all_shipped_configs_parse` only checks parsing.

Also `[mcp.servers.kubernetes]` declares no `headers`, so the mapping injects `authorization` into a local `npx kubernetes-mcp-server` that ignores it. The example does not demonstrate anything observable.

Either ship this commented out, or add the required response header to Prerequisites.

PR #665, opened 2026-09-08
fix(init): exit prompts on terminal interrupts

### LINKED ISSUE #629: [BUG]: aura init prompts swallow ctrl-c/z
### Summary

During the `aura init` process, if you want to exit early, and press CTRL-C, the UI prints the characters, and does not exit.

### Reproduction

1. `aura init`
2. `ctrl-c` the first prompt
3. Expected: process exits
  * actual: does not exit

### Relevant log output

```shell

```

### Additional Context

_No response_

### Upload screenshots

_No response_

### Searched Issues

- [x] No similar issues found

### Code of Conduct

jronaldi0509:
## What

Make every interactive `aura init` prompt exit immediately on Ctrl-C or
Ctrl-Z, including the masked API-key prompt. The shared terminal reader keeps
line editing controls and restores terminal settings before handling SIGINT or
SIGTERM.

## Why

`aura init` currently reads prompts through cooked stdin and can print or
buffer interrupt keys instead of exiting. This fixes #629 without changing the
scripted and non-interactive input path.

## Testing

- `cargo test --workspace` — 2,295 passed, 15 ignored, 0 failed
- `make fmt-check`
- `make lint`
- `make lint-commits`
- `cargo test -p aura-cli --test init_interrupts -- --nocapture` — 4 passed

The pseudo-terminal tests cover Ctrl-Z at the masked API-key prompt, Ctrl-U
editing, terminal restoration on SIGTERM during a prompt, and default SIGTERM
termination after a prompt returns.

Fixes: #629

### DIFF
--- a/crates/aura-cli/src/init.rs
+++ b/crates/aura-cli/src/init.rs
@@ -112,10 +112,17 @@ pub struct InitArgs {
 pub fn run_init(args: &InitArgs) -> Result<()> {
     let is_tty = std::io::stdin().is_terminal();
     let interactive = !args.non_interactive && is_tty;
+    let stdin: Box<dyn std::io::BufRead> = if is_tty {
+        // The terminal editor opens stdin directly. Keeping another Stdin
+        // handle here can hold its global input lock while readline starts.
+        Box::new(std::io::empty())
+    } else {
+        Box::new(std::io::BufReader::new(std::io::stdin()))
+    };
     let mut prompter = Prompter {
         interactive,
         is_tty,
-        stdin: std::io::stdin().lock(),
+        stdin,
     };
     if prompter.interactive {
         println!(
--- a/crates/aura-cli/src/init/prompt.rs
+++ b/crates/aura-cli/src/init/prompt.rs
@@ -2,9 +2,357 @@
 //! logic stays testable without a terminal — tests construct a `Prompter` over
 //! scripted stdin.
 
-use std::io::{BufRead, Write as _};
+use std::io::{BufRead, Write};
+use std::sync::{
+    Arc, LazyLock,
+    atomic::{AtomicBool, AtomicUsize, Ordering},
+};
+use std::time::Duration;
 
 use anyhow::{Result, bail};
+use crossterm::{
+    cursor::{MoveLeft, MoveRight},
+    event::{self, Event, KeyCode, KeyEvent, KeyEventKind, KeyModifiers},
+    queue,
+    terminal::{Clear, ClearType},
+};
+use signal_hook::{
+    SigId,
+    consts::signal::{SIGINT, SIGTERM},
+};
+use unicode_width::UnicodeWidthChar;
+
+fn is_prompt_interrupt(key: &KeyEvent) -> bool {
+    key.modifiers.contains(KeyModifiers::CONTROL)
+        && matches!(key.code, KeyCode::Char('c' | 'C' | 'z' | 'Z'))
+}
+
+struct PromptSignals {
+    outside_prompt: Arc<AtomicBool>,
+    pending_signal: Arc<AtomicUsize>,
+    _signal_ids: Vec<SigId>,
+}
+
+impl PromptSignals {
+    fn install() -> std::io::Result<Self> {
+        let outside_prompt = Arc::new(AtomicBool::new(true));
+        let pending_signal = Arc::new(AtomicUsize::new(0));
+        let mut signal_ids = Vec::new();
+        for signal in [SIGINT, SIGTERM] {
+            // Keep the signal-hook actions for the process lifetime. Removing
+            // the final action would leave the signal ignored rather than
+            // restoring its default disposition.
+            signal_ids.push(signal_hook::flag::register_conditional_default(
+                signal,
+                Arc::clone(&outside_prompt),
+            )?);
+            signal_ids.push(signal_hook::flag::register_usize(
+                signal,
+                Arc::clone(&pending_signal),
+                signal as usize,
+            )?);
+        }
+        Ok(Self {
+            outside_prompt,
+            pending_signal,
+            _signal_ids: signal_ids,
+        })
+    }
+}
+
+static PROMPT_SIGNALS: LazyLock<std::io::Result<PromptSignals>> =
+    LazyLock::new(PromptSignals::install);
+
+struct RawModeGuard {
+    signals: &'static PromptSignals,
+}
+
+impl RawModeGuard {
+    fn enter() -> std::io::Result<Self> {
+        let signals = PROMPT_SIGNALS
+            .as_ref()
+            .map_err(|error| std::io::Error::new(error.kind(), error.to_string()))?;
+        signals.pending_signal.store(0, Ordering::SeqCst);
+        signals.outside_prompt.store(false, Ordering::SeqCst);
+        if let Err(error) = crossterm::terminal::enable_raw_mode() {
+            signals.outside_prompt.store(true, Ordering::SeqCst);
+            return Err(error);
+        }
+        Ok(Self { signals })
+    }
+
+    fn pending_signal(&self) -> usize {
+        self.signals.pending_signal.load(Ordering::SeqCst)
+    }
+}
+
+impl Drop for RawModeGuard {
+    fn drop(&mut self) {
+        let _ = crossterm::terminal::disable_raw_mode();
+        self.signals.outside_prompt.store(true, Ordering::SeqCst);
+    }
+}
+
+#[derive(Default)]
+struct LineEditor {
+    chars: Vec<char>,
+    cursor: usize,
+}
+
+impl LineEditor {
+    fn text(&self) -> String {
+        self.chars.iter().collect()
+    }
+
+    fn cursor_width(&self) -> usize {
+        display_width(&self.chars[..self.cursor])
+    }
+
+    fn insert(&mut self, value: char) {
+        self.chars.insert(self.cursor, value);
+        self.cursor += 1;
+    }
+
+    fn backspace(&mut self) -> bool {
+        if self.cursor == 0 {
+            return false;
+        }
+        self.cursor -= 1;
+        self.chars.remove(self.cursor);
+        true
+    }
+
+    fn delete(&mut self) -> bool {
+        if self.cursor == self.chars.len() {
+            return false;
+        }
+        self.chars.remove(self.cursor);
+        true
+    }
+
+    fn clear_to_start(&mut self) -> bool {
+        if self.cursor == 0 {
+            return false;
+        }
+        self.chars.drain(..self.cursor);
+        self.cursor = 0;
+        true
+    }
+
+    fn erase_word(&mut self) -> bool {
+        let initial = self.cursor;
+        while self.cursor > 0 && self.chars[self.cursor - 1].is_whitespace() {
+            self.cursor -= 1;
+        }
+        while self.cursor > 0 && !self.chars[self.cursor - 1].is_whitespace() {
+            self.cursor -= 1;
+        }
+        if self.cursor == initial {
+            return false;
+        }
+        self.chars.drain(self.cursor..initial);
+        true
+    }
+
+    fn redraw(&self, old_cursor_width: usize, echo: bool) -> std::io::Result<()> {
+        if !echo {
+            return Ok(());
+        }
+        let mut stdout = std::io::stdout();
+        queue_move_left(&mut stdout, old_cursor_width)?;
+        queue!(stdout, Clear(ClearType::UntilNewLine))?;
+        let text = self.text();
+        write!(stdout, "{text}")?;
+        queue_move_left(
+            &mut stdout,
+            display_width(&self.chars).saturating_sub(self.cursor_width()),
+        )?;
+        stdout.flush()
+    }
+}
+
+fn display_width(chars: &[char]) -> usize {
+    chars
+        .iter()
+        .map(|character| character.width().unwrap_or(0))
+        .sum()
+}
+
+fn queue_move_left(output: &mut impl Write, mut columns: usize) -> std::io::Result<()> {
+    while columns > 0 {
+        let step = columns.min(u16::MAX as usize) as u16;
+        queue!(output, MoveLeft(step))?;
+        columns -= usize::from(step);
+    }
+    Ok(())
+}
+
+fn queue_move_right(output: &mut impl Write, mut columns: usize) -> std::io::Result<()> {
+    while columns > 0 {
+        let step = columns.min(u16::MAX as usize) as u16;
+        queue!(output, MoveRight(step))?;
+        columns -= usize::from(step);
+    }
+    Ok(())
+}
+
+fn read_tty_answer(echo: bool) -> Result<Option<String>> {
+    let raw_mode = RawModeGuard::enter()?;
+    let mut editor = LineEditor::default();
+
+    loop {
+        let signal = raw_mode.pending_signal();
+        if signal != 0 {
+            bail!("init interrupted by signal {signal}");
+        }
+        if !event::poll(Duration::from_millis(50))? {
+            continue;
+        }
+        let Event::Key(key) = event::read()? else {
+            continue;
+        };
+        if !matches!(key.kind, KeyEventKind::Press | KeyEventKind::Repeat) {
+            continue;
+        }
+        if is_prompt_interrupt(&key) {
+            print!(
+                "^{}\r\n",
+                match key.code {
+                    KeyCode::Char(c) => c.to_ascii_uppercase(),
+                    _ => unreachable!(),
+                }
+            );
+            std::io::stdout().flush()?;
+            bail!("init interrupted");
+        }
+
+        match key.code {
+            KeyCode::Enter => {
+                print!("\r\n");
+                std::io::stdout().flush()?;
+                return Ok(Some(editor.text()));
+            }
+            KeyCode::Char('d')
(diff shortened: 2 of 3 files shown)


crates/aura-cli/tests/init_interrupts.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        set_nonblocking(master.as_raw_fd());
+        let child = Command::new(env!("CARGO_BIN_EXE_aura"))
+            .args(args)
+            .arg("--output")
+            .arg(output_path)
+            .env("HOME", home.path())
+            .env("TERM", "xterm-256color")
+            .env_remove("OPENAI_API_KEY")
+            .env_remove("ANTHROPIC_API_KEY")
```
justintime4tea, line 40: Two of the four tests in this file fail on any machine that exports `GEMINI_API_KEY` or `OPENROUTER_API_KEY`.

Init senses a provider from all four keys `default_key_env` returns, not just these two. A sensed provider gives the choice prompt a default, so it renders as `Provider [1]: ` rather than `Provider: `, and `wait_for("Provider: ")` never matches. `ctrl_u_clears_the_current_answer_before_provider_selection` and `sigterm_exits_an_active_prompt_and_restores_terminal` then spin for 10s and panic.

Reproduced on this branch:

```
$ cargo test -p aura-cli --test init_interrupts
test result: FAILED. 2 passed; 2 failed

$ env -u GEMINI_API_KEY cargo test -p aura-cli --test init_interrupts
test result: ok. 4 passed; 0 failed
```

CI almost certainly has no provider keys set, so this passes there and only breaks contributors, who CLAUDE.md asks to run the workspace tests before committing.

Scrubbing all four would fix it:

```rust
.env_remove("OPENAI_API_KEY")
.env_remove("ANTHROPIC_API_KEY")
.env_remove("GEMINI_API_KEY")
.env_remove("OPENROUTER_API_KEY")
```

There is a second symptom of the same root cause worth covering: with `OPENROUTER_API_KEY` set, menu entry `2` becomes OpenRouter, whose key is present, so `spec.rs` asks "Use this environment variable?" instead of "Enter your API key: " and the Ctrl-U test fails a different way.

crates/aura-cli/src/init/prompt.rs
```diff
@@ -2,9 +2,404 @@
...
+        if !self.active {
+            return None;
+        }
+        let _ = crossterm::terminal::disable_raw_mode();
+        self.active = false;
+        self.signals.finish_prompt()
+    }
+
+    fn finish<T>(mut self, value: T) -> Result<T> {
+        if let Some(signal) = self.restore() {
```
justintime4tea, line 131: SIGTERM during a prompt exits with status 1 instead of dying by the signal. That is inconsistent with `sigterm_keeps_its_default_action_after_a_prompt_completes` in this same PR, which asserts `status.signal() == Some(SIGTERM)` for the post-prompt case.

Verified against the built binary, signalling while the provider prompt was up:

```
EXIT: status 1
OUTPUT AFTER PROMPT: 'Error: init interrupted by signal 15\r\n'
```

So one signal produces two different observable outcomes depending on timing, and a supervisor or shell cannot distinguish "an operator terminated init" from "init failed".

The guard has already restored the terminal by this point, so it can re-raise rather than bail:

```rust
fn finish<T>(mut self, value: T) -> Result<T> {
    if let Some(signal) = self.restore() {
        let _ = low_level::emulate_default_handler(signal as i32);
        bail!("init interrupted by signal {signal}"); // only if emulation fails
    }
    Ok(value)
}
```

The same applies to the Ctrl-C path at line 273. It previously terminated via SIGINT's default action for exit code 130, and now returns an `anyhow` error that `Termination` maps to 1, so scripts branching on 130 stop seeing it.

crates/aura-cli/src/init.rs
```diff
@@ -112,10 +112,17 @@ pub struct InitArgs {
 pub fn run_init(args: &InitArgs) -> Result<()> {
     let is_tty = std::io::stdin().is_terminal();
     let interactive = !args.non_interactive && is_tty;
+    let stdin: Box<dyn std::io::BufRead> = if is_tty {
+        // The terminal editor opens stdin directly. Keeping another Stdin
+        // handle here can hold its global input lock while readline starts.
```
justintime4tea, line 117: Two problems here.

**The rationale is not accurate.** crossterm's `tty_fd()` borrows the raw stdin file descriptor when stdin is a terminal and opens `/dev/tty` otherwise. It never constructs a `std::io::Stdin`, so Rust's global input lock is never contended by the terminal editor, and there is no readline anywhere in this path.

**More importantly, this removes the cooked-mode fallback.** Before this PR every non-secret prompt went through `self.stdin.read_line`. Now, whenever stdin is a terminal, `read_tty_answer` is the only input path. `PROMPT_SIGNALS` is a `LazyLock<io::Result<..>>`, so a single failure of `PromptSignals::install()` (restricted `sigaction` in a hardened container, say) is cached for the process lifetime, and every prompt afterwards aborts init with an opaque io error. Same story if `enable_raw_mode` fails. Those runs worked before this change.

Restoring `std::io::stdin().lock()` costs nothing and brings the fallback back. To make it actually engage, `read_answer` could drop to the cooked path when `RawModeGuard::enter()` errors rather than propagating.

crates/aura-cli/src/init/prompt.rs
```diff
@@ -2,9 +2,404 @@
...
+            self.cursor -= 1;
+        }
+        if self.cursor == initial {
+            return false;
+        }
+        self.chars.drain(self.cursor..initial);
+        true
+    }
+
+    fn redraw(&self, old_cursor_width: usize, echo: bool) -> std::io::Result<()> {
```
justintime4tea, line 205: This tracks the typed text's width but never the prompt's width or the terminal's column count, so any answer that wraps corrupts the display on every subsequent keystroke.

On an 80-column terminal `ask_model` prints roughly 40 columns of prompt. Type a 45-character model id and the line wraps. The next keystroke calls `queue_move_left(stdout, 40)`, which emits CUB(40); that clamps at column 0 of the current row and cannot move up a row. `Clear(ClearType::UntilNewLine)` then erases only that row, and the full text is rewritten from there. The prompt row keeps stale text and the answer is drawn twice. Backspace, Ctrl-U and Ctrl-W behave the same way, and text that shrinks never clears the trailing wrapped rows.

The captured value stays correct, so this is display corruption rather than data loss. Masked prompts are immune because `echo` is false and this returns early.

Fixing it properly means tracking prompt width and terminal size and moving by rows, which is most of what a line editor does. See the note on line 247.

crates/aura-cli/src/init/prompt.rs
```diff
@@ -2,9 +2,404 @@
...
+        if let Some(signal) = self.restore() {
+            bail!("init interrupted by signal {signal}");
+        }
+        Ok(value)
+    }
+}
+
+impl Drop for RawModeGuard {
+    fn drop(&mut self) {
+        self.restore();
```
justintime4tea, line 140: `restore()` returns the pending signal and this discards it, so a signal recorded moments earlier is lost whenever the guard drops on an error path rather than through `finish`.

Concretely: SIGTERM lands mid-prompt and is recorded, then `editor.redraw(..)?` or `event::poll(..)?` returns an io error before the loop reaches its `pending_signal()` check. The `?` unwinds, `Drop` swallows `Some(15)`, and the process exits 1 with an unrelated io message instead of reflecting the termination it was sent.

The comment on `finish_prompt` says a signal "cannot be swallowed", which is more than the protocol actually delivers. See also line 47.

crates/aura-cli/src/init/prompt.rs
```diff
@@ -2,9 +2,404 @@
...
+/// teardown already won the atomic transition and the default action must run.
+fn record_prompt_signal(state: &AtomicUsize, signal: i32) -> bool {
+    state
+        .compare_exchange(
+            ACTIVE_PROMPT,
+            signal_state(signal),
+            Ordering::SeqCst,
+            Ordering::SeqCst,
+        )
+        .is_err_and(|current| current == OUTSIDE_PROMPT)
```
justintime4tea, line 47: This returns `true` only when the swap found `OUTSIDE_PROMPT`. Once a signal has already been recorded, `current` holds that signal's state value, which is neither sentinel, so `is_err_and` yields `false` and the new signal is neither recorded nor given its default action.

Send SIGINT and then SIGTERM during a single prompt and the SIGTERM disappears entirely: no record, no default action, no exit. That is exactly the sequence a user produces when the first interrupt looks like it did nothing and they escalate.

Either handle `current > ACTIVE_PROMPT` by running the default action, or soften the "cannot be swallowed" claim on `finish_prompt` so it matches what this guarantees.

crates/aura-cli/src/init/prompt.rs
```diff
@@ -2,9 +2,404 @@
...
+}
+
+fn is_prompt_interrupt(key: &KeyEvent) -> bool {
```
justintime4tea, line 50: Mapping Ctrl-Z to the same action as Ctrl-C breaks job control. The conventional meaning is suspend, and users press it expecting to `fg` back afterwards. Raw mode has cleared ISIG so no SIGTSTP is generated, and this exits instead, discarding every answer given so far in the run.

`low_level::emulate_default_handler` is already linked here and its `DefaultKind::Stop` arm raises SIGSTOP, so suspending correctly is within reach: restore the terminal, raise SIGTSTP, re-enter raw mode on resume.

If exiting is the deliberate choice, it is worth stating in the PR description, since #629 only asked for Ctrl-C not to hang.

crates/aura-cli/src/init/prompt.rs
```diff
@@ -2,9 +2,404 @@
...
+impl PromptSignals {
+    fn install() -> std::io::Result<Self> {
+        let state = Arc::new(AtomicUsize::new(OUTSIDE_PROMPT));
+        let mut signal_ids = Vec::new();
+        for signal in [SIGINT, SIGTERM] {
+            // Keep the signal-hook actions for the process lifetime. Removing
+            // the final action would leave the signal ignored rather than
+            // restoring its default disposition.
+            let state = Arc::clone(&state);
+            signal_ids.push(unsafe {
```
justintime4tea, line 69: This can be written without `unsafe`, using a pattern this crate already uses.

`ResizeWatcher` in `repl/loop.rs` builds a `signal_hook::iterator::Signals`, keeps its handle, and runs `signals.forever()` on a thread. Doing the same here moves the closure out of signal-handler context, which is the only reason `low_level::register` is unsafe. `emulate_default_handler` is already a safe function, so the body is unchanged:

```rust
fn install() -> std::io::Result<Self> {
    let state = Arc::new(AtomicUsize::new(OUTSIDE_PROMPT));
    let mut signals = Signals::new([SIGINT, SIGTERM])?;
    let delivered = Arc::clone(&state);
    std::thread::Builder::new()
        .name("aura-init-signals".into())
        .spawn(move || {
            for signal in signals.forever() {
                if record_prompt_signal(&delivered, signal) {
                    let _ = low_level::emulate_default_handler(signal);
                }
            }
        })?;
    Ok(Self { state })
}
```

I checked this in a standalone build: a signal during an active prompt is recorded and the process stays alive, and a signal outside a prompt still takes its default action and terminates. The atomic protocol and its regression test are untouched. `_signal_ids` goes away because the `Signals` value moved into the thread holds the registration open for the process lifetime, which is the property the comment above is describing.

Costs one lazily spawned thread and a pipe round-trip of latency. The safe `signal_hook::flag` combinators are not an option: a separate load and store cannot reproduce the compare-and-exchange and would reinstate exactly the race 848a7ed closed.

Related, and optional: `ui::state::install_sigint_handler` is a second hand-written unsafe handler in this crate. Converging both on the threaded pattern would leave one mechanism instead of three.

crates/aura-cli/src/init/prompt.rs
```diff
@@ -2,9 +2,404 @@
...
+        };
+        if !matches!(key.kind, KeyEventKind::Press | KeyEventKind::Repeat) {
+            continue;
+        }
+        if is_prompt_interrupt(&key) {
+            print!(
+                "^{}\r\n",
+                match key.code {
+                    KeyCode::Char(c) => c.to_ascii_uppercase(),
+                    _ => unreachable!(),
```
justintime4tea, line 269: This holds only because `is_prompt_interrupt` currently tests `KeyCode::Char`. Adding any non-character key to that predicate turns this into a production panic the moment a user presses it. Binding the character removes the coupling:

```rust
if let KeyCode::Char(c) = key.code
    && is_prompt_interrupt(&key)
{
    print!("^{}\r\n", c.to_ascii_uppercase());
    ...
}
```

crates/aura-cli/src/init/prompt.rs
```diff
@@ -2,9 +2,404 @@
...
+                return raw_mode.finish(Some(editor.text()));
+            }
+            KeyCode::Char('d')
+                if key.modifiers.contains(KeyModifiers::CONTROL) && editor.chars.is_empty() =>
+            {
+                print!("\r\n");
+                std::io::stdout().flush()?;
+                return raw_mode.finish(None);
+            }
+            KeyCode::Char('a') if key.modifiers.contains(KeyModifiers::CONTROL) => {
```
justintime4tea, line 289: Four pairs of arms in this match are byte-for-byte identical, and the `if echo { let mut stdout = ...; queue_move_*(..)?; stdout.flush()?; }` block is copy-pasted eight times.

The pairs are `Char('a')` with `Home`, `Char('e')` with `End`, `Char('b')` with `Left`, and `Char('f')` with `Right`.

Merging each pair with an or-pattern and hoisting the echo block into a single helper would cut this down a lot. It matters beyond tidiness: the wrapping fix on line 205 otherwise has to be applied in eight places and will drift.

crates/aura-cli/src/init/prompt.rs
```diff
@@ -2,9 +2,404 @@
...
+fn queue_move_right(output: &mut impl Write, mut columns: usize) -> std::io::Result<()> {
+    while columns > 0 {
+        let step = columns.min(u16::MAX as usize) as u16;
+        queue!(output, MoveRight(step))?;
+        columns -= usize::from(step);
+    }
+    Ok(())
+}
+
+fn read_tty_answer(echo: bool) -> Result<Option<String>> {
```
justintime4tea, line 247: Worth asking whether this needs to be hand-rolled at all. `rustyline` 17 is already a direct dependency of this crate and is used across nine modules including `repl/input_reader.rs`. It ships the emacs keymap this reimplements, returns `ReadlineError::Interrupted` and `Eof` for Ctrl-C and Ctrl-D, and tracks prompt width and terminal columns so wrapped lines redraw correctly, which is the bug on line 205.

Two honest caveats, so this is a refactor rather than a deletion: it would not remove the raw-mode guard or the signal handling, which is the genuinely novel part of this PR, and masked input needs a masking highlighter rather than a built-in mode.

Separately, none of this dispatch is unit-testable. `test_support::scripted()` sets `is_tty: false`, so every existing prompt test bypasses `read_tty_answer` entirely. Ctrl-A/E/B/F, Home/End, the arrows, Delete, the Ctrl-D-on-empty path and the `KeyEventKind` filter have no unit coverage, only whatever the four pty tests happen to touch. Extracting a pure `fn apply_key(&mut LineEditor, &KeyEvent) -> Action` would make the whole table testable without a terminal, which is what this module's header promises.

crates/aura-cli/src/init/prompt.rs
```diff
@@ -105,20 +515,18 @@ impl<R: BufRead> Prompter<R> {
         }
     }
 
-    /// Prompt for an API key with masked input. On a real terminal the input
-    /// is read with echo suppressed via `rpassword`. In test contexts
-    /// (`is_tty = false`), falls back to `read_line` on the injected stdin.
-    /// Returns `None` on empty input or EOF.
+    /// Prompt for an API key with masked input. On a real terminal the shared
+    /// interrupt-aware reader suppresses echo. In test contexts (`is_tty =
+    /// false`), input comes from the injected reader. Returns `None` on empty
```
justintime4tea, line 520: This now restates what the `is_tty` field comment above it already says, and CLAUDE.md asks not to describe the same behavior in two places.

That field comment is also stale on two counts. It points at `ask_secret`, which does not exist in this module; the method here is `ask_secret_masked`, and the name `ask_secret` belongs to `repl/mcp/wizard.rs`. And it describes only secret prompts, when after this PR every prompt branches on `is_tty` through `read_answer`.

It also fails the drift test CLAUDE.md specifies for type-level comments: it had to change because a different function was added. The behavior belongs on the function, and the field should say what the value means and stop.

crates/aura-cli/tests/init_interrupts.rs
```diff
@@ -0,0 +1,242 @@
...
+            if terminal_flags(self.monitor.as_raw_fd()) & libc::ICANON == 0 {
+                return;
+            }
+            thread::sleep(Duration::from_millis(10));
+        }
+        panic!("prompt never entered raw mode");
+    }
+
+    fn send(&mut self, bytes: &[u8]) {
+        self.master.write_all(bytes).unwrap();
```
justintime4tea, line 86: `spawn` puts the pty master into `O_NONBLOCK` for the benefit of `read_available`, but that flag applies to writes too. `write_all` retries only `ErrorKind::Interrupted`, not `WouldBlock`, so if the child has not drained the tty input queue yet this panics with "Resource temporarily unavailable" after a possible partial write.

Unlikely locally, plausible on a loaded CI runner where the child is still inside `enable_raw_mode` when the test writes. Retrying on `WouldBlock` with a short sleep, or keeping a separate blocking dup of the master for writes, would remove the flake.

justintime4tea (changes requested):

PR #675, opened 2026-09-09
refactor: brand the request id

jakedipity:
RequestId becomes a newtype over String instead of a type alias, so a bare string can no longer stand in for a request id. It threads through SSE routing, the tool-event broker, MCP cancellation, HITL approval gates, the session store, and the scratchpad's per-request directory.

The brokers and the cancellation registry key their maps by RequestId rather than String. In mcp/client.rs, rmcp's own RequestId is imported as McpRequestId, so the HTTP request id and the MCP protocol request id are no longer the same type in signatures that take both.

The webhook payload and stored approval records keep their bare-string shape, which serde(transparent) guarantees and a record test now pins.

### DIFF
--- a/crates/aura-cli/src/repl/mcp/wizard.rs
+++ b/crates/aura-cli/src/repl/mcp/wizard.rs
@@ -401,7 +401,10 @@ fn verify_server(
             })
             .ok_or_else(|| format!("`{name}` missing from the connection status snapshot"));
         manager
-            .cancel_and_close_all("mcp-add-verify", "verification complete")
+            .cancel_and_close_all(
+                &aura::RequestId::new("mcp-add-verify"),
+                "verification complete",
+            )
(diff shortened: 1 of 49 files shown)


dhable (dismissed):

Shearerbeard (changes requested):
My only feedback here is that lib.rs is to general - if were starting to consolidate code around building our domain types and events lets just call it domain

Shearerbeard (dismissed):
This looks good - as we progress we might have to better module classify identifiers a cross application concerns (like RequestId) vs agent specific new types like AgentId. There are probably a few cross cutting concerns to belong in a root domain vs a specific one. Also a good place to start looking for duplicated construction and validation code if it exists (I don't see that here - just the kind of thing I like to think ahead on when designing by types)

PR #683, opened 2026-09-10
Webhook poll delivery (1/4): config surface, pending-approval scan, owner-only store files

Shearerbeard:
First of four PRs that add an asynchronous "poll" delivery mode to the HITL webhook route. This one adds the config keys and the store method the later PRs build on, and keeps the mode switched off.

What changes:

- `[hitl.route]` gains `delivery` (`sync`, the default, or `poll`), `poll_url`, `poll_interval_secs`, and `poll_request_timeout_secs`. Config validation rejects `delivery = "poll"` outright in this PR, since the code that serves it lands in the next two. It also requires `[hitl.park].enabled` for poll and refuses a plaintext `poll_url` when an HMAC secret is set.
- `ApprovalStore` gains `list_pending`, implemented for the memory, file, and Redis backends (the fault-injecting test store delegates to its inner store). Nothing calls it yet; the reconciler in the next PR does. It is a required trait method with no default, so an out-of-tree backend has to implement it. That is deliberate: a default that returns an error would show up as a failure on every reconciler tick instead of at compile time.
- Approval files and parked-run checkpoints hold credentials and tool output, so they are created 0600 in 0700 directories instead of inheriting the umask. A directory left permissive by an earlier build is tightened when the store opens or a checkpoint is written. The file store's pending scan also unlinks two kinds of leftover approval file: one that has expired, and one whose decision is already fully recorded. Redis rows already expire by TTL.
- On the existing park resume path, a checkpointed call whose store approval has vanished past the checkpoint's expiry is now an expiry error rather than a store-mismatch error. Nothing in the merged tree calls that path yet; the resume endpoint is separate work.
- The spawned-server fixture from the header-forwarding integration tests moves into the shared test module so later suites can reuse it.

Not in this PR: the reconciler and its wire format. Those come next; the park-path activation and credential persistence follow, and the last PR turns the mode on.

Testing: unit tests for the new config fields and each validation error; backend-contract tests for `list_pending` on the memory, file, and Redis stores (live rows only, empty store, expired excluded); file-store tests for unlinking an expired file and skipping an undecodable one; Redis tests for wrong-typed and undecodable keys; unit tests for the file modes. `cargo test --workspace`, clippy, and fmt are clean, and the Redis tests ran against a local Valkey.

Stack: based on `nightly`. Next: `mshearer/271-poll-protocol`.

### DIFF
--- a/crates/aura-web-server/src/session_store/redis/approval_store.rs
+++ b/crates/aura-web-server/src/session_store/redis/approval_store.rs
@@ -18,6 +18,10 @@
 //! the record TTL and pruned best-effort on resolve/remove. The cancel sweep
 //! prunes per id, never the whole index key; `SWEEP_TAKE_SCRIPT` states what
 //! each id yields.
+//! `list_pending` SCANs the parked-record keys in batches (never KEYS),
+//! skipping decision and index keys by segment and wrong-typed or
+//! undecodable records per key; the native TTL is the primary expiry, with
+//! a post-decode filter as defense in depth.
 
 use std::sync::LazyLock;
 
@@ -36,6 +40,25 @@ const MIN_TTL_SECS: u64 = 1;
 const REQ_INDEX_TTL_MARGIN_SECS: u64 = 60;
 /// Decision TTL margin over the parked record's remaining TTL.
 const DECISION_TTL_MARGIN_MS: u64 = 60_000;
+const SCAN_BATCH_SIZE: usize = 200;
+/// Type-guarded GET for the `list_pending` scan: a string key's value,
+/// integer 0 for a wrong-typed key (the caller warns and skips), nil for a
+/// key that expired or resolved between SCAN and here. A bare GET maps the
+/// server's WRONGTYPE to an extension error that would fail the whole scan.
+static TYPED_GET_SCRIPT: &str = r#"
+if redis.call('TYPE', KEYS[1]).ok == 'string' then
+    return redis.call('GET', KEYS[1])
+end
+if redis.call('EXISTS', KEYS[1]) == 1 then
+    return 0
+end
+return nil
+"#;
+/// Key prefixes under `{p}:approval:` that are not parked records: the
+/// recorded decisions and the `cancel_request` index sets. Matched against
+/// the key remainder after the `{p}:approval:` prefix is stripped.
+const DECISION_KEY_SEGMENT: &str = "decision:";
+const REQ_KEY_SEGMENT: &str = "req:";
 
 /// Sweep one approval key (KEYS[1]) out of its request index (KEYS[2]):
 /// a string key is GETDEL'd and its id SREM'd; a wrong-typed key returns 0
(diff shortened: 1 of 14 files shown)


crates/aura-web-server/src/session_store/redis/approval_store.rs
```diff
@@ -254,6 +277,92 @@ impl ApprovalStore for RedisApprovalStore {
         }
         Ok(cleared)
     }
+
+    async fn list_pending(&self) -> Result<Vec<ParkedApproval>, SessionStoreError> {
+        let mut conn = self.conn.clone();
+        let pattern = format!("{}:approval:*", self.key_prefix);
```
justintime4tea, line 283: nit: SCAN pattern doesn't escape glob chars. This may be a non-issue but thought I'd get your thoughts here.

A prefix like aura[prod] or aura?1 silently matches nothing and list_pending returns empty with no error and every parked approval hangs until expiry. Will we ever have prefix which contain these glob chars? 🤷 Maybe I'm being overly paranoid 😅

crates/aura-web-server/src/session_store/redis/approval_store.rs
```diff
@@ -254,6 +277,92 @@ impl ApprovalStore for RedisApprovalStore {
...
+            };
+            if rest.starts_with(DECISION_KEY_SEGMENT) || rest.starts_with(REQ_KEY_SEGMENT) {
+                continue;
+            }
+            let value = redis::cmd("EVAL")
+                .arg(TYPED_GET_SCRIPT)
+                .arg(1)
+                .arg(&key)
+                .query_async::<redis::Value>(&mut conn)
+                .await
```
justintime4tea, line 323: nit: we could use cached script for TYPED_GET_SCRIPT

TYPED_GET_SCRIPT (L#48) is a plain &str, so EVAL re-uploads the body on every key, and the .await is inside the loop so it's one sequential round-trip per key.

RESOLVE_SCRIPT (L#80) shows the cheaper form — wrapping this in `LazyLock<redis::Script>` and calling .`key(&key).invoke_async(&mut conn)` gets EVALSHA with automatic NOSCRIPT fallback. We'd need to drop the manual .arg(1) but otherwise it leaves every match arm below unchanged.

**Bigger win:** MGET returns nil for wrong-typed keys rather than erroring, so the filtered keys could go out in SCAN_BATCH_SIZE chunks which could take 500 round-trips (500 parked approvals) down to just 3. We'd lose the wrong typed warning, since wrong-typed and absent both result as nil, but would be much more efficient. I think 500 is a rather large number but it's illustrative, not like a real case necessarily. That said, if keys are stored for multiple AURA instances and each instance has N number of parked approvals we'd be iterating through all of them since the instance filter happens later in poller (L#126) with using MGET the roundtrips ceiling is `N / SCAN_BATCH_SIZE` rather than just N.

crates/aura/src/session_store/file.rs
```diff
@@ -325,6 +334,81 @@ impl Inner {
         }
         Ok(cleared)
     }
+
+    fn list_pending_sync(&self) -> Result<Vec<ParkedApproval>, SessionStoreError> {
+        let _guard = self.lock();
```
justintime4tea, line 339: With this lock here at the top, while a scan runs, every other store operation blocks right - register/get/resolve/remove/cancel_request? So a gated tool call parking a new approval or a human approving waits behind a background scan? Could we take a snapshot of the paths we need and just hold the lock for that? This seems like a cheap optimization that could turn  O(N)-under-lock into O(1)-under-lock. 

Something like this:

```rust
 fn list_pending_sync(&self) -> Result<Vec<ParkedApproval>, SessionStoreError> {
      // Snapshot the listing under the lock; the per-file work below already
      // tolerates a NotFound from a concurrent resolve/remove.
      let paths: Vec<PathBuf> = {
          let _guard = self.lock();
          match fs::read_dir(self.approvals_dir()) {
              Ok(entries) => entries
                  .filter_map(|e| e.ok().map(|e| e.path()))
                   // This is the filter from L#351, mid-publish temp file
                  .filter(|p| p.extension().and_then(|x| x.to_str()) == Some("json"))
                  .collect(),
              Err(err) if err.kind() == io::ErrorKind::NotFound => return Ok(Vec::new()),
              Err(err) => return Err(request_err(err)),
          }
      }; // ← lock released here

      // reads, decodes, and unlinks happen unlocked
      ...
  }
```

The part I'm less sure of is whether the read-decision-then-unlink at 379-396 needs to stay atomic against a concurrent resolve_sync? With the comment at 370-377 it says this is the retry path for resolve's own best-effort unlink, so I'm guessing not? What are your thoughts?

Issue #579, opened 2026-08-24
[EPIC]:  Virtual File System and Memory


Shearerbeard:
### Summary

Multi Pod compatible VFS for orchestration artifacts and agent memory. This allows us to claim a dir per AgentSession atomically and ensure we can always reify proper artifacts and free form memory dirs are available to read and write to.

GOALS

- VFS allows free form file schema as well as housing typed orchestration artifacts
- Multi pod and light "HA" survivability
- Session dir from orchestration artifacts create/read only - no edits needed here - flushed consistency
- Memory dir - transactional writes and AURA secure shell reachable (eventual consistency is fine)
- Solidify the idea of "claims" - an instance claims a session and its write dirs atomically one at a time (instance claims for whole agent turn before returning to a user or an other consumer #578 

STRECH GOALS

- can we do this with Archil to avoid rolling our own FUSE + cache? (this is already in process)

Blocks #271 

### Searched Issues

- [x] No similar issues found

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #685, opened 2026-09-11
feat(docker): add cargo_features build arg to the release image

teriyakichild:
## What

Adds a `CARGO_FEATURES` build arg to the Dockerfile's `cook-release` and `release-build` stages, applied to both the `aura` and `aura-web-server` builds, plus a short DEVELOPMENT.md note with the example:

```bash
docker build --build-arg CARGO_FEATURES=session-store-redis --target release -t aura:redis .
```

## Why

The release image compiles with no optional cargo features, so the published image cannot run with `AURA_SESSION_STORE=redis`: startup fails with `session store backend 'redis' requires the 'session-store-redis' cargo feature`. A multi-instance deployment built from this Dockerfile had no way to enable that feature short of editing the file. The aura-sandbox HA rig works around it today by compiling its own binary inside a runner container; moving that rig to Kubernetes needs an image that carries the feature. Context: the shared `memory_dir` / VFS work in #579.

## Design notes

- Empty by default, so the default build renders exactly the commands it did before (`${CARGO_FEATURES:+...}` expands to nothing when unset).
- Applied on the cook stage as well as the build stage so the cargo-chef dependency cache is keyed on the same feature set.
- Redeclared in `release-build` because an `ARG` does not cross a stage boundary without it.

## Verification

- `docker build --target release` with no arg: builds, and a container started with `AURA_SESSION_STORE=redis` exits 1 with the feature message above (negative control).
- `docker build --build-arg CARGO_FEATURES=session-store-redis --target release`: builds, and against a Valkey container `/health` reports `"session_store":{"backend":"redis","ping":{"latency_ms":0,"ok":true}}` with `Session store backend: redis` in the log.
- `make lint`, `make lint-commits`, hadolint (with the file's existing `DL3008` ignore), and shellcheck on the RUN bodies are clean.

Not in scope, possible follow-ups: a `make docker-build` passthrough for the arg, and the chart's `sessionStore.*` values from `docs/design/session-storage.md` section 10 (the chart's existing `extraEnv` covers it in the meantime).

Refs: #579

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -87,6 +87,17 @@ Make targets are composed from modular includes under `.makefiles/` (rust, docke
 | `make docker-test`     | Run the Docker build's lint/test stage         |
 | `make start` / `make stop` | Start/stop the Docker Compose setup        |
 
+### Docker image features
+
+The release image compiles `aura` and `aura-web-server` with no optional
+cargo features. The `CARGO_FEATURES` build arg (a comma-separated cargo
+feature list) enables them for both binaries, for example the Redis/Valkey
+session store a multi-instance deployment needs:
+
+```bash
+docker build --build-arg CARGO_FEATURES=session-store-redis --target release -t aura:redis .
+```
+
 ## Testing
 
 ### Unit Tests
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -90,9 +90,11 @@ Make targets are composed from modular includes under `.makefiles/` (rust, docke
 ### Docker image features
 
 The release image compiles `aura` and `aura-web-server` with no optional
-cargo features. The `CARGO_FEATURES` build arg (a comma-separated cargo
-feature list) enables them for both binaries, for example the Redis/Valkey
-session store a multi-instance deployment needs:
+cargo features. The `CARGO_FEATURES` build arg accepts a comma-separated list
+of features shared by both packages (`aura-cli` and `aura-web-server`) and
+enables them for both binaries; a feature only one package defines fails the
+build. For example, the Redis/Valkey session store a multi-instance deployment
+needs:
 
 ```bash
 docker build --build-arg CARGO_FEATURES=session-store-redis --target release -t aura:redis .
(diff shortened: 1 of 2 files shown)


Shearerbeard (dismissed):

Dockerfile
```diff
@@ -288,24 +288,29 @@ COPY crates/ ./crates/
...
+# binaries (e.g. session-store-redis). Empty by default; the cook and the
+# build both apply it so the dependency cache is keyed on the same set.
 FROM core AS cook-release
+ARG CARGO_FEATURES=""
 WORKDIR /usr/src/app
 COPY --from=planner /usr/src/app/recipe.json recipe.json
 ENV CARGO_TARGET_DIR=/usr/src/app/target
-RUN cargo chef cook --release -p aura-cli --bin aura --recipe-path recipe.json \
- && cargo chef cook --release --bin aura-web-server --recipe-path recipe.json
+RUN cargo chef cook --release -p aura-cli --bin aura ${CARGO_FEATURES:+--features "$CARGO_FEATURES"} --recipe-path recipe.json \
```
jmoses: might be useful to do the expansion once, but  ¯\_(ツ)_/¯

jmoses (dismissed):

DEVELOPMENT.md
```diff
@@ -87,6 +87,17 @@ Make targets are composed from modular includes under `.makefiles/` (rust, docke
 | `make docker-test`     | Run the Docker build's lint/test stage         |
 | `make start` / `make stop` | Start/stop the Docker Compose setup        |
 
+### Docker image features
+
+The release image compiles `aura` and `aura-web-server` with no optional
+cargo features. The `CARGO_FEATURES` build arg (a comma-separated cargo
+feature list) enables them for both binaries, for example the Redis/Valkey
```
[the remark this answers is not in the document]
teriyakichild: Taken in 6fc5a350: the doc now says the list must name features both packages define, and the Dockerfile comment carries the same rule.

Dockerfile
teriyakichild: Done in 6fc5a350: each RUN does `set -- ${CARGO_FEATURES:+--features "$CARGO_FEATURES"}` once and passes `"$@"` to both cargo calls. Keeps shellcheck quiet too.

Shearerbeard (approved):

PR #687, opened 2026-09-11
Webhook poll delivery (3/4): park-path activation and credentials at rest

Shearerbeard:
Third of four. Routes webhook-poll approvals through the park path and persists what the reconciler needs after a restart; the mode is still off.

What changes:

- `DecisionRoute` gains a `park_registry()` accessor, present for the conversational and webhook-poll routes and absent for webhook-sync. The park guard, `park_enabled`, the park route gate, and the pre-call gate use it instead of matching on the conversational route. Two fixes fall out: decided approvals are removed after resume on any park route, and a worker that dies before its snapshot cancels its parked approvals on any park route rather than only the conversational one.
- The parked approval record carries the resolved `headers_from_request` values captured when the request came in, so the reconciler sends each notify with the row's own headers. A mapped header with no usable value and no static fallback fails the registration closed, and a value outside visible ASCII counts as unusable, since the values are persisted as text.
- The decision record carries optional approver identity. It travels on a new `ResolvedDecision` carrier through `ApprovalStore::resolve` and `ApprovalStore::decision`, the registry, and the recorded decisions; the decision bus still publishes only the credential-free decision. Both new record fields have serde defaults, and records written by current builds still decode.
- Where credentials live at rest. The egress headers sit on the undecided approval row and are cleared from the decision envelope at resolve, because a decided id is never notified again. Approver identity sits on the decision record from resolve onward. On disk both are 0600 files in 0700 directories (from the first PR); expired file rows are unlinked by the pending scan, while Redis rows expire through a TTL equal to the remaining decision window. `Debug` on both records prints header names only.
- There is one change to the conversational park path too. A call decided between the gate hit and the checkpoint commit is kept in the checkpoint and consumed at resume, where it used to be dropped. It costs one recorded-decision read per pending call on the commit path.
- The webhook config fingerprint gains `delivery`.
- `docs/design/hitl.md`, `docs/design/session-storage.md`, and `docs/adr/2026-08-13-approver-identity-forwarding.md` record the same lifetime.

Everything except the switch that turns the mode on is now in place.

Tested with file and Redis store contract tests for both new fields and the cleared envelope, legacy-record decode tests, the status-envelope matrix (all three statuses with absent, null, and string reasons, plus unknown-status rejection), registry and route unit tests, and tests for the two park-route fixes. Redis tests ran against a local Valkey; the header-forwarding integration suite ran against mock-mcp.

Stack: based on `mshearer/271-poll-protocol`. Next: `feature/aura-next-271-poll-delivery`.

### DIFF
--- a/crates/aura/src/hitl/decision.rs
+++ b/crates/aura/src/hitl/decision.rs
@@ -63,6 +63,67 @@ pub enum ApprovalDecision {
     Denied { reason: Option<String> },
 }
 
+/// A decision plus the approver identity captured alongside it.
+#[derive(Clone, PartialEq, Eq)]
+pub enum ResolvedDecision {
+    Approved {
+        /// Captured approver identity.
+        identity: Option<crate::approver_headers::ApproverHeaders>,
+    },
+    Denied {
+        reason: Option<String>,
+    },
+}
+
+impl ResolvedDecision {
+    /// An approved decision with its captured identity.
+    #[must_use]
+    pub fn approved(identity: Option<crate::approver_headers::ApproverHeaders>) -> Self {
+        Self::Approved { identity }
+    }
+
+    /// The credential-free decision, the only part that reaches the bus, the
+    /// wake, and the events.
+    #[must_use]
+    pub fn decision(&self) -> ApprovalDecision {
+        match self {
+            Self::Approved { .. } => ApprovalDecision::Approved,
+            Self::Denied { reason } => ApprovalDecision::Denied {
+                reason: reason.clone(),
+            },
+        }
+    }
+}
+
+impl From<ApprovalDecision> for ResolvedDecision {
+    fn from(decision: ApprovalDecision) -> Self {
+        match decision {
+            ApprovalDecision::Approved => Self::Approved { identity: None },
+            ApprovalDecision::Denied { reason } => Self::Denied { reason },
+        }
+    }
+}
+
+impl std::fmt::Debug for ResolvedDecision {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            // Identity values are credentials: names only, the ApproverHeaders
+            // precedent.
+            Self::Approved { identity } => f
+                .debug_struct("Approved")
+                .field(
+                    "identity_names",
+                    &identity
+                        .as_ref()
+                        .map(crate::approver_headers::ApproverHeaders::captured_names)
+                        .map(Iterator::collect::<Vec<_>>),
+                )
+                .finish(),
+            Self::Denied { reason } => f.debug_struct("Denied").field("reason", reason).finish(),
+        }
+    }
+}
+
 /// Why a parked approval was cancelled rather than decided or timed out.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize)]
 pub enum CancelReason {
--- a/crates/aura/src/hitl/gate.rs
+++ b/crates/aura/src/hitl/gate.rs
@@ -12,7 +12,7 @@ use aura_config::GlobPattern;
 use rig::tool::ToolError;
 use serde_json::Value;
 
-use super::decision::{AgentScope, ApprovalOrigin, ApprovalOutcome, DecisionId};
+use super::decision::{AgentScope, ApprovalOrigin, DecisionId};
 use super::protocol::{ApprovalItem, ApprovalRequest, PROTOCOL_VERSION};
 use super::registry::{ParkedApproval, PendingApprovals};
 use super::route::{ApprovalError, DecisionRoute, GateDecision};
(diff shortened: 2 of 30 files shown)


crates/aura/src/hitl/gate.rs
```diff
@@ -339,6 +349,45 @@ fn approval_result_to_pre_call(
...
+            if identity.is_none() && route.requires_identity() {
+                return Err(ToolError::ToolCallError(
+                    "resume mismatch: approved call is missing required approver identity"
+                        .to_string()
+                        .into(),
+                ));
+            }
+            Ok(PreCallOutcome::Proceed {
+                overrides: identity,
+            })
```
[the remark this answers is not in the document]
Shearerbeard: The mapping is outside the fingerprint. `config_fingerprint`'s contract keeps credential fields out so operator tweaks don't invalidate in-flight approvals. The identity replayed at resume was captured legitimately under the mapping in force, bounded by decision expiry. The ADR records the composition and record-then-block behavior.

crates/aura/src/session_store/record.rs
```diff
@@ -71,37 +103,79 @@ pub enum OriginRecord {
...
 pub struct DecisionRecord {
     pub approved: bool,
     pub reason: Option<String>,
     pub decided_at: Timestamp,
+    /// Approver identity captured off the poll-200 (lowercased outbound
+    /// header name → value), persisted in the same record as the decision.
+    /// Additive: absent on records stored before identity docking existed,
+    /// decoding to `None` (uncaptured).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub identity: Option<BTreeMap<String, String>>,
```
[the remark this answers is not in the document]
Shearerbeard: Documented posture. The ADR addendum bounds each value by its last use: resolve clears the egress headers, the pending scan unlinks expired undecided rows, resume removes the decision envelopes it consumes on park routes (that removal is this PR's fix). A run whose process dies after resolve keeps the decision-record retention at 0600 with names-only Debug. A sweep for consumed envelopes is a reasonable follow-up.

crates/aura/src/hitl/decision.rs
```diff
@@ -63,6 +63,73 @@ pub enum ApprovalDecision {
     Denied { reason: Option<String> },
 }
 
+/// A decision plus the approver identity captured alongside it: the carrier
+/// every decision-bearing store API (`resolve`, `decision`) and
+/// `RecordedDecisions` moves as a unit.
+///
+/// Identity exists only on an approval — a denial carries nothing to forward —
+/// so the pairing is the enum itself, not two parallel `Option`s.
```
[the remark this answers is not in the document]
Shearerbeard: Fixed in 6772c5a5.

crates/aura/src/hitl/decision.rs
```diff
@@ -63,6 +63,67 @@ pub enum ApprovalDecision {
...
 
+/// A decision plus the approver identity captured alongside it.
+#[derive(Clone, PartialEq, Eq)]
+pub enum ResolvedDecision {
+    Approved {
+        /// Captured approver identity.
+        identity: Option<crate::approver_headers::ApproverHeaders>,
+    },
+    Denied {
+        reason: Option<String>,
```
justintime4tea, line 74: Would it make sense to capture the identity of who denied? Obviously, there are timeouts and other mechanisms for denial, but when a real person/subject does the deny, would we want to know their identity? Maybe that's captured a different way somewhere I haven't come across yet?

crates/aura/src/hitl/route.rs
```diff
@@ -806,16 +951,36 @@ impl WebhookClient {
     /// receiver recorded the row but the 2xx never came back — would fault on
     /// every tick until the human decides.
     ///
-    /// A timeout is a transport fault, not a decision-shaped outcome — the
-    /// reconciler simply retries on the next tick.
-    pub(crate) async fn notify(&self, request: &ApprovalRequest) -> Result<(), ApprovalError> {
+    /// A 409 means delivered too, but only in one pinned shape: the receiver's
```
justintime4tea: L#954 - L#965 is a duplicate of the text from L#941 - L#952. It seems it became duplicated when adding the `row_headers` field and comment?

crates/aura/src/hitl/route.rs
```diff
@@ -806,16 +951,36 @@ impl WebhookClient {
...
+    /// its unique `decision_id` index answering 409 with
+    /// `{"code": "ECONFLICT"}`. Only that body is a delivery proof; a bare or
+    /// differently-shaped 409 is not, and faults so the reconciler retries.
+    /// The contract is pinned by the receiver's own integration test
+    /// (answerbook/ai-governance-service,
+    /// `test/integration/routes/v3/governance/workflow/post-authorize.js`:
+    /// re-sending a recorded `decision_id` expects 409 with `code: ECONFLICT`
+    /// and writes no second row). Without this arm, a torn notify — the
+    /// receiver recorded the row but the 2xx never came back — would fault on
+    /// every tick until the human decides.
```
justintime4tea, line 965: L#954 - L#965 is a duplicate of the text from L#941 - L#952. It seems it became duplicated when adding the `row_headers` field and comment?

crates/aura/src/hitl/route.rs
```diff
@@ -301,20 +326,69 @@ async fn webhook_round_trip<T>(
...
+    /// marker; this switch must not grow a second one.
+    pub(crate) fn park_registry(&self) -> Option<(&PendingApprovals, Duration)> {
+        match self {
+            Self::Conversational { registry, timeout } => Some((registry, *timeout)),
+            Self::Webhook {
+                client,
+                registry,
+                timeout,
+                ..
+            } => client.poll_delivery().then_some((registry, *timeout)),
```
justintime4tea, line 357: Under poll delivery, both sync consumers of timeout are behind poll_delivery_guard. That makes timeout_secs the parked row's lifetime rather than a request deadline and it keeps the sync default of 300s, so a poll approval expires 5 minutes after parking unless the operator knows to raise it. 

I see the note at gate.rs:147 that this is interim until a park TTL exists but my concern is the default. Would a poll-specific default (or requiring an explicit timeout_secs under poll) be worth it before the TTL lands? 300s seems rather short lived, no?

PR #701, opened 2026-09-15
feat(mcp): add you.com web search to the /mcp add catalog

mouse-value-add:
## What

Adds You.com as a sixth entry in the curated `/mcp add` catalog (`crates/aura-cli/src/repl/mcp/catalog.rs`), using the keyless free profile (`https://api.you.com/mcp?profile=free`) — no credentials collected, no env var, no new dependency. It's the same `Template::Http` shape as the Mezmo entry, just with an empty header list.

## Why

The catalog today is entirely observability and ticketing (Mezmo, PagerDuty, Datadog, Linear, Kubernetes). An SRE agent answering "what's in the latest containerd release notes" or "summarize the advisories for this CVE" has no blessed search path — the only option is the custom-server flow, where you have to know the transport, URL, and auth shape by hand. This gives `/mcp add` a one-pick answer for web search, and it works with zero setup because the free profile needs no key.

The wizard already does the right things around it: it shows the prerequisites, asks `Continue? [Y/n]`, verifies the connection live before writing anything, and prints a starter prompt. Nothing is enabled by default — nothing changes for anyone who doesn't pick the entry.

## What it writes

```toml
[mcp.servers.youcom]
transport = "http_streamable"
url = "https://api.you.com/mcp?profile=free"
description = "You.com web search and research (keyless free tier, no credentials)"
```

The free profile exposes `you-search` (web search with ranked results) and `you-discover`. For users who want the fuller tool set (page-content extraction, research, finance), the authenticated endpoint `https://api.you.com/mcp` takes a `YDC_API_KEY` — that's reachable today through the existing custom-server path, so I didn't add a second catalog entry for it.

## Validation

- `cargo fmt --check` — clean
- `cargo clippy -p aura-cli` — clean, no warnings
- `cargo test -p aura-cli` — pass (crate has no unit tests; the change is a const table)
- `cargo test --workspace` — 1185 passed; 3 failures in `orchestration::park` / `orchestrator` tests, which reproduce identically on a clean checkout of `nightly` (they simulate temp-file write failures via permissions, which don't fail the same way when running as root) — pre-existing and unrelated to this change
- Live probe of the endpoint the wizard will verify against: streamable-HTTP `initialize` + `tools/list` against `https://api.you.com/mcp?profile=free` returns `you-search` and `you-discover`

## Notes

Happy to adjust the wording of the description/prerequisites or the starter prompt. If you'd rather keep the catalog strictly observability-focused, the alternative shape is a docs-only mention of the endpoint in the custom-server walkthrough — say the word and I'll rework it.

You.com also ships prebuilt agent skills for this kind of setup (`npx skills add youdotcom-oss/agent-skills`), in case that's useful context for AURA's own skills story.


mouse-value-add:
Good catches — fixed in 1595aff4:

- Added the missing space after `(rate-limited web search).` so the wizard renders the two sentences correctly.
- Dropped the "read page content" claim from the prerequisites — the keyless free profile exposes `you-search`/`you-discover`, and page-content extraction needs the authenticated endpoint, so the text now only promises web search.

On the documentation point: user-facing docs live in the `mezmo/documentation` repo (`aura/`) rather than here, and I don't have a change staged there yet. Happy to open a companion PR there covering the new `youcom` catalog entry — let me know which page under `aura/` it should land on (MCP setup/reference?) and I'll add it.

One note on the CLA check: signing the CLA is a legal agreement I'd need to route through my side before replying with the signature line — I'll follow up on that shortly.

mouse-value-add:

mouse-value-add:
The outstanding docs requirement is now handled: the usage guide for the You.com catalog server lives at [mezmo/documentation#129](https://github.com/mezmo/documentation/pull/129) — `aura/youcom-web-search.mdx` (wizard flow, manual `http_streamable` TOML, upgrade path to the authenticated endpoint, privacy note), plus You.com rows in the `quickstart.mdx` catalog table and the `cli-reference.mdx` wizard list, per the contributing guide's rule that new features get a usage guide in the docs repo rather than here. Happy to fold it into a different page if the docs team prefers.

PR #711, opened 2026-09-19
test(telemetry): replace the sink mock server with shimforge

tanglearncode:
## What

The classify tests in `crates/aura-telemetry/src/sink.rs` no longer start a wiremock server or connect to a closed port. Each test gets a real `reqwest::Error` from a URL that does not parse, and uses [shimforge](https://crates.io/crates/shimforge) to mock only the check it covers (`status`, `is_timeout` or `is_connect`). Test names and expected labels are unchanged.

## Why

- The tests no longer depend on a local server or timing.
- `connection_refused_is_network` fails on Windows: a refused localhost connect takes about 2s there and hits the test's 2s timeout. It fails on unchanged `nightly` too ([run in my fork](https://github.com/tanglearncode/aura/actions/runs/35418190510)).

Real timeout and 5xx handling is still covered end to end by `tests/wire_format.rs`.

## Testing

`cargo test -p aura-telemetry --lib`, fmt and clippy pass on Linux, macOS and Windows, x86-64 and ARM64 ([run in my fork](https://github.com/tanglearncode/aura/actions/runs/35418679798)).

Disclosure: I maintain shimforge (MIT). It is added as a pinned dev-dependency only.


tanglearncode:

